PR #216, opened 2015-01-07
ignore my first commit, use this

Zuckonit:
judge whether an element is in the list before removing it

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -1,5 +1,5 @@
-"""isort.py.
 
+"""isort.py.
 Exposes a simple library to sort through imports within Python code
 
 usage:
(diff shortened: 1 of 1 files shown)


timothycrosley:
Hi @Zuckonit,

Thanks for the help! It looks like the only thing your missing in this commit is to return the import_string after removing the unneeded tokens. If you can fix this pull request to pass the unit tests I will merge right away.

Thanks!

~Timothy

timothycrosley:
Hi @Zuckonit, 

Thanks so much for your solution! @timgraham helped incorporate into a commit that is now in the mainline repo, and you have been added to the acknowledgements file.

Thanks!

~Timothy

PR #218, opened 2015-01-15
Added git hook and documentation

helenst:
A git hook for isort. Can be called from pre-commit to either prevent commit or just warn about problems.

I had to make sure file_contents took precedence over file_name - it passes in staged contents from the git repo without needing to create tempfiles, but stlil needs to pass in the original filename so that SortImports can report which files have problems. I hope this won't cause problems.


timothycrosley:
This is great! Thanks so much

~Timothy

PR #219, opened 2015-01-15
Prevent line containing isort:import tripping up over itself

helenst:
I ran isort on isort.py and the line for `isort:imports` found and modified itself. Fixing that by breaking up strings.


timothycrosley:
Looks great!

Thanks

~Timothy

PR #222, opened 2015-01-23
Support forcing line wrap for from imports

### LINKED ISSUE #217: How do I set the config for one import per line with hanging vertical formatting?
What config options do I set to get the following results:

When there is a single import from a library it should be on one line and where there are multiple imports they should be using the hanging vertical formatting (#3). I can only seem to force this behavior with the line_length option, but that often results in a \ being used as a continuation.

```
from my.very.long.import.path import module
from os import path
from sys import (
    argv,
    stdout,
)
```

and not

```
from my.very.long.import.path \
    import module
from os import \
    path
from sys import (
    argv,
    stdout,
)
```

this is the config I am currently using
```
[isort]
add_imports=__future__.absolute_import
default_section=THIRDPARTY
import_heading_firstparty=Project Library
import_heading_stdlib=Standard Library
import_heading_thirdparty=External Libraries
indent=4
known_third_party=django,celery,disqus
line_length=1
multi_line_output=3
diff=1
```

NorthIsUp:
fixes #217

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -379,32 +379,34 @@ class SortImports(object):
                         if not from_imports:
                             import_statement = ""
                         if len(import_statement) > self.config['line_length']:
-                            if len(from_imports) > 1:
-                                output_mode = settings.WrapModes._fields[self.config.get('multi_line_output',
-                                                                                         0)].lower()
-                                formatter = getattr(self, "_output_" + output_mode, self._output_grid)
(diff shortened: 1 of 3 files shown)


NorthIsUp:
bump?

timothycrosley:
Hi @NorthIsUp, 

Soo sorry for the delay, but this looks great! Thanks so much!

~Timothy

PR #233, opened 2015-02-04
(Hopefully) fixed verbatim string indentation doc

reinout:
The example for a verbatim string ("equivalent to 4") was a set of double quotes with a tab inside it.

* I assume it should be four spaces instead of a tab.
* I added backticks so that the spaces aren't collapsed in html.

The weird thing is that it still shows up as a single space when rendered. I'm submitting the pull request anyway as I assume the tab is wrong. And perhaps you see what needs to be done to get four spaces out of markdown :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ own line
 Note: to change the how constant indents appear - simply change the indent property with the following accepted formats:
 *   Number of spaces you would like. For example: 4 would cause standard 4 space indentation.
 *   Tab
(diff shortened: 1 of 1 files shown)


timothycrosley:
Hi @reinout,

Thanks so much for finding this error! I've fixed the problem here:

https://github.com/timothycrosley/isort/pull/235

additionally I've added you to the acknowledgement section for finding this error, I really appreciate it!

~Timothy

PR #237, opened 2015-02-05
add date to unified diff format

NorthIsUp:
most of the tools I need to use expect a date to be included with the united diff.


timothycrosley:
Hi @NorthIsUp,

This looks great! Thanks!

~Timothy

PR #248, opened 2015-02-06
Accept [isort] section in ~/.isort.cfg

reinout:
It is a bit weird to have an `[isort]` section in a `setup.cfg` file, but to require a `[settings]` section if it is in `~/.isort.cfg`.

At least, I had to look at the isort source code to figure out why it wasn't picking up my settings :-)


timothycrosley:
Looks great, and I agree supporting both is a good idea, thanks!

PR #251, opened 2015-02-11
Add command-line switch for combining as imports

brutasse:
It'd be useful to have this option available as a command-line switch :)


timothycrosley:
Looks grreat thanks!

PR #264, opened 2015-03-31
Implement `use_parentheses`.

mvantellingen:
Instead of using the `\` as line continuation allow wrapping the the long line within parentheses as suggested in PEP8 using a config setting.

Note that I'm not sure if I've implemented this correctly, but it seems to work :-)


patrickyevsukov:
Thank you for this, @mvantellingen !

timothycrosley:
This looks great, thanks @mvantellingen!

~Timothy

blueyed:
This is not documented currently - you can only find it by looking at the source.

PR #269, opened 2015-05-01
Print the number of skipped files, if any

jdunck:
It wasn't obvious to me that `__init__.py` files were being skipped.  I see that -vb would print each one, but that is also not default.  I think default-skip should at least be visible with a summary line.

$ PYTHONPATH=/Users/jdunck/open-source/isort isort -rc common/
Skipped 11 files


timothycrosley:
Hi @jdunck,

 I think this is a very logical and informative change and will be including it in the next release.

Thanks!

~Timothy Crosley

PR #274, opened 2015-05-31
Imports in known stanzas should override stdlib modules

dpetzold:
Hello, 

If a module is defined in known_first_party but has the same name as a stdlib module it would not end up in the first_party section. This pull requests changes that. I believe that is the correct behavior.

I also added a flake8 config to setup.cfg. Without it about 60 lint errors were generated. I'd be happy to fix them in another pull request if you like.


dpetzold:
Just going to merge to the other one. They will conflict.

PR #276, opened 2015-06-04
add "io" as known standard library

cmeeren:
The io module has been standard since 2.6, see https://docs.python.org/2/library/io.html


timothycrosley:
Looks great, thanks!

PR #277, opened 2015-06-05
Respect the coding declaration 

wimglenn:
`isort` doesn't handle different source code encodings very well - if you make a python file like 

``` 
# coding: latin1

s = u'ã'
```

Where the file is saved in latin1 encoding, it is a perfectly valid python module in both python 2 and python 3.  But unfortunately `isort`will raise an unhandled exception, because it assumes files are saved in utf8 coding.  Furthermore, `isort` will currently save all files as utf-8 encoding, *but not change the coding declaration of the source*, which is dangerous.  We should either respect the original encoding, or rewrite the coding declaration to mention that the new file is saved in utf-8.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Packages
 *.egg
 *.egg-info
+.eggs
 build
 eggs
 parts
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -756,6 +756,7 @@ def coding_check(fname, default='utf-8'):
     # see https://www.python.org/dev/peps/pep-0263/
     pattern = re.compile(br'coding[:=]\s*([-\w.]+)')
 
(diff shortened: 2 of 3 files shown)


wimglenn:
@timothycrosley  could you have a look at this build please.  I'm not expert in tox , but it looks to me like 
1. the python2.6 tests pull in wrong version of natsort (one which makes use of `'{}'.format(...)` syntax which is unavailable in python 2.6)
2. the python3.2 test seem to fail because pies.overrides don't build properly for some reason 
I don't think either of these errors are related to my changes.

timothycrosley:
Hi @wimglenn,

Thanks for fixing this issue! I've merged it in here https://github.com/timothycrosley/isort/pull/287

~Timothy

PR #281, opened 2015-06-10
Fix Import Line Splitting Bug

patrickyevsukov:
Hi Timothy!

Thanks for creating isort! It's quite useful.

I ran into a bug when using isort to lint a file that contained an import in which the split token "import" appeared in a module's name. This resulted in the module's name being split into non-sense strings.

e.g.
```python
from mycompany.model.size_value_array_import_func import (
    get_size_value_array_import_func_jobs,
)
```
... was split into ...
```python
[u'mycompany.model.size_value_array_', u'_func ', u' get_size_value_array_', u'_func_jobs']
```
I have patched the isort _wrap() function to prevent this from happening. I ensured that all tests passed after my patch was introduced, and I added an additional test to ensure that line splitting is word boundary aware.

All the best,

-Patrick

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -30,6 +30,7 @@ import codecs
 import copy
 import itertools
 import os
+import re
 from collections import namedtuple
 from datetime import datetime
 from difflib import unified_diff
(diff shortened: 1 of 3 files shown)


patrickyevsukov:
Please note: This is in place of #280. I closed that PR in favor of this one, in which I branched before pushing my changes. The changes are identical.

timothycrosley:
@patrickyevsukov,

This looks great! Thanks so much for your contribution to the project. 

~Timothy

PR #283, opened 2015-06-25
Add less than 4.0 limit to natsort dependency

relekang:
The 4.0.0 breaks usage of natsort in isort. This will pin natsort to the latest 3.x.x version of natsort until the project is compatible with the new major version.

[Example of how it breaks](https://ci.frigg.io/frigg/frigg-worker/247/) (see the tox -e isort task)


dupuy:
The only reason the build for this PR failed is that there is an underlying problem with [pip for Python 3.2 builds](https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1095272) - would be really great if this PR could be merged and a minor point release of isort pushed out, as isort is currently broken on pretty much every Python version right now.  Only workarounds are adding an explicit requirement on `natsort<4.0` to any package that requires isort, or modifying the requirement of isort itself to `isort<1.0.1` to eliminate dependency on natsort entirely (the latter is probably not a useful workaround).

Also maybe since pip will probably never be fixed for Python 3.2 (launchpad bug open for 18 months) it might be useful to remove py32 support from the Travis CI test matrix...

relekang:
[natsort 4.0.3 fixes this](https://github.com/SethMMorton/natsort/commit/03cc17b9f5c9e6ff30f5c12ed276e672c91b0371)

PR #294, opened 2015-08-06
Sort all imports alphabetically

gforcada:
Fixes: https://github.com/timothycrosley/isort/issues/271

As it is a bit of an outsider from the rest of the methods I decided to use this approach (basically doing all the sort by myself).

I hope it's acceptable. I event wrote a test, which, btw. was easy and fun! Thanks for having a nice test suite easy to extend!

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -131,7 +131,10 @@ class SortImports(object):
         self._first_comment_index_end = -1
         self._parse()
         if self.import_index != -1:
-            self._add_formatted_imports()
+            if self.config.get('force_alphabetical_sort', False):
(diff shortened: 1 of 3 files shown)


gforcada:
@timothycrosley friendly ping

timothycrosley:
I think this looks great! Thanks for putting the time into this, and sorry I've been busy / didn't get back sooner, will be including this in the next release.

~Timothy

NullSoldier:
I would love a release for this, as it's a feature I'm looking to use. I also hope this is a supported config option and not just a command line option.

PR #303, opened 2015-08-18
Bug test: output repeats comment twice

tomviner:
Re issue https://github.com/timothycrosley/isort/issues/304

XFailing test showing that:

```python
from flask import url_for
# Whole line comment
from service import demo  # inline comment
from service import settings
```

becomes:

```python
from flask import url_for
# Whole line comment
from service import demo  # inline comment
# Whole line comment
from service import settings
```

### DIFF
--- a/test_isort.py
+++ b/test_isort.py
@@ -28,6 +28,8 @@ import os
 import shutil
 import tempfile
 
+import pytest
+
 from isort.isort import SortImports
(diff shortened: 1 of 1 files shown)


timothycrosley:
Thanks for this! I've integrated the test, as failing, and then with a fix in the development branch and so am now closing this pull request.

Thanks!

~timothy

Issue #292, opened 2015-07-23
skip setting should take globs


stchris:
I have a feature request: the "skip" setting should take glob expressions, to exclude multiple files at once. For instance `**/*_pb2.py`.

dchanm:
I'm unsure the proper way to write tests, but I will write them with some guidance on what to do with the base64 blob.

The patch adds an additional settings flag -sg / --skip-glob which works similar to -s / --skip . It uses fnmatch to perform path globbing. 

Currently there are issues with backslash escapes and {} wildcards. A workaround for the former is to use [], e.g to match a literal \*, use [\*] instead of \\\* . The {} limitation is due to the settings splitting on commas.

PR #306, opened 2015-08-19
Add globbing support, fixes timothycrosley/isort#292

dchanm:
This adds support for skipping based on globs. There are more details in the original issue #292 . 

Tests need to be added. I didn't see any file skipping tests in test_isort.py . What would be the proper way to add those tests, something similar to  test_other_file_encodings()?

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -28,6 +28,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 
 import codecs
 import copy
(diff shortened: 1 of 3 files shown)


timothycrosley:
This looks great! Thanks for implementing it, will merge in, in preparation for the next release. I'll write some tests up and then comment here to let you see how they were written if you decide to contribute in the future.

Thanks!

~Timothy

PR #307, opened 2015-08-21
Convert README to rst

### LINKED ISSUE #301: Turn README into reStructuredText for PyPI
It's ugly and unformatted on PyPI, since only reStructuredText is supported there :( https://pypi.python.org/pypi/isort

I will have a look at translating.

adamchainz:
Resolves #301 . Looks a bit better to me since `ini`/`python`/`bash` highlighting is specified :)

You can preview on https://github.com/adamchainz/isort/tree/readme_rst . The command `python setup.py check --restructuredtext` passes which implies it will look fine when uploaded to PyPI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,19 +1,38 @@
-![isort](https://raw.github.com/timothycrosley/isort/master/logo.png)
-=====
+.. image:: https://raw.github.com/timothycrosley/isort/master/logo.png
+    :alt: isort
 
-[![PyPI version](https://badge.fury.io/py/isort.png)](http://badge.fury.io/py/isort)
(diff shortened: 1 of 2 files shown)


timothycrosley:
I really prefer writing markdown over rst, but the converter is always breaking on me / adds another layer to the build process. I consider this a better approach with that consider, thanks so much for taking the time to do this!

~Timothy

adamchainz:
Cool. Yeah markdown is easier but all the python tooling is rst only.

If you run `python setup.py register` it will reupload to PyPI and should improve the look: https://pypi.python.org/pypi/isort :+1:

PR #319, opened 2015-08-31
Project root directory settings for git_hook

aRkadeFR:
When using the git_hook, it searches the settings on every base folder path of the files commited instead of keeping the settings of the root folder.

This fix provides a way that during a git commit, the git_hook search only in the root folder where the command has been run.

Let me know if I'm doing something wrong.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -58,7 +58,7 @@ class SortImports(object):
     def __init__(self, file_path=None, file_contents=None, write_to_stdout=False, check=False,
                  show_diff=False, settings_path=None, **setting_overrides):
 
-        if not settings_path and file_path:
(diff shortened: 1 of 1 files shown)


timothycrosley:
Hi @aRkadeFR, 

Thanks for looking into how to improve the git_hook, the suggested solution can't be accepted simply because it will break existing isort plugins (specifically ones for text-editors) as it removes the ability to run isort from a plugin, and load the settings file the same as it would be if ran from the command line at the path of that file. I would however be open to an alternative solution, especially if it was limitted in scope to the git_hook itself. Or one that added an explicit option not to look up the settings path chain.

Thanks!

~Timothy

aRkadeFR:
ok, thanks

PR #320, opened 2015-08-31
Allow files that are skipped to be non-absolute paths

Amwam:
Particularly useful for situations where `not_skip=__init__.py` and `skip=my/module/__init__.py`


timothycrosley:
Looks good, thanks!

PR #330, opened 2015-09-09
skip files even when file_contents is provided

terite:
git hooks provide both file_path and file_contents. Previously,
should_skip only checked if file_path was provided without
file_contents. This introduced a situation where the isort git hook
would fail on files that isort wouldn't touch.

Related PR: https://github.com/timothycrosley/isort/pull/218


timothycrosley:
Looks great! Thanks!

PR #332, opened 2015-09-15
Point pylama extension to the correct entry point.

AbsoluteMSTR:
Point pylama extension to the correct entry point.


timothycrosley:
Looks great! Thanks!

PR #335, opened 2015-09-16
Fix compatibility with Python 3.5

leepa:
Difflib in py3.5 checks the variable types passed in to unified_diff -
as such it would break saying they aren't strings. Previous Python
versions would just call str() on them rather than checking. This just
adds the str() check and will subsequently fix the issue and still work
on previous Python versions.


timothycrosley:
Looks great! Thanks!

Issue #305, opened 2015-08-19
Add support to run library module (python -m)


AbsoluteMSTR:
Add support to run library module (python -m)
example:
`python -m isort -rc .`
like you can with
`python -m pip install -U isort`

timothycrosley:
Hi @AbsoluteMSTR, I'm definitely open to implementing this, I'm not sure how to do this or what the use case is. Do you have a use case in mind and?

Thanks!

~Timothy

timothycrosley:
Your fix for this has been pushed out with release 4.2.0,

Thanks!

~Timothy

PR #336, opened 2015-09-16
Add support to run library as a module #305

AbsoluteMSTR:
Add support to run library as a module #305.


timothycrosley:
Looks great! Thanks!

PR #344, opened 2015-09-22
Allow imports to be sorted by module, independant of import_type

jmullan:
Allows sorting by module within sections like:

```python
import a
from b import c
import d
```

This is to support how my employer does it.


timothycrosley:
This looks looks great, Thank you!

Issue #339, opened 2015-09-18
Unclear error messages could be more helpful


jmullan:
```
$ isort --version
isort 4.0.0
```

If `place_module` returns an empty string, a KeyError is raised rather than giving an error that the sort order can't be determined.
```
Traceback (most recent call last):
  File "/opt/boxen/homebrew/bin/isort", line 11, in <module>
    sys.exit(main())
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/main.py", line 192, in main
    sort_attempt = SortImports(file_name, **arguments)
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/isort.py", line 133, in __init__
    self._parse()
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/isort.py", line 731, in _parse
    root = self.imports[self.place_module(import_from)][import_type]
```

I fixed that by registering a first-party module in my config, but it's non-obvious, and difficult to figure out which import was causing the problem. :(

Additionally, a relative import (which is it's own problem!) caused this error to be raised:
```
Traceback (most recent call last):
  File "/opt/boxen/homebrew/bin/isort", line 11, in <module>
    sys.exit(main())
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/main.py", line 192, in main
    sort_attempt = SortImports(file_name, **arguments)
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/isort.py", line 133, in __init__
    self._parse()
  File "/opt/boxen/homebrew/lib/python2.7/site-packages/isort/isort.py", line 761, in _parse
    self.imports[self.place_module(module)][import_type].add(module)
```

If I get a change, I'll see about putting up a diff.

timothycrosley:
Hi @jmullan,

Can you retry with the latest release of isort (4.2.0), isort 4.0.0 was released awhile ago, and several issue have been fixed since then. In general you should never be seeing random exceptions when using isort, and if you are it's a bug that should and will be fixed ASAP.

Thanks!

~Timothy

jmullan:
Same problem
```root = self.imports[self.place_module(import_from)][import_type]```

I'll see if I can get a diff together

timothycrosley:
This has been fixed in release 4.2.2 using your merge request, thanks for reporting this issue, and your assistance in getting it resolved.

Thanks!

~Timothy

PR #345, opened 2015-09-22
Better errors if place_module cannot place a module.

jmullan:
See issue #339


timothycrosley:
This looks looks great, and I understand what you mean now. Thank you!

PR #352, opened 2015-10-10
Check key for None; use function def over named lambda

graingert:


isort/natural.py
```diff
@@ -38,9 +38,10 @@ def _natural_keys(text):
 
 def nsorted(to_sort, key=None):
     """Returns a naturally sorted list"""
-    if not key:
+    if key is not None:
```
timothycrosley: I think you accidentally reversed the logic here: should be `if key is None`, otherwise good change! If you can modify it and all the test pass (as I believe they will) will merge ASAP

Thanks!

~Timothy

timothycrosley:
Looks great!

Thanks!

PR #362, opened 2015-11-09
Pylama expects a iterable to be returned.

aequitas:
# pylama . -v
    [.....]
    Run isort {}
      File "/.../lib/python3.4/site-packages/pylama/core.py", line 65, in run
        select=params.get("select", set()), params=lparams):
    TypeError: 'NoneType' object is not iterable


timothycrosley:
This looks great, thanks!

PR #363, opened 2015-11-12
Add default value to config settings

cheungpat:
These default settings are required because `_update_with_config_file`
convert the value from config file to an existing value type. The existing
value type is determined from the value of the `access_key` in the
`default` dict.


timothycrosley:
Looks great!

Thanks!

PR #371, opened 2015-12-10
Make naming of --force-single-line-imports consistent with other params

### LINKED ISSUE #367: Inconsistent naming: --force_single_line_imports
Is very minor, but the command line argument --force_single_line_imports uses underscores as the word separator whereas every other argument uses dashes as the word separator.

levic:
Kept --force_single_line_imports as a hidden option for backwards compat

Fixes #367


timothycrosley:
Looks great! Thanks!

PR #372, opened 2015-12-11
Use `glob` to add VIRTUAL_ENV to paths

blueyed:
The list of hardcoded versions is already out of date.

Apart from that we might also assume that `VIRTUAL_ENV` should be in
`sys.path` already, and remove this completely?!

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -36,6 +36,7 @@ import sys
 from collections import namedtuple
 from datetime import datetime
 from difflib import unified_diff
+from glob import glob
 from sys import path as PYTHONPATH
 from sys import stderr, stdout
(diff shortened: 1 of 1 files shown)


blueyed:
Regarding the last comment about removing it, I've found that it has been added in 94d9d851.

/cc @spookylukey

timothycrosley:
This looks great! Thanks @blueyed

blueyed:
Thanks for merging!

@timothycrosley 
Can you shed some light into why https://github.com/timothycrosley/isort/commit/94d9d851cd990c918c2a4b3da0d8c02ca95c2dba is needed / useful after all?

PR #377, opened 2015-12-14
Do not consider single-letter modules to be constants

blueyed:
Fixes https://github.com/timothycrosley/isort/issues/376.

Does it make sense?
Should have a test then.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -272,7 +272,7 @@ class SortImports(object):
         prefix = ""
         module_name = str(module_name)
         if sub_imports and config['order_by_type']:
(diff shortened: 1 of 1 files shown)


timothycrosley:
This is a tough one, I think ideally we create a list of explicitly single letter constants to go with this, really there's no good way to solve (since a single letter could also likely be a constant) this lack of contextual clues is one of many reasons single letter variables are bad practice. However, I think it's more likely for a single letter to be a class then a constant so I'll merge this in as is, then we can think of creating more complete behavior. Thanks for all your great contributions! I can't tell you how appreciated they are!

PR #381, opened 2015-12-22
Allow specifying virtual env directory

teeberg:
This is helpful for running [pre-commit](https://github.com/pre-commit/pre-commit) hooks since they run in their own virtual environment where `isort` fails to correctly infer whether a package is third-party or not, since no packages are installed there.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -227,7 +227,7 @@ class SortImports(object):
                     return placement
 
         paths = PYTHONPATH
-        virtual_env = os.environ.get('VIRTUAL_ENV', None)
+        virtual_env = self.config.get('virtual_env') or os.environ.get('VIRTUAL_ENV')
         if virtual_env:
             paths = list(paths)
             for version in ((2, 6), (2, 7), (3, 0), (3, 1), (3, 2), (3, 3), (3, 4)):


timothycrosley:
Awesome! Thanks for implementing this!

bartekbrak:
Hi, I just spent 6hrs trying to find the "bug" related to this, a different behaviour observed in [my hook](https://github.com/bartekbrak/brak_hooks/blob/master/brak_hooks.py) run outside virtual env. This is awesome, it didn't occur to me to check develop and recent pull requests. Thanks!

But... this won't work for my team, we can't specify a `virtual_env` config setting in project settings as everyone has a different setup and path, even using different tools, some use pyenv, some virtualenvwrapper, some pure env located somewhere crazy. 

On the other hand, I see no easy way out of this. A mechanism to discover location of the env will never be fully robust, if possible at all. Do you have any thoughts on this? 

One bad idea is to have a local, vcs-ignored file in repository root, a symlink to the env and point the config to it, asking every developer to set it up manually. 

Second bad idea is to ask every developer to set up an env variable with venv path and point config to it. This is worse, as some people will install isort as a root and will never have any user env. 

Thanks anyway for your work @teeberg, I will defnitely check http://pre-commit.com/

teeberg:
I prefer explicit over implicit, so I think the best way would be to allow your developers to have a project-wide configuration to define the path to their virtual environment and use that if specified. If `isort` is installed globally, you may be able to set the system-wide library path as the virtual env. I haven't tried this, but I think the code should be able to handle that.

PR #383, opened 2016-01-07
Fix 'not_skip' not ignoring working, when there is a more specific filepath in the skip list

Amwam:
This would occur when not_skip is applied to `__init__.py`, and skip has a file like `some/module/__init__.py`. This meant that all `__init__.py` files were skipped


timothycrosley:
Looks great, thanks!

PR #387, opened 2016-01-14
Add a lines_between_sections setting

dcwatson:
Pretty self-explanatory. I'd like to be able to still split imports into sections, but not have a blank line between each section. This seems to do the trick nicely, and would allow people to put more space between sections if they wanted, too. Test included.


timothycrosley:
LGTM! Thanks for implementing this :)

~Timothy

dcwatson:
Thanks for accepting it! Any chance I could pester you for a PyPI update?

PR #388, opened 2016-02-02
Add missing shebang and import to git hook

merwok:
Also remove unneeded __main__ check.


merwok:
The py32 testenv failed because of pypa/pip#3390

timothycrosley:
Looks great,

Thanks!

PR #395, opened 2016-02-21
Fix multiline sorting with force-alphabetical-sort

bootandy:
Raised in https://github.com/timothycrosley/isort/issues/364

The separate code for --force-alphabetical-sort has been removed and
a boolean flag added to handle sorting with or without case in the
_module_key function (which handles the sorting).

Note this commit alters these 2 tests:
test_alphabetic_sorting_no_newlines, test_alphabetic_sorting
This results in the lone imports being moved to the top of the file.
I think this is better as it means the force_alphabetical_sort tests
now behave similarly to the other tests in the file.

Please advise if it is ok to change the tests like this. 

The big red commit in isort.py is because lines 444 -> 468 were removed and the 
if statement was removed

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -269,9 +269,13 @@ class SortImports(object):
         return self.index == self.number_of_lines
 
     @staticmethod
-    def _module_key(module_name, config, sub_imports=False):
+    def _module_key(module_name, config, sub_imports=False, ignore_case=False):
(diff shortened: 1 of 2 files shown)


timothycrosley:
Awesome work! Great job reducing the line count while improving the functionality. Thanks!

~Timothy

PR #400, opened 2016-03-22
Add support for globbing in forced_separate

danpalmer:
Addresses https://github.com/timothycrosley/isort/issues/399

This uses `fnmatch` to provide globbing to the `forced_separate` options. This allows things like "all Django models", for example, to be separated into another section.

The current limitation of this is that it does not apply to local imports. Unfortunately, adding this looks like much more work, so I'm inclined to leave the change at this.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -36,6 +36,7 @@ import sys
 from collections import namedtuple
 from datetime import datetime
 from difflib import unified_diff
+from fnmatch import fnmatch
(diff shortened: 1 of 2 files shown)


danpalmer:
Like some of the other PRs here, this is passing except for the `py32` environment because of https://github.com/pypa/pip/issues/3390.

timothycrosley:
Looks great thanks!

PR #410, opened 2016-03-28
Imports added incorrectly in function with docstring

timheap:
If running `isort -a "from __future__ import unicode_literals" file.py`, where `file.py` consists of a function like:

```python
def foo():
    """ Single line triple quoted doctring """
    pass
```

`isort` incorrectly adds the import like so:

```python
def foo():
    """ Single line triple quoted doctring """
from __future__ import unicode_literals

    pass
```

Which is a syntax error. I've written a failing test for this case in this PR, although I am unsure where to start looking to fix it.


timothycrosley:
Thanks for reporting and providing a test case, I'll try to push a fix tonight.

Thanks!

~Timothy

PR #416, opened 2016-03-30
Fix single line imports for hanging indents. 

bootandy:
Fix single line imports for hanging indents. 
Changed 1 test for the above.

Fixes:
https://github.com/timothycrosley/isort/issues/365

I added an if on line 429 to say "If line is too long & we have at least 1 import & we are using a hanging indent wrap mode" -> then we format the line. 

It looks ok, but please check it and point out if I have made a mistake.


timothycrosley:
This looks great! Thanks for the great implemention w/ matching test case

~Timothy

PR #418, opened 2016-04-08
Add Appveyor build control file

jayvdb:


jayvdb:
Note, it can be renamed to `.appveyor.yml` ( I usually do this ), but this requires that everyone sets that filename in their project settings.

timothycrosley:
Looks good to me, thanks!

PR #426, opened 2016-04-16
Fix hanging imports being removed

bootandy:
As mentioned in https://github.com/timothycrosley/isort/issues/425
This fixes hanging imports being removed by isort.

Code was missing a whitespace when glueing the import line together


timothycrosley:
Looks good to me, thanks!

PR #427, opened 2016-04-21
setup.py - remove download_url

adamchainz:
It's unnecessary as pip downloads from PyPI.


timothycrosley:
Fair enough, this is really just still here for historical reasons (from before PYPI being as stable as it is now)

Issue #432, opened 2016-04-29
Adds import inside class when class starts at top of file after encoding comment


adamchainz:
``` sh
$ cat test.py
```

``` py
# -*- coding:utf-8 -*-
class Foo(object):

    def bar(self):
        return 'baz'
```

``` sh
$ isort --diff -a 'from __future__ import absolute_import' test.py
```

``` patch
--- /Users/adamj/Documents/Projects/pyticketswitch/test.py:before   2016-04-29 15:32:17
+++ /Users/adamj/Documents/Projects/pyticketswitch/test.py:after    2016-04-29 15:32:36.902051
@@ -1,5 +1,6 @@
 # -*- coding:utf-8 -*-
 class Foo(object):
+from __future__ import absolute_import

     def bar(self):
         return 'baz'
```

This doesn't happen if there is no encoding comment at the start of the file.

timothycrosley:
Closing as this is now fixed in develop, thanks!

adamchainz:
Awesome 👌

PR #433, opened 2016-05-02
Add failing test case for issue 432

adamchainz:
I've tried to fix it but it seems to be rather tricky and related to a kind of off-by-one error between `self.index` and `output_at` that only occurs when there are no blank lines between the encoding comment and the first piece of code in the module.

For #432.


timothycrosley:
There was some merge issues with this test, but I've copied and pasted it verbatim and added it to my test suite, thanks for providing such a thourough example of the issue to work off of!

PR #434, opened 2016-05-05
Add --virtual-env command line switch

teeberg:
Expose `virtual_env` config on the command line. This is helpful e.g. when running `isort` as an external tool from PyCharm, which doesn't seem to let you specify any environment variables like `VIRTUAL_ENV`.


timothycrosley:
Looks great! Thanks!

PR #444, opened 2016-07-08
Show diff only once

teeberg:
As you can nicely see in the diff, the condition

``` python
if show_diff or self.config.get('show_diff', False) is True:
```

exists twice and results in printing the diff twice if there are any differences. Remove one of them.


timothycrosley:
Looks great, thanks!

PR #450, opened 2016-07-22
Use a copy of sys.path

oxplot:
On a 1300 file project at my company, this change reduces the run-time from 11 minutes down to 5 seconds, a 130x improvement.


teeberg:
Wooow, for me too :-) From 43 minutes down to 10 seconds. Thank you for this patch! It has some merge conflicts now, could you check those out @oxplot so that we can get this merged? :-)

oxplot:
@teeberg done

timothycrosley:
Looks good! Thanks so much!

Issue #445, opened 2016-07-12
Applying 'use_parentheses' should depend on 'multi_line_output' and 'include_trailing_comma'


ryabtsev:
For example:
1. multi_line_output = 0, 1 or 2; include_trailing_comma = true
   from third_party import (
   long_name_of_the_library,)
2. multi_line_output = 3 or 5; include_trailing_comma = true
   from third_party import (
   long_name_of_the_library,
   )
3. multi_line_output = 0, 1 or 2; include_trailing_comma = false
   from third_party import (
   long_name_of_the_library)
4. multi_line_output = 3 or 5; include_trailing_comma = false
   from third_party import (
   long_name_of_the_library
   )

RomuloOliveira:
👍

ryabtsev:
Fixed in https://github.com/timothycrosley/isort/pull/451

timothycrosley:
Closing this a as a fix has been merged in develop, thanks!

PR #451, opened 2016-07-24
Fix issue #445: Applying 'use_parentheses' should depend on 'multi_line_output' and 'include_trailing_comma'

ryabtsev:
For example:
1. multi_line_output = 0, 1 or 2; include_trailing_comma = true
   from third_party import (
   long_name_of_the_library,)
2. multi_line_output = 3 or 5; include_trailing_comma = true
   from third_party import (
   long_name_of_the_library,
   )
3. multi_line_output = 0, 1 or 2; include_trailing_comma = false
   from third_party import (
   long_name_of_the_library)
4. multi_line_output = 3 or 5; include_trailing_comma = false
   from third_party import (
   long_name_of_the_library
   )

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -334,7 +334,14 @@ class SortImports(object):
 
                     cont_line = self._wrap(self.config['indent'] + splitter.join(next_line).lstrip())
                     if self.config['use_parentheses']:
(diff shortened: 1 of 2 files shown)


timothycrosley:
Looks great! Thanks!

PR #452, opened 2016-07-26
Make stdout capturable

harai:
This makes `stdout` capturable by using [this hack](http://stackoverflow.com/a/10743550/1160341).


timothycrosley:
Looks good, thanks!

Issue #420, opened 2016-04-09
No module named isort.__main__


jayvdb:
`isort` doesnt support `python -m` syntax.

```
$ python -m isort 
/usr/bin/python: No module named isort.__main__; 'isort' is a package and cannot be directly executed
```

timothycrosley:
Closing as a fix has been merged into develop, thanks!

PR #461, opened 2016-08-24
Rename __main__ to __main__.py #420

AnkurDedania:
Rename **main** to **main**.py #420


sebix:
Why does this exists?

AbsoluteMSTR:
allows you to call isort from python. Example, `python3 -m isort` instead of just calling `isort` useful if you want to explicitly make python use python3 instead of 2.x. Since `pip install -U isort` and `pip3 install -U isort` overwrite each other.

MinchinWeb:
Turns out this already exists, so I'll add my :+1: rather than opening another pull request!

timothycrosley:
Looks good, thanks!

PR #465, opened 2016-09-01
Place modules found in sys.prefix/lib in stdlib section

### LINKED ISSUE #464: Some builtins not properly placed on non-standard Python installations (conda)
Consider this file:

``` python
import thirdparty
import os
import ftplib
import ABC
import numbers
import colorama
```

On Windows with a **conda** Python 2.7 installation, `isort` produces this:

``` python
import os

import colorama

import ABC
import ftplib
import numbers
import thirdparty
```

(Note that `ftplib` and `numbers` were incorrectly placed in the "third party" section).

It seems the problem lies in the logic which decides which files are third-party (`isort.py:265`):

``` python
                elif 'python2' in prefix.lower() or 'python3' in prefix.lower():
                    return self.sections.STDLIB
```

Standard python is installed by default in `C:\Python27` or `C:\Python34` which explains why the above works in general, but in systems which have Python from anaconda or even Python installed in a non-standard location that logic does not work and the library is considered "thirdparty".

My Python is installed at `E:\Miniconda\python.exe` , so here's the path to `ftplib`:

```
$ python -c "import ftplib;print(ftplib.__file__)"
E:\Miniconda\lib\ftplib.pyc
```

A workaround is to add those builtin files in the `known_standard_library` config options.

nicoddemus:
This uses `sys.prefix` to decide if a file is from the stdlib instead of relying that `python2` or `python3` appear anywhere on the PATH.

This solves the problem of Python from official distribution installed in a non-standard location or other distributions (such as conda).

I don't know how I can write a test for the actual problem (Python installed in a non-standard location) because both AppVeyor and Travis have Python installed in a standard location. :sweat: 

Fix #464

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -254,15 +254,20 @@ class SortImports(object):
             paths += [path for path in glob('{0}/src/*'.format(virtual_env)) if os.path.isdir(path)]
             virtual_env_src = '{0}/src/'.format(virtual_env)
 
+        # handle case-insensitive paths on windows
+        stdlib_lib_prefix = os.path.normcase(get_stdlib_path())
+
         for prefix in paths:
             module_path = "/".join((prefix, module_name.replace(".", "/")))
             package_path = "/".join((prefix, module_name.split(".")[0]))
-            if (os.path.exists(module_path + ".py") or os.path.exists(module_path + ".so") or
-               (os.path.exists(package_path) and os.path.isdir(package_path))):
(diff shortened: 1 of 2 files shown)


nicoddemus:
cc @gqmelo @fogo @prusse-martin

timothycrosley:
Looks good, thanks!

nicoddemus:
:+1:

PR #467, opened 2016-09-06
Fix multiline comments being moved

bootandy:
Fix issue reported here: https://github.com/timothycrosley/isort/issues/460

Multiline comments were being moved into the wrong place.

Test added.

***_Please Note**_*
isort.py  line: 865 was changed in commit: f157b06 and I do not see
why the more complex comparison was needed so it has been removed. Please reread the original commit before accepting this merge request.

### DIFF
--- a/test_isort.py
+++ b/test_isort.py
@@ -1767,6 +1767,18 @@ def test_import_by_paren_issue_375():
     assert SortImports(file_contents=test_input).output == 'from .models import Bar, Foo\n'
 
 
+def test_import_by_paren_issue_460():
+    """Test to ensure isort can doesnt move comments around """
+    test_input = """
+# First comment
+# Second comment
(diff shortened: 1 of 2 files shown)


test_isort.py
```diff
@@ -1767,6 +1767,18 @@ def test_import_by_paren_issue_375():
     assert SortImports(file_contents=test_input).output == 'from .models import Bar, Foo\n'
 
 
+def test_import_by_paren_issue_460():
+    """Test to ensure isort can doesnt move comments around """
+    test_input = """
```
nicoddemus: Quick comment: using implicit string literal continuations using parentheses seem to be preferred in the rest of the tests.

timothycrosley:
Looks good, thanks!

PR #469, opened 2016-09-08
Add always_wrap_multiple_from setting.

carljm:
At Instagram we would like to switch to `isort` to enforce our imports coding style, but our style requires use of the "vertical - hanging indent" style for _all_ `import from` statements with multiple names imported. It doesn't matter how long they would be on a single line, they should always be wrapped with each imported name on its own line. E.g. even if `from foo import bar, baz` easily fits within 80 characters, we always write it as:

```
from foo import (
    bar,
    baz,
)
```

The reason for this is that it minimizes pollution of `git annotate` and makes it easier to track down who added/removed a single import, and when.

This pull request adds an `always_wrap_multiple_from` boolean setting which implements this.


ryabtsev:
A similar setting is already there: "force_grid_wrap" (force from imports to be grid wrapped regardless of line length)

ryabtsev:
A similar setting is already there: "force_grid_wrap" (force from imports to be grid wrapped regardless of line length)

carljm:
@ryabtsev Thanks! Looks like `force_grid_wrap=true` along with `multi_line_output=3` does exactly what I want; just mis-led by "grid" in the setting name.

PR #475, opened 2016-10-07
added trailing comma if there is wrapping of a single item in from import

ufranske:
Initial import statement to wrap:

```
from third_party_librariry_with_long_name import single_item_to_import
```

How it worked if `include_trailing_comma=True` and `use_parentheses=True`:

```
from third_party_librariry_with_long_name import (
    single_item_to_import)
```

How it will work if `include_trailing_comma=True` and `use_parentheses=True`:

```
from third_party_librariry_with_long_name import (
    single_item_to_import,
)
```

How it will work if `include_trailing_comma=True`, `use_parentheses=True` and `multi_line_output=3` or `multi_line_output=5`:

```
from third_party_librariry_with_long_name import (
     single_item_to_import,
)
```


timothycrosley:
Looks good, thanks!

Issue #476, opened 2016-10-07
make force_grid_wrap configurable by max number of from imports


pjenvey:
force_grid_wrap could be more useful as an int value configuring the number of from imports that force the wrap

This could be somewhat backwards compatible by parsing the current enabling value of True (or additionally 1, maybe with a warning) to mean a value of 2

timothycrosley:
Closing this as your fix is now merged into develop, thanks!

PR #482, opened 2016-11-26
switch force-grid-wrap to an integer value (#476)

pjenvey:
This switches --force-grid-wrap to a configurable int (#476)

Almost fully backwards compatible except one small case: when -fgw is the last arg specified before files. That's a simple fix for anyone running into it after seeing the new --help text.

More importantly this doesn't impact config files, which I'd think is more relied upon than command line args.


timothycrosley:
Looks good, thanks!

PR #485, opened 2016-12-07
should match indentation with other parts of code

tokenmathguy:
technically this works, but it should be four spaces, not three


timothycrosley:
Looks good, nice catch, thanks!

PR #486, opened 2016-12-08
Add missing python standard libs

rudineirk:
This libs were missing:

* asyncio
* concurrent
* http
* html
* typing
* xmlrpc
* ftplib
* numbers


timothycrosley:
Looks good, thanks!

tkukushkin:
@timothycrosley `typing` isn't standart module in python2. This PR breaks pep8.

timothycrosley:
@tkukushkin this PR doesn't break pep8: 

isort can't know what version of Python your code will run on, and even if you are making a Python2 app it is good practice to avoid using Python3 module names in case you ever decide to port that app to Python3 or have it run on both - a very common decision.Finally,  If you really do want to include `typing` in a Python2 project you can always tell isort that it's a project level import and it will sort it correctly.

Thanks!

~Timothy

PR #492, opened 2017-01-10
add new wrap mode: single import in one line

viewstar000:
add new wrap mode, to adapt my company's code style


MinchinWeb:
Would it be simple to add an example to the [Readme](https://github.com/timothycrosley/isort#multi-line-output-modes)? I not quite sure what this would look like.

Issue #453, opened 2016-07-26
`async def` syntax erroneously parsed as something import-related


underyx:
## Steps to reproduce

1. Create the following file:

```
import asyncio


async def main():
    return
```

2. Run `isort` on it.

`isort` should not modify the file at all, but it actually changes it to

```
import asyncio

async def main():
    return
```

samuelcolvin:
I don't think it's that isort thinks `async def` is related to import, but rather that isort isn't aware that `async def` should have two blank lines before it.

This is now enforced by pycodestyle as of 2.1.0 so running isort will break flake8 etc.

Is there anything we can do to help fix this?

samuelcolvin:
**bump** any update on this @timothycrosley?

It's fairly annoying that currently isort breaks pep8 on almost all my test files.

timothycrosley:
Closing as a fix for this has been merged into develop

PR #503, opened 2017-03-20
Fix smart lines after import section for async functions

ethifus:
With this changes `isort` should properly add default two blank lines after import section when the next line starts with `async def` #453 .


timothycrosley:
Looks good, thanks!

PR #507, opened 2017-05-07
Drop support for Python 2.6

jdufresne:
I would humbly like to suggest isort drops support for Python 2.6.

The last release of Python 2.6 was 2013-10-29, over 3 years ago. It is
no longer receiving security fixes.

https://www.python.org/dev/peps/pep-0361/

The pip project itself has recently dropped support for 2.6. Their
numbers estimate that Python 2.6 accounts for ~2% of their downloads.
See:

https://github.com/pypa/pip/pull/4343

For projects that still use Python 2.6, they can continue to pip install
an older version of isort.

I've tried my best to remove as much 2.6 specific code as I can,
including the 'Programming Language :: Python :: 2.6' trove classifier
from setup.py. I've also removed Travis CI testing, which should result
in faster testing and fewer wasted resources.

Code changed:

- Removed Python2.6 from documentation
- Removed Python2.6 from testing configuration
- Use set literals
- Use dict comprehension
- Pass generator expressions to builtin functions instead of lists
- Use builtin OrderedDict
- Removed deprecated U open mode; replace with io.open() which support
  universal newlines
- Removed Python2.6 workarounds throughout the code

Thanks for considering.


timothycrosley:
I agree with this sentiment. I'll try to push one more version of isort with 2.6 support using the many fixes that are waiting in develop and that afterwards merge this in and go with 2.7+ going forward. Honestly, I look forward to the day I only have to support 3, like the case is for hug.

Thanks for all the great work!

~Timothy

jdufresne:
> Honestly, I look forward to the day I only have to support 3

If you'd like, I can modify this PR or submit a follow up PR to drop all Python 2 support. What is your preference? I was trying to be conservative by continuing support for Python 2.7. I'm only involved with Python 3 projects, so my projects would continue using the latest isort version. Let me know what you think.

timothycrosley:
I think you initial intuition was correct, as there unfortunately still many 2.7 projects. I'll push out one final 2.6 release this week, merge in this change and support 2.7 - through the rest of the year and then 3+ only.

Thanks!

~Timothy

jdufresne:
Rebased to resolve merge conflicts

timothycrosley:
The last Python2.6 release has just been pushed! `4.2.8`. Thanks for the great work!

jdufresne:
Any reason this was reverted?

timothycrosley:
Definitely unintentionally. I accidentally merged it into master from a hot fix branch and then had to revert it to do further hotifixes (this is what happens when you don't do a release in over a year O:-) and I'm sure I at some point merged master back in to make develop in sync with the latest.

jdufresne:
No worries. I've restored the change in PR #562.

PR #508, opened 2017-05-08
Fix incomplete variable rename

arrdem:


timothycrosley:

~Timothy

PR #526, opened 2017-05-28
Remove unnecessary CLI arg defaults

jdufresne:
When using CLI action store_true or store_false, the opposite value is
always the implied default. No need to specify it.

From https://docs.python.org/3/library/argparse.html#action

> 'store_true' and 'store_false' - These are special cases of
> 'store_const' used for storing the values True and False respectively.
> In addition, they create default values of False and True
> respectively.


timothycrosley:
Looks good, thanks!

PR #527, opened 2017-05-28
Remove unnecessary dict.get() calls when retrieving config values

jdufresne:
Now simply use dict indexing instead. Uses simpler and more conventional Python syntax for directly accessing a value in a dict. A small micro optimization for keys known to exist as they're specified in
`isort.settings.default`. Will also more easily expose typos and programming errors as an incorrect key now results in a `KeyError` instead of silently falling back to a default value.

There remain a few `config.get()` calls as changing them causes test failures. I wasn't sure which case they fall under:

* Truly optional values that may not exists, dict.get() is the best choice here
* Should be specified in settings.default
* Exposing an unnoticed error

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -72,7 +72,7 @@ class SortImports(object):
             else:
                 self.config[key] = value
 
-        if self.config.get('force_alphabetical_sort', False):
+        if self.config['force_alphabetical_sort']:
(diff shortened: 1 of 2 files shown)


timothycrosley:
I agree with this, merging in the change, thanks!

PR #529, opened 2017-05-28
Reorder CHANGELOG to be in reverse chronological order

jdufresne:
It is a bit unusual and unconventional to order a CHANGELOG with the
most recent release at the bottom. Due to this, it is easy for a new
user to incorrectly read version 4.0.0 as the latest release. Bring
CHANGELOG into more typical ordering to reduce confusion.

Follows suggestions from "Keep a CHANGELOG": http://keepachangelog.com/.

> List releases in reverse-chronological order (newest on top).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,62 @@
 Changelog
 =========
 
-### 4.0.0
-- Removed all external dependencies
+### 4.2.8 - In progress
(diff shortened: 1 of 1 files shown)


timothycrosley:
100% agree with this, it's been bothering me as well, and I actually do it the correct way in my other projects - for example: https://github.com/timothycrosley/hug/blob/develop/CHANGELOG.md

PR #578, opened 2017-07-19
debug place_module

didib:
Show what section each import was placed in, if in verbose mode.

This can help debug differences between systems, in case e.g. some
library was installed on only one of them and therefore was placed
differently.

See also: https://bugzilla.redhat.com/1450161


didib:
Works for me. You might want to squash, change strings, whatever. Perhaps even add '--debug' to activate instead of '--verbose'.

jniederm:
@timothycrosley could you please have a look at this pull?

timothycrosley:
Looks great! Thanks! and sorry for the review delay

PR #580, opened 2017-07-31
Allow customization of comments prefix

mitar:
We use only single space before `#` in comments. This allows one to configure this and other styles of inline comments.


timothycrosley:
I like this concept, currently the tests fail however. If you can update to have passing tests I'm down to merge this, thanks!

mitar:
I think I fixed them. Let's see.

timothycrosley:
Looks good, thanks!

PR #585, opened 2017-08-15
Add a space after the input string.

jwg4:
It looks nicer if the character the user types appears after a space rather than right next to the question mark.


timothycrosley:
Agreed! Thanks for implement this!

PR #588, opened 2017-08-17
Set skipped=True when we skip the file

danpalmer:
We already do this when the configuration specifies skipping a file, but not when the file is empty or the _file_ specifies skipping.

The "skipped" flag is relied upon by some tooling (I'm not sure if it constitutes public API, but it should be fixed anyway), and that tooling is failing in these cases.


danpalmer:
@timothycrosley hey just wondered if you'd had a chance to look at this yet?

timothycrosley:
Looks great, thanks for such a clean PR!
~Timothy

PR #598, opened 2017-09-22
Update parameter order to improve help text readability

bobh66:
I had a hard time looking through all of the options printed by isort -h and thought they might be easier to read if they were in alphabetical order.  It loses a little but of the grouping, especially for the KNOWN_X options, but overall I think it's easier to understand.

Let me know what you think.

Thanks


akonsta:
I would like to work on alphabetizing the argument order.

PR #600, opened 2017-10-10
Fix case-sensitive path existence check in Mac OS X

GeyseR:
Mac OS X works in a case-insensitive manner when checking path existence. This PR applies the same logic for Mac OS X installations as isort has for windows


timothycrosley:
Looks great thanks!

Issue #447, opened 2016-07-12
Option to eliminate localfolder section/combine with firstparty


dan-passaro:
By default isort will take a module with PEP8-compliant imports like this:

```
import decimal

import django

from .utils import spellcheck
from myblog.models import Post
```

And sort them into four sections like this:

```
import decimal

import django

from myblog.models import Post

from .utils import spellcheck
```

I understand the rationale behind the four sections and their ordering, but PEP8 only mentions three distinct sections for imports (see https://www.python.org/dev/peps/pep-0008/#imports). It would be nice if isort accommodated projects which want to maintain a three-section PEP8 import style. I suspect this can be done most easily by having an option where any import that would go into the "localfolder" section gets put into the "firstparty" section instead.

timothycrosley:
This is now supported via no_lines_before config option.

Thanks!

~Timothy

PR #601, opened 2017-10-10
Add no_lines_before option to let sections not be split from previous

GeyseR:
I would like to propose a new option for optionally skip splitting by empty lines for some sections.

In our internal code style we don't split local imports and first party imports because they are parts of the same project and this will add additional 6th level of imports when we already have 5: futures,stdlib,django,thirdparty,firstparty+local

Also, it would help to solve #447


timothycrosley:
Looks good to me, thanks!

blueyed:
Very useful, thanks.

It is missing from https://github.com/timothycrosley/isort/wiki/isort-Settings however.

PR #602, opened 2017-10-12
Fixed improper line seperator conversion

### LINKED ISSUE #493: Unwanted conversion to Windows line endings
Hello,

I am using "isort" with a git checkout that doesn't change line endings. When running "isort" on a file, it changes the Unix new lines into Windows new lines, generating not only a huge diff, but also an unwanted change. Effectively this means I can't use isort on a mixed platform project as e.g. a commit hook or so.

Can you please keep the new line convention used and / or allow to disable usage Windows new files, or allow to specify which kinds of new lines to use.

Thanks,
Kay Hayen

Plazmaz:
This Fixes #493. From the [io.open docs](https://docs.python.org/2/library/io.html#io.open):  

> newline controls how universal newlines works (it only applies to text mode). It can be None, '', '\n', '\r', and '\r\n'. It works as follows:

> On input, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.
> **On output, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '', no translation takes place**. If newline is any of the other legal values, any '\n' characters written are translated to the given string.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -202,7 +202,7 @@ class SortImports(object):
                         return
                     if answer in ('quit', 'q'):
                         sys.exit(1)
-            with io.open(self.file_path, encoding=self.file_encoding, mode='w') as output_file:
+            with io.open(self.file_path, encoding=self.file_encoding, mode='w', newline='') as output_file:
                 output_file.write(self.output)
 
     def _show_diff(self, file_contents):


thebigmunch:
Thanks for the credit, but this is an incomplete and, in turn, incorrect fix.

You have only added this on the write back to disk, leaving the initial read to translate all newlines to ``'\n'`` in the first place. This will still cause an issue with any CRLF line endings. As will the fact that there are hardcoded ``'\n'`` in multiple places for modified code which can cause a problem for CRLF line endings.

So, even if you changed both read and write to pass through line endings untranslated, there is, unfortunately, more to solve here.

Plazmaz:
@thebigmunch better?

thebigmunch:
The option is great.

Not sure if there should be special handling for cases where there are more than one kind of line ending per file or not. This doesn't cover that. This would still change all LF to CRLF if there happened to be even a single stray CRLF in the file when no config is set, for example.

The ideal would be to redo the parsing so that it's determining the line ending based on each original line, so there's no potential to change anything. At least, that's what I imagine the ideal is. As an alternative requiring less work, counting the line endings and using the one most prevalent would be better than just checking for existence.

It's up to @timothycrosley which is the best direction to go for now.

Plazmaz:
@thebigmunch my intent here was not for this to be perfect, or even for you to specify line endings. I simply wanted to remove the auto \r\n conversion, which this has effectively done. You can also force line ending conversion if the autodetect doesn't work, so for the weird cases with mixed line endings(why would anyone actually want this?), you can choose which one you prefer.

thebigmunch:
> mixed line endings(why would anyone actually want this?)

They wouldn't. But I think most people also don't want an import linting tool deciding for them.

IDEs and editors tend to use auto-detection for deciding which line ending to use. These don't always get it right, sometimes often. I know Atom had serious issues with its auto line ending detection for a long time.

> auto \r\n conversion

Well, the issue wasn't limited to converting of '\n' to '\r\n', as isort would also convert the other way around on Linux. That is just isn't as likely to come up.

In the most common case, yes it's fixed that. But, As I pointed out above, in a still reasonable case, it also causes the same kind of issue. And, in the case where this happens, it also breaks parsing as the code now splits lines on what isn't the line ending for the entire file. That is a different regression from the original added by this code.

If isort is going to go with good enough, it *needs* to at least use the most common line ending rather than just checking existence. That is most likely to be the one intended by the user.

> my intent here was not for this to be perfect, or to work at all with mixed line endings

In my opinion, this wouldn't fix the issue. It just makes it less likely to occur. Does this help users? Sure. Should it still be fixed properly down the line? Yes. I was more interested in a proper solution, else I would have made a PR long ago much like this.

Plazmaz:
Alright, well as it currently stands, I've been unable to use isort on Windows for several months. While I can appreciate the need for a full fix, I don't really have time to do that, but need to use the project. If someone else can do that without taking another several months, then that's great, feel free to close this pr and make one that fixes it properly. I do agree... This is not the best fix, it's simply a shim for myself and others who have experienced problems on cross platform repos

timothycrosley:
I think this should get merged as it's a big improvement over the current situation, and I appreciate the great work that was put into it.

Thanks!

~Timothy

PR #604, opened 2017-10-20
In iter_source_code() include Python scripts without an extension

jdufresne:
A file starting with the line:

```
  #!/usr/bin/env python
```

Or

```
  #!/usr/bin/python
```

And other variations, are Python scripts and frequently do not have a `.py` extension. These will now also be included by isort processing

---

Use case: I noticed that scripts in my project did not have sorted imports. I want my scripts to go through the same static analysis as the rest of the project.

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import argparse
 import glob
 import os
+import re
 import sys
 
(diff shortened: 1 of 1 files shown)


blissdev:
Nice change! Very nice for executable scripts, etc.

timothycrosley:
LGTM Thanks!

PR #605, opened 2017-10-20
Remove unconventional space after shebang

jdufresne:
While allowed by most tools, it is very unconventional.


timothycrosley:
LGTM Thanks!

Issue #606, opened 2017-11-01
"import x as y" lost


Anthchirp:
running isort on 
```
import os
import os as _os
```
results in
```
import os as _os
```

Ran into this problem in a large file that imported the same thing twice in different places with different names. Appreciate that this is bad code, but it would only be obvious after sorting, and isort breaks it.

kchomski-reef:
@timothycrosley should this be enabled by flag? or by default?

timothycrosley:
I think this should be configurable, and probably off by default

Anthchirp:
Sorry, but the fix doesn't make a lot of sense to me.

Given this file:
```python
import os
import os as _os
import sys as _sys
_sys.stdout.write("pathsep on %s is %s\n" % (os.name, _os.pathsep))
```

The default behaviour (isort 4.3 without parameters) results in
```python
import os as _os
import sys as _sys

_sys.stdout.write("pathsep on %s is %s\n" % (os.name, _os.pathsep))
```
which is broken code. I do not understand why breaking code should be the default behaviour.

Running ```isort -k``` adds an unused import:
```python
import os
import os as _os
import sys
import sys as _sys

_sys.stdout.write("pathsep on %s is %s\n" % (os.name, _os.pathsep))
```
and while that code still works I don't want to sprinkle unused ```import```s all across my project just to avoid breaking code.

timothycrosley:
This is fixed in develop and will make it's way in isort 5.0.0

Thanks!

~Timothy

PR #607, opened 2017-11-01
add test for #606

Anthchirp:


timothycrosley:
This issue has now been resolved, thanks for reporting it and providing a test case!

PR #610, opened 2017-11-06
merge only neighbour sections in case of enabled no_lines_before option

GeyseR:
Hi!

I've found an unexpected behavior in earlier introduced `no-lines-before` option. Suppose we have STDLIB and local imports. We obviously won't want to merge them into one group. So I think it has much more sense to merge only neighbor sections, which might be configured via `sections` or `forced_separate` options

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -547,6 +547,7 @@ class SortImports(object):
             sections = ('no_sections', )
 
         output = []
+        prev_section_has_imports = False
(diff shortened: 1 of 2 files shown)


GeyseR:
Hi! Is there any change that this PR will be merged soon?

timothycrosley:
This looks good to me, thanks! And sorry about the merge delay

PR #617, opened 2017-12-09
Process files in parallel(!)

thedrow:
This PR contains the necessary code to run isort on multiple files in parallel.
The new class and function are there to avoid pickling problems that occur because the Sections namedtuple is a dynamically generated one.
The new class also allows us to serialize less information back to the main process which provides a performance boost.

I tested it on Celery which is a pretty large code base.
No jobs:
real	0m1.130s
user	0m1.083s
sys	0m0.052s
2 jobs:
real	0m0.730s
user	0m1.272s
sys	0m0.141s
8 jobs:
real	0m0.551s
user	0m2.066s
sys	0m0.128s

We're trying to cut down build times for Celery and this would be immensely useful for us and other large code bases.


thedrow:
I don't know what's wrong. I did resolve the conflict.

timothycrosley:
This looks awesome thanks!

PR #618, opened 2017-12-12
Fix `output` property missing when the input has `isort:skip_file`.

h-hirokawa:
When the input file contains comments `isort:skip_file`, `SortImports` instance doesn't have the `output` property.
This behavior is incompatible with the file-name-based skipping which returns `None`, and could cause `AttributeError`.

I added simply `self.output = None` to fix it.


timothycrosley:
This looks great, thank you!

PR #619, opened 2017-12-12
Drop support for Python 3.3.

h-hirokawa:
The End-of-life of Python 3.3 is [already over](https://devguide.python.org/#status-of-python-branches), and Pytest >= 3.3.0 no longer supports 3.3.
To prevent build errors in Travis, it is better to stop supporting the version even in isort.


hugovk:
Here's the pip installs for isort from PyPI for the last month (via `pypinfo --percent --pip --markdown isort pyversion`), with low numbers for 3.3:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   52.9% |        314,042 |
| 3.6            |   25.0% |        148,089 |
| 3.5            |   14.3% |         84,643 |
| 3.4            |    6.9% |         41,234 |
| 3.7            |    0.4% |          2,159 |
| 3.3            |    0.3% |          1,765 |
| 2.6            |    0.3% |          1,503 |
| 3.2            |    0.0% |             24 |
| None           |    0.0% |              2 |

timothycrosley:
Agreed that it should be dropped; merge requests looks good!

Thanks!

PR #623, opened 2018-01-09
minor things to improve usability 

bagerard:
While configuring isort as a pre-commit hook, I had some issue because I was misusing the command (I was using path with tilde to --virtual-env i.e: ~/dev/env) and it took me some debugging to find out why it wasn't sorting my imports correctly. In fact when providing a virtual_env directory that does not exist, it silently ignores it but this has a significant impact on the import sorting.

So I ended up implementing the following:
- Use abspath for virtual_env (to make it consistent with settings_path)
- print warning in case virtual_env or settings_path directories don't exist
- print 'Fixing {filepath}' whenever isort actually overwrites a file (I believe it is a good practice for such library and it's very helpful when used as a hook)

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -198,9 +198,10 @@ class SortImports(object):
         elif write_to_stdout:
             sys.stdout.write(self.output)
         elif file_name and not check:
+            if self.output == file_contents:
+                return
+
             if ask_to_apply:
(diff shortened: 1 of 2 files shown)


timothycrosley:
Awesome work, thanks!

PR #625, opened 2018-01-11
Fix 'lines_after_imports' detection

timstaley:
Fix a bug where variables are mistaken for function definitions. 

This fixes a bug wherein a variable definition prefixed ``default_`` would
be confused with a ``def foo()``, because they both start with ``def``.
(Whether it's a function def or a variable changes the
expected default lines-after-imports behaviour. Noticed this when
isort and yapf began clashing over these cases.)


timothycrosley:
looks great thanks!

PR #631, opened 2018-01-19
Don't add un-neccesary line break in grouped vertical grid.

obmarg:
Prior to this change, isort was always adding 1 to the current line length when
determining whether or not the current line had got too long when outputting a
vertical grid.

- For imports that are not last, this is fine.  We need to account for the comma
  after the import.
- For the final import in vertical grid mode this is also fine.  We need to
  account for the closing `)`.
- For the final import in the grouped vertical grid we put the closing ) on the
  next line, so adding 1 to the line length causes isort to add un-neccesary
  line breaks when the final import would only just fit onto the line by one
  character.

For example, with a line length of 20:

```
from x import (
    abcd, efgh, ijkl
)
```

will end up reformatted as

```
from x import (
    abcd, efgh,
    ijkl
)
```

This is causing problems for us, as we run yapf on our code as well as isort,
and our imports change depending on which tool ran last.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -693,12 +693,17 @@ class SortImports(object):
             "," if self.config['include_trailing_comma'] else "",
          )
 
-    def _output_vertical_grid_common(self, statement, imports, white_space, indent, line_length, comments):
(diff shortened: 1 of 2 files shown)


timothycrosley:
Looks great! Thanks!

PR #632, opened 2018-01-19
Smart newlines no longer skips multi-line string assignments

obmarg:
Prior to this change, the smart newline code was doing the wrong thing when
faced with code like this:

```
from x import y

something = """
test"""

class Blah():
    pass
```

The smart new lines code attempts to find the first construct after the import
block, and uses the `_skip_line` function to determine where that construct
starts.  However, `_skip_line` seems to incorrectly assume that a line ending
with `"""` is a docstring.   It then goes on to treat `class Blah` as the first
construct, and adds 2 new lines after imports instead of the 1 that should be
chosen.

I've fixed this by updating `_skip_lines` to stop iterating over the characters
of a line if it finds something that is neither whitespace, the start of a
string or a comment.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -762,6 +762,8 @@ class SortImports(object):
                         self._in_quote = line[index]
                 elif line[index] == "#":
                     break
+                elif line[index] not in (' ', '\t'):
+                    break
                 index += 1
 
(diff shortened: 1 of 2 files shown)


timothycrosley:
Looks good! Thanks!

PR #637, opened 2018-01-24
Fix 613 Addresses error recognizing "from .import foo"

akonsta:
I chose to replace " .import" with " . import" because without the initial space, the change might create an error when ".import" was intentional, for example:

from CoolPackage.import ExcellentClass


timothycrosley:
Looks great! Thanks!

PR #640, opened 2018-01-24
Fix 638

akonsta:
Allowed for "from.import foo"  by converting to "from . import" and added test for Issue 613 and a test for the new case added here.


akonsta:
Removed blank line in test causing a flake8 error.

timothycrosley:
Woah, this looks great! Thank you!

Issue #594, opened 2017-09-04
Incorrect grouping of from ... import ...


fritzo:
**Version:** isort 4.2.9

Isort incorrectly places `from six.moves import range` in a the package group, but correctly places `from six import add_metaclass` in the third party group. The following is from a package named `treecat`:
```py
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import itertools
import logging
from abc import ABCMeta
from abc import abstractmethod

import numpy as np
from scipy.special import gammaln
from six import add_metaclass

from six.moves import range  # <------------ This import is in the wrong group.
from treecat.structure import OP_IN
from treecat.structure import OP_OUT
from treecat.structure import OP_ROOT
from treecat.structure import OP_UP
from treecat.structure import TreeStructure
from treecat.structure import estimate_tree
from treecat.structure import make_propagation_program
from treecat.structure import sample_tree
from treecat.util import SERIES
from treecat.util import TODO
from treecat.util import jit
from treecat.util import parallel_map
from treecat.util import profile
from treecat.util import sample_from_probs
from treecat.util import set_random_seed
```

kchomski-reef:
I've switched to `4.2.9` and my output after running `isort t.py --force-single-line-imports` is as follows:
```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import itertools
import logging
from abc import ABCMeta
from abc import abstractmethod

import numpy as np
from six import add_metaclass
from treecat.structure import OP_IN
from treecat.structure import OP_OUT
from treecat.structure import OP_ROOT
from treecat.structure import OP_UP
from treecat.structure import TreeStructure
from treecat.structure import estimate_tree
from treecat.structure import make_propagation_program
from treecat.structure import sample_tree
from treecat.util import SERIES
from treecat.util import TODO
from treecat.util import jit
from treecat.util import parallel_map
from treecat.util import profile
from treecat.util import sample_from_probs
from treecat.util import set_random_seed

from scipy.special import gammaln
from six.moves import range  # <------------ This import is in the wrong group.
```
with a diff from your output being (`diff my.py gh.py`):
```diff
10a11
> from scipy.special import gammaln
11a13,14
> 
> from six.moves import range  # <------------ This import is in the wrong group.
27,29d29
< 
< from scipy.special import gammaln
< from six.moves import range  # <------------ This import is in the wrong group.
```

timothycrosley:
Being worked on by @jasonbrackman

[...]

PR #641, opened 2018-01-24
added a test for imports created at load time, such as six.moves.

jasonbrackman:
also added a simplified test of import being a module or package.
fixing #594.


timothycrosley:
Looks great, thank you!

PR #642, opened 2018-01-24
Fix606

akonsta:
Added parameter -k that will suppress default behavior of treating as imports (import Xxx as x) as superseding any other import of the same module, thereby removing it.   This parameter will keep all imports.

import os as _os
import os

By default, this will return:

import os as _os

With this parameter set, it will return

import os
import os as _os


timothycrosley:
Looks good, thanks!

Issue #575, opened 2017-07-11
Can't handle backslash and bracket


jirikuncar:
I have found a following imports that can't be automatically sorted:

```python
from sympy import S, pi, GoldenRatio as phi, sqrt
from sympy.ntheory.continued_fraction import \
    (continued_fraction_periodic as cf_p,
     continued_fraction_iterator as cf_i,
     continued_fraction_convergents as cf_c,
     continued_fraction_reduce as cf_r)
from sympy.utilities.pytest import raises
```

Results to:
```python
from sympy import GoldenRatio as phi
from sympy import S, pi, sqrt
from sympy.ntheory.continued_fraction import \
    continued_fraction_periodic as cf_p
from sympy.utilities.pytest import raises

     continued_fraction_iterator as cf_i,
     continued_fraction_convergents as cf_c,
     continued_fraction_reduce as cf_r)
```

katlings:
Working on a fix for this at Pycascades sprints!

jirikuncar:
@katlings thanks!

PR #644, opened 2018-01-24
Handle parenthetically enclosed imports starting after an escaped line

katlings:
Fixes issue #575


timothycrosley:
Looks great thank you!

PR #645, opened 2018-01-25
Correct typo in usage text

gaul:


timothycrosley:

PR #648, opened 2018-02-01
Don't attempt to process files we can't open

rouge8:
If a file does not exist (e.g. a symlink to a file that does not exist yet), don't attempt to sort it.


timothycrosley:
Looks good, thanks!

PR #655, opened 2018-02-01
Version bump

Plazmaz:
This caused me more headache than I'd like to admit.
EDIT: This is noted as "In progress" on the changelog, but this is the version on pypi.


timothycrosley:
Looks good! And sorry for this confusion! I'll be working to update everything tonight

Issue #671, opened 2018-02-11
isort is corrupting CRLF files


jchv:
With Python 3.6.3 and isort 4.3.3, my CRLF files are being corrupted. Every line ends with CR CR LF instead of CR LF; a lot of software treats this sequence differently, but it usually manifests as either two line endings or one line ending and an excess CR. Needless to say, this is painful once it occurs, leaving a lot of a codebase damaged. :)

The problem is really obvious now that I look, so I'll submit a PR. Basically:

**1.** CRLF is applied upon joining the lines.
```python
        self.output = self.line_separator.join(self.out_lines)
```

**2.** CRLF is applied again upon saving the file. The `newline` argument seems to blindly replace all instances of `\n` with whatever you stick in it, even though the lines already end in `\r\n`.
```python
            with io.open(self.file_path, encoding=self.file_encoding, mode='w', newline=self.line_separator) as output_file:
                print("Fixing {0}".format(self.file_path))
                output_file.write(self.output)
```

Kind of off-topic: anyone have any idea why this is broken now? It seems really unlikely to me that it literally never worked, and yet since the PR that was supposed to fix line endings, it doesn't seem the code has changed dramatically. My best guess is that Python's behavior for `newline` used to have heuristics to prevent this, but that seems asinine...

timothycrosley:
I'm not sure why this broke, or why it ever worked, but I've taken your fix and made a hotfix release with it - I appreciate you finding and fixing the issue! I've added a test to hopefully ensure it won't happen again: https://github.com/timothycrosley/isort/commit/682c2ea44dc36edab3eaa66e7e55449c95b55dc7

jchv:
👍 I didn't even think to make a test, but the care is appreciated.

PR #672, opened 2018-02-11
Do not apply line endings twice. (#671)

jchv:
At least on some occasions, setting `newline` to whatever the line endings currently are can cause Python to replace the `\n`s with whatever you pass in, causing lines to be encoded twice. This leads to CRLF files containing `\r\r\n` sequences in place of `\r\n`.

Fixes issue #671.


timothycrosley:
Looks good, thank you!

PR #683, opened 2018-03-03
Use dict comprehension instead of a dict() constructor

jdufresne:
The dictionary comprehension syntax is available on all supported Python versions. Dictionary comprehension is modern Python syntax and is always faster than the `dict()` constructor, e.g.:

```
  $ python3 -m timeit 'dict((a, a) for a in range(20))'
  100000 loops, best of 3: 2.16 usec per loop
  $ python3 -m timeit '{a: a for a in range(20)}'
  1000000 loops, best of 3: 0.953 usec per loop
```

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -416,8 +416,11 @@ class SortImports(object):
                                 self.remove_imports]
 
             sub_modules = ['{0}.{1}'.format(module, from_import) for from_import in from_imports]
-            as_imports = dict((from_import, "{0} as {1}".format(from_import, self.as_map[sub_module])) for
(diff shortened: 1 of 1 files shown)


timothycrosley:
Looks great!

mxr:
You could add [flake8-comprehension](https://pypi.python.org/pypi/flake8-comprehensions) to the project to help enforce this programmatically. I think you would add it [here](https://github.com/timothycrosley/isort/blob/4bcf991723af2b7cf78f27145264842aa895f578/tox.ini#L16).

PR #684, opened 2018-03-03
Skip installing isort when running the "lint" tox environment

jdufresne:
isort is not a dependency to install & run flake8. Use the `skip_install` configuration option for a slightly faster build.

For details on skip_install, see:

https://tox.readthedocs.io/en/latest/config.html#confval-skip_install=BOOL


timothycrosley:
+1

PR #687, opened 2018-03-15
Handle non-int `indent_size` in .editorconfig

RobertDeRose:
Editorconfig allows the word `tab` to be used for `indent_size`, which
will result in it using the value from `tab_width`. However, this was
causing a `ValueError` because this condition was not handled.


timothycrosley:
Thanks for finding and fixing this! Great work!

PR #695, opened 2018-04-16
Allowed grid mode to fall back to hanging grid

NeilGirdhar:
When there are very long package or module names, the grid mode sometimes outputs lines longer than line_length.   We now fall back to mode 4 if it's more compact or if mode 0 violates the line_length.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -503,6 +503,16 @@ class SortImports(object):
 
                     if do_multiline_reformat:
                         import_statement = self._multi_line_reformat(import_start, from_import_section, comments)
+                        if self.config['multi_line_output'] == 0:
+                            self.config['multi_line_output'] = 4
+                            try:
+                                other_import_statement = self._multi_line_reformat(import_start, from_import_section, comments)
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
See https://github.com/timothycrosley/isort/issues/688

NeilGirdhar:
@timothycrosley Would you mind taking a look at this?  I'll fix the test if it's likely to be accepted.

timothycrosley:
I think this looks great! As soon as the tests pass I will merge. Sorry for not responding sooner.

NeilGirdhar:
What is the correct behavior for the failing test?

timothycrosley:
Hi @NeilGirdhar!

Looks like assertion just needs to be changed to include the fallback to hanging grid for that one test case with a very long comment. Since it's trying to test a very long comment it's probably best to change it in that direction rather than make the line length longer.

NeilGirdhar:
@timothycrosley My pleasure.  Thank you for your great project.

PR #700, opened 2018-04-28
Update line length docs to match the code

jacebrowning:
According to the code, the default line length is 79: https://github.com/timothycrosley/isort/blob/7e2fb3b21b4b588dc7a62ac65693781ae39e8a98/isort/settings.py#L52


timothycrosley:
Thanks!

PR #702, opened 2018-05-03
Remove '__init__.py' from skipped list.

briabar:
I fixed issue 690 by removing __init__.py from the default ignore list.


timothycrosley:
Great work, thank you!

PR #703, opened 2018-05-04
Support path patterns for known sections to find sub packages

### LINKED ISSUE #347: Wildcards for import groups
Hi,
I haven't managed to get this to work, wondered if there's a way.
Using isort, with the config in an editorconfig file.
Use case:
In django, one ends up having multiple apps/modules inside the project. A common way to organize these, is to use an `projectName_*` prefix (i.e. something_accounts, something_api, etc...). I would like to have all these sorted by isort within a section (first_party or named section), but the only way i've managed to do this is by referencing them explicitly one by one in editorconfig.
Can i not simply do something like `first_party=something_.\S*`?
It would make things much easier and not have update the config everytime a new project app is created.

### LINKED ISSUE #692: Support paths which are known to be in a particular section
isort currently supports specifying that a particular module is known to belong in a particular section.  But it would be great for us to be able to do the same for a particular path-prefix.  That is, if we do
```python
sys.path.insert(0, 'third_party')
import module1  # third_party/module1.py
import module2  # third_party/module2.py
```
currently we'd have to say `known_third_party=module1,module2` to get the right behavior; it would be great if instead we could say instead `known_third_party_paths=third_party`.  This would match how virtualenv directories are [currently handled](https://github.com/timothycrosley/isort/blob/2071da8/isort/isort.py#L287).

lundberg:
Configured known patterns ending with a path separator gets expanded and replaced with sub packages. Probably solves a few different scenarios like grouping section for packages under path, pointing out first party folder etc.

**Related issues:** (?)
Fixes #347 
Fixes #692

**Example:**
Lets say the repo has a `src` folder with a few parallel packages.
``` sh
$ find .
./src
./src/foo
./src/foo/__init__.py
./src/foo/bar.py
./src/ham
./src/ham/__init__.py
./src/ham/spam.py
```

Instead of manually adding `foo`, `ham`, etc. as first party modules, the following *pattern* ...
``` sh
$ isort -p src/ foo/bar.py
```
... gets automatically expanded, as if you would have ...
``` sh
$ isort -p foo -p ham foo/bar.py
```

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -150,6 +150,11 @@ class SortImports(object):
             known_placement = KNOWN_SECTION_MAPPING.get(placement, placement)
             config_key = 'known_{0}'.format(known_placement.lower())
             known_patterns = self.config.get(config_key, [])
+            known_patterns = [
+                pattern
+                for known_pattern in known_patterns
+                for pattern in self._parse_known_pattern(known_pattern)
+            ]
             for known_pattern in known_patterns:
                 self.known_patterns.append((re.compile('^' + known_pattern.replace('*', '.*').replace('?', '.?') + '$'),
                                             placement))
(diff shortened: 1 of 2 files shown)


timothycrosley:
This is great work as always, thank you!

PR #707, opened 2018-05-20
Pass output to stdout even on skipped files

ambv:
This fixes current behavior where tools that accept output piped from isort sometimes get empty files, which can cause files to be deleted instead of left alone.


timothycrosley:
LGTM, thank you!

PR #709, opened 2018-05-27
Don't change Python2 str encoding on import

slai:
isort has a Python3 compatibility module that, among other things, changes the Python environment when imported. These changes aren't restricted to the module importing though; they apply to all code running in the environment.

This is fine when executing isort as a command. However, this leads to weird issues when isort is used as a library, e.g. pytest-isort, because once loaded, it changes the behaviour of Python from that point onwards.

This led to an interesting issue where, depending on if pytest-isort (which imports isort) is installed, UnicodeDecodeErrors are or aren't raised on Python 2.7 due to the sys.setdefaultencoding call. See https://github.com/moccu/pytest-isort/issues/10

I don't think isort should be changing the Python environment through the simple act of importing the code - it's totally unexpected.

In this PR, I've extracted the code that changes the Python environment into a separate function that's only called if isort is executed as a command.

An unfortunate side-effect of this PR is that for some users, this may lead to new exceptions now that the default encoding is no longer being changed (arguably it should not be changed at all - https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script).

### DIFF
--- a/isort/__main__.py
+++ b/isort/__main__.py
@@ -1,5 +1,8 @@
 from __future__ import absolute_import
 
+from .pie_slice import apply_changes_to_python_environment
+apply_changes_to_python_environment()
+
 from isort.main import main
 
 main()
--- a/isort/__main__.py
+++ b/isort/__main__.py
@@ -3,6 +3,6 @@ from __future__ import absolute_import
 from .pie_slice import apply_changes_to_python_environment
 apply_changes_to_python_environment()
 
-from isort.main import main
+from isort.main import main  # noqa: E402
 
 main()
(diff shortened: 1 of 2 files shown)


timothycrosley:
The last major release of Python 2 was ~8 years ago, and the first release of Python3 was ~10 years ago. The stackoverflow question, and all implications within was definitely known about when the hack was put in place, the reality though, is that supporting Python2 at all is basically a hack. Whatever makes it the least effort hack is all we can aim for since Python3 and the future is what we are building for. I'll merge this pull request in for the next release, but just be aware that the next major isort release will almost certainly have no Python2 support at all.

PR #715, opened 2018-06-22
requirements.txt and Pipfile support

### LINKED ISSUE #459: Allow to give a requirements.txt as parameter to set know third party
When developing using docker or vagrant my libs are installed inside the VM or container and I'm developing using my editor in my machine. Because of that isort can't say when libraries are third-party or not. Since I have a bunch of deps in my requiments.txt replicate all of those to the isort config file doesn't look to me as the best idea.

I'd like to have a requirements_file parameter that would parse the file and add the libs to my 3rd-party section.

orsinium:
1. Split `place_module` to separated classes named "finders". It's make developing and extending easier.
1. Move some operations from "parse_module" stage to "init".
1. Add finder for Pipfile via [requirementslib](https://github.com/sarugaku/requirementslib/) (part of [pipenv](https://github.com/pypa/pipenv/))
1. Add finder for requirements.txt via pip.
1. Both finders (requirements and pipfile) uses [pipreqs](https://github.com/bndr/pipreqs/) for converting PyPI package name to module name.

Fix #459

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.py[cod]
+__pycache__
 .DS_Store
 # C extensions
 *.so
@@ -25,6 +26,7 @@ npm-debug.log
 
 # Unit test / coverage reports
 .coverage
+.pytest_cache
 .tox
 nosetests.xml
 htmlcov
--- a/README.rst
+++ b/README.rst
@@ -86,11 +86,23 @@ Installing isort is as simple as:
 
     pip install isort
 
-or if you prefer
+Install isort with requirements.txt support:
 
 .. code-block:: bash
 
-    easy_install isort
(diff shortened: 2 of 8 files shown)


orsinium:
@timothycrosley, can you review my pull request?

orsinium:
@timothycrosley , do you alive?

timothycrosley:
Sorry for the long delay, reviewing now and it looks great! Merging, thanks for the great work

n6g7:
Can't wait to use that!
@timothycrosley We can we expect a new release? 🙂

orsinium:
@n6g7, as temporary solution you can use my fork:
```
pip install isort-plus
```

PR #716, opened 2018-06-28
Add testing and document support for Python 3.7

jdufresne:
Python 3.7 was released on June 27, 2018.

https://docs.python.org/3/whatsnew/3.7.html


cclauss:
@jdufresne Can you please incorporate the changes from #731 so that we can move from Python 3.7-dev to Python 3.7 production release?

jdufresne:
Done! Thanks for the review and suggestion.

.travis.yml
```diff
@@ -4,8 +4,7 @@ cache: pip
 matrix:
   include:
   - env: TOXENV=isort-check
-  - python: 3.6
-    env: TOXENV=lint
+  - env: TOXENV=lint
```
cclauss: I would feel more comfortable if __isort-check__ and __lint__ were run on Python 3.7 instead of 2.7.  This was done in #731 and the tests pass.  There are several keywords and built-ins like __async__, __basename__, raw_input__, __unicode__, __xrange__ that we want the lint step to ensure are handled properly.  Python 3 is almost always more strict in this regard than Python 3.

.env
```diff
@@ -16,6 +16,8 @@ if [ ! -d "venv" ]; then
      if ! hash pyvenv 2>/dev/null; then
         function pyvenv()
         {
+            if hash pyvenv-3.7 2>/dev/null; then
+                pyvenv-3.7 $@
             if hash pyvenv-3.6 2>/dev/null; then
```
cclauss: This could be __elif hash pyvenv-3.6__ instead of __if hash pyvenv-3.6__ but it is minor thing.

.travis.yml
jdufresne: In tox.in, `lint` is defined with `basepython = python2.7`. So it doesn't matter what the default system Python is, tox will always use `python2.7`. Therefore, I don't think changing the python version here adds any value.

https://github.com/timothycrosley/isort/blob/fcd80d46e66179f9464c2df745ffed07622d9ee3/tox.ini#L14-L15

Here is the tox documentation on the configuration option:

https://tox.readthedocs.io/en/latest/config.html#confval-basepython=NAME-OR-PATH

Removing the `basepython` causes the tox target to fail for me. So I don't think the lint target is compatible with Python 3 yet.

jdufresne:
Thanks again for the review. Updated the `.env` script to use `elif`.

cclauss:
#732 shows the changes required to make the lint step work under Python 3.  It is all about applying the Python porting best practice "Use feature detection, not version detection".

jdufresne:
Thanks @cclauss

I'd prefer to handle the changes to lint under Python 3 as a separate PR. I feel they are distinct enough from the goals of merely adding Python 3.7 testing that they can be reviewed and land separately. If they land first, I'll happily rebase this PR and adjust it accordingly.

PR #730, opened 2018-08-02
Make known path patterns relative to config file containing the setting

### LINKED ISSUE #729: Relative known path patterns not found when running isort in sub package dir
Directory paths in i.e. `known_first_party` gets expanded to found packages within that dir.
If such a path is relative i.e. `foo/`, it gets by default resolved against current working dir.

In some cases, like for me in `vim`, isort gets executed with working dir relative to the actual file being processed. This causes the path/package expansion to fail due to the configured `known_first_party` path is set to be relative to the project/source root dir.

lundberg:
Fixes #729 

To solve this, easiest would be to walk the path up looking for `__init__.py` files until not found anymore and that way find the source root, but since this is python 2 and 3, init files may not be present.

This pull request uses the directory of the config file, setting the known paths as the root dir.

**Example:**
Lets say we have a project in directory `/tmp/myapp` with a `src` dir and a config file:
``` yaml
tmp/
 - myapp/
   - setup.cfg
   - src/
      - foo/
        - __init__.py
        - bar.py
```

``` ini
# setup.cfg
[isort]
known_first_party = src/
```

Executing `isort bar.py` from `/tmp/myapp/src/foo` will make the path in `known_first_party` absolute while reading the config file `setup.cfg`.
``` py
>>> settings['known_first_party]
['/tmp/myapp/src/']
```

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -172,6 +172,7 @@ def _update_settings_with_config(path, name, default, sections, computed_setting
 
 
 def _update_with_config_file(file_path, sections, computed_settings):
+    cwd = os.path.dirname(file_path)
     settings = _get_config_data(file_path, sections).copy()
     if not settings:
         return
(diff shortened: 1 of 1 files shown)


lundberg:
Any thoughts on this PR @timothycrosley or ready for merge?

timothycrosley:
This looks great, sorry for the long review delay, merged

PR #734, opened 2018-08-13
Make formatting import compatible with black check

michal800106:
When the very long line is met isort should wrap import into one of
multi line outputs. Black says that vertical hanging indent should be
choosen. An for most cases it works correctly. But when only one
import is on the list there is some difference.
Black suggests that there shouldn't be a trailing comma which isort is
giving.


timothycrosley:
Great work, thanks!

earlzo:
@timothycrosley waiting for a release, thks

PR #747, opened 2018-09-13
explicitly state the section is "isort" and add support for "tool:isort"

terencehonles:
I wasn't sure if the section was explicitly "isort" or something else. I added styling to make that obvious and added support for using "tool:isort" as the section name.


Congee:
the README says
```
Or, if you prefer, you can add an isort or tool:isort section to your project's setup.cfg or tox.ini file with any desired settings.
```

just tested with the latest 4.3.4, isort `tool:isort` does not work with isort but `[isort]` is detected.

Please fix this. @timothycrosley

Issue #737, opened 2018-08-19
isort does not support unix paths in config when run on Windows


jayvdb:
The following config does not work on Windows 
```ini
[isort]
skip=tests/test_config.py,tests/test_integration.py
```

I raised this previously as https://github.com/timothycrosley/isort/issues/419 , and it doesnt appear to work for 4.2.8, or 4.3, or in latest.

timothycrosley:
This is fixed in the release to be distributed to PYPI today,

Thanks!

~Timothy

PR #768, opened 2018-11-06
Fix Windows path issue

Dr-Irv:
Fixes issue #737 with using Unix-style paths in the skip list when running isort on Windows.

Tested via the settings used on pandas.


timothycrosley:
Awesome, thank you!

PR #775, opened 2018-11-15
Maintain order when loading iterable values from config files

### LINKED ISSUE #774: isort is non-deterministic with multiple forced_separate groups
`forced_separate` groups that are specified in a config file can result in non-deterministic sorting. For example:

The following configuration in a setup.cfg, tox.ini etc

```
[isort]
forced_separate:
    separate1
    separate2
```

Should result in:

```
import time

from separate1 import foo

from separate2 import bar
```

But due to the use of `set`  [here](https://github.com/timothycrosley/isort/blob/b5d7cd20696b7deed47cca3f2888c98fe34e5d50/isort/settings.py#L204), sometimes the order is switched to:

```
import time

from separate2 import bar

from separate1 import foo
```

mattbennett:
Fixes #774.


peterdemin:
Is there anything I can do to facilitate merge of this PR?

timothycrosley:
Looks great! Thank you!

PR #777, opened 2018-11-21
Add option to ignore trailing comments

### LINKED ISSUE #776: Do not count comments into line_length
`import os # pylint: disable-unused`

from above example, we need an option to not count characters of comments

jinmel:
Fixes #776


timothycrosley:
Looks great, thank you!

PR #781, opened 2018-11-27
Replace custom OrderedSet class with ordered-set package

jdufresne:
Reduce code to maintain by leveraging existing ordered-set package. The
package is based on a recipe originally posted to ActiveState Recipes by
Raymond Hettiger, and released under the MIT license.

https://pypi.org/project/ordered-set/


jdufresne:
I think it'd be better off just using `OrderedDict` as a set.

PR #782, opened 2018-12-02
Make isort run with Python 3 only

jdufresne:
As isort is primarily used as a command-line tool, it is in a good position to move to Python 3 only. Other command-line-only tools have begun moving to Python 3. For example: black, mypy, pylint, and pypinfo. It will continue to support formatting Python 2 code, so users can still format their Python 2 projects.

Allows for many code cleanups. By removing these workarounds, the project will be easier to maintain and reduce testing resources.

Python 2 is scheduled to be EOL on January 1, 2020. The Python community is moving on.

https://www.python.org/dev/peps/pep-0373/

For reasons to drop Python 2 support, see:

https://python3statement.org/

For an overview of projects that have dropped Python 2 support entirely, see:

https://hugovk.github.io/drop-python/2.7/


jdufresne:
@timothycrosley I see you gave a thumbs up, but I wanted to get your feedback on this before merging. Do you think this should be a part of the next release or the release after next? Curious if you think there is a need for another Python 2 release. If not, I'll happily merge. Thanks!

r-darwish:
Is there a list of tasks that need to be done? If so, I may might help with implementing some.

timothycrosley:
Hi @jdufresne,

First off - thanks for all the continued amazing work!
I do think one final Python2 release would be nice - followed immediately with this Python3 only implementation. I will make sure to have these out by the end of next week.

I'd also like to move to enable you, and other trusted contributors, to do releases as well in the future. 

Thanks again!

~Timothy

jdufresne:
Sounds good @timothycrosley Thanks!

I've rebased to resolve the merge conflicts, let me know if I can help in any other way.

Issue #763, opened 2018-10-26
Fix support for toml


maroux:
As mentioned in https://github.com/timothycrosley/isort/issues/705#issuecomment-423797271, the pyproject.toml file can't be read with config parser and must be read with `toml` library. It's broken in current form.

calvertvl:
One specific example that causes problems with the current implementation is a multi-line string in `pyproject.toml`, such as this example configuration for black:
```
[tool.black]
py36 = true
# include = '\.pyi?$' # the default
exclude = '''
/(
\.git
| \.mypy_cache
| \some-other-folder
| __pycache__
| _build
| build
| dist
)/
'''
```

ofek:
@timothycrosley @jdufresne Any progress on this?

PR #790, opened 2018-12-26
Fix reading quotes in pyproject.toml files

rr-:
This fixes the problem described here: https://github.com/timothycrosley/isort/issues/760#issuecomment-449881438


jdufresne:
According to issue #763, `pyproject.toml` will need to be parsed using a toml parser, not configparser. WDYT? If that is true, simply stripping quotes will not be enough.

rr-:
I have no opinion, I know it fixes my edge case but not sure about the others.
I can adjust the PR so that it uses `toml`, but this will add external dependancy to isort.

PR #806, opened 2019-01-23
Read configuration from XDG_CONFIG_HOME

r-darwish:
In addition to `~/.isort.cfg`, isort will try to read a configuration file from
`$XDG_CONFIG_HOME/isort.cfg`


r-darwish:
Waiting for #782 in order to use the xdg package.

jdufresne:
appdirs may be a suitable alternative to consider. https://github.com/ActiveState/appdirs

It is well established, cross platform (Windows), and supports all current Pythons.

r-darwish:
appdirs returns `~/Library/Application Support` in macOS instead of
`$XDG_CONFIG_HOME/isort.cfg`. Is that acceptable?

jdufresne:
I don't use macOS, but I gather that is the typical spot for configuration files. If so, then yeah, that makes sense to me.

r-darwish:
It's the typical spot for GUI applications. Command line applications such as fish or git still look
for their configuration file in `~/.config` in macOS. I'll do [this](https://github.com/ActiveState/appdirs/issues/78#issuecomment-366947045)

timothycrosley:
Looks good! Thank you!

PR #809, opened 2019-01-25
Test coverage improvements

akx:
* The main.py module had been left mostly untested as far as coverage showed, so this PR adds a test to see that the main bits in there don't smoke an awful lot when plugged in.
* There was needless duplication in the multiprocessing code, which is now neatly DRYed out.  
  In addition, use of `concurrent.futures` (3.2+) was replaced with `multiprocessing` (2.6+). (Not that it matters an awful lot with #782 pending, but still, it felt cleaner like this.)


timothycrosley:
This is great! Thank you!

PR #810, opened 2019-01-25
Refuse to work in system/.tox/venv/... directories by default

### LINKED ISSUE #759: isort can trash Python installations
Heya :)

This is a bit of a corner case, but it's happened to me more than once already...
So `isort` is prooobably not meant to be used on Python stdlib code, since it can be a little tricky with imports in places.

However, if you happen to have a virtualenv within your working directory, or a `.tox` directory containing those, `isort -rc .` may recurse down into them, like in the logs below.

This will be a problem in any environment where the python stdlib files are writable, as they are on MacOS when you've installed Python with Homebrew; running within a virtualenv doesn't matter, since these files are usually linked into it, not copied.

```
$ isort -rc .
Fixing ./.tox/flake8/bin/python-config
Fixing ./.tox/flake8/bin/activate_this.py
[...]
Fixing ./.tox/flake8/lib/python3.7/os.py
[...]
Fixing ./.tox/flake8/lib/python3.7/site.py
[...]
```

At this point, I hit ctrl+c, because this wasn't what I wanted, removed the `.tox` directory and tried to re-run `isort`:

```
$ rm -r .tox
$ isort -rc .
Fatal Python error: initsite: Failed to import the site module
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site.py", line 73, in <module>
    import os
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/os.py", line 30, in <module>
    from os.path import (altsep, curdir, defpath, devnull, extsep, pardir, pathsep,
ModuleNotFoundError: No module named 'os.path'; 'os' is not a package
$
```
Oh no! Remembering I'd seen this before, I tried to run a non-virtualenv `python3`:
```
~ $ python3
Fatal Python error: initsite: Failed to import the site module
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site.py", line 73, in <module>
    import os
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/os.py", line 30, in <module>
    from os.path import (altsep, curdir, defpath, devnull, extsep, pardir, pathsep,
ModuleNotFoundError: No module named 'os.path'; 'os' is not a package
```
It's dead, Jim 😞 

This is fixed by reinstalling Python (`brew reinstall python3` for me) to replace the isort-tainted stdlib modules with fresh ones.

I think it'd be worth it to add special-casing in `isort` to ignore (by default; could be overridden with some sort of footgun switch) files in:

* `*/.tox/`
* `*/lib/python*.*`
* any `site-packages` directory
* any `dist-packages` directory (Debian special)

Maybe some others?

akx:
This PR makes sure isort won't wander into directories where it shouldn't. By @jdufresne's suggestion, this list is based on that of [Black's](https://github.com/ambv/black/blob/024c9cab55da7bd3236fd88759c9735d6149b464/black.py#L53-L55), with my addition of "lib/pythonX.X", which appears within venvs/site-packages/etc., but should never appear elsewhere.

This behavior can be disabled with the `--unsafe` switch.

Fixes #759

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -285,11 +285,16 @@ def create_parser():
                         help='Tells isort to ignore whitespace differences when --check-only is being used.')
     parser.add_argument('-y', '--apply', dest='apply', action='store_true',
                         help='Tells isort to apply changes recursively without asking')
+    parser.add_argument('--unsafe', dest='unsafe', action='store_true',
+                        help='Tells isort to look for files in standard library directories, etc. '
+                             'where it may not be safe to operate in')
     parser.add_argument('files', nargs='*', help='One or more Python source files that need their imports sorted.')
 
     arguments = {key: value for key, value in vars(parser.parse_args()).items() if value}
     if 'dont_order_by_type' in arguments:
         arguments['order_by_type'] = False
+    if arguments.pop('unsafe', False):
+        arguments['safety_excludes'] = False
     return arguments
 
 
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -27,6 +27,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import fnmatch
 import io
 import os
+import re
 import posixpath
 import sys
 import warnings
@@ -47,6 +48,10 @@ except ImportError:
 
 MAX_CONFIG_SEARCH_DEPTH = 25  # The number of parent directories isort will look for a config file within
 DEFAULT_SECTIONS = ('FUTURE', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER')
+
+safety_exclude_re = re.compile(
+    r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|_build|buck-out|build|dist|lib/python[0-9].[0-9]+)/"
+)
 
 WrapModes = ('GRID', 'VERTICAL', 'HANGING_INDENT', 'VERTICAL_HANGING_INDENT', 'VERTICAL_GRID', 'VERTICAL_GRID_GROUPED',
              'VERTICAL_GRID_GROUPED_NO_COMMA', 'NOQA')
@@ -145,7 +150,9 @@ default = {'force_to_top': [],
            'show_diff': False,
            'ignore_whitespace': False,
            'no_lines_before': [],
-           'no_inline_sort': False}
+           'no_inline_sort': False,
+           'safety_excludes': True,
+           }
 
 
 @lru_cache()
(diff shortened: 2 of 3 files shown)


test_isort.py
```diff
@@ -2592,3 +2593,20 @@ def test_monkey_patched_urllib():
         # Previous versions of isort monkey patched urllib which caused unusual
         # importing for other projects.
         from urllib import quote  # noqa: F401
+
+
+@pytest.mark.parametrize('enabled', (False, True))
+def test_safety_excludes(tmpdir, enabled):
+    from isort import settings, main
```
jdufresne: Can you move the import to the top level, please?

isort/main.py
```diff
@@ -285,11 +285,16 @@ def create_parser():
                         help='Tells isort to ignore whitespace differences when --check-only is being used.')
     parser.add_argument('-y', '--apply', dest='apply', action='store_true',
                         help='Tells isort to apply changes recursively without asking')
+    parser.add_argument('--unsafe', dest='unsafe', action='store_true',
+                        help='Tells isort to look for files in standard library directories, etc. '
+                             'where it may not be safe to operate in')
```
jdufresne: Rather than implement of a new command line option could we adapt this to be the default value for `--skip` (same as Black's approach)?

isort/settings.py
```diff
@@ -48,6 +49,10 @@
 MAX_CONFIG_SEARCH_DEPTH = 25  # The number of parent directories isort will look for a config file within
 DEFAULT_SECTIONS = ('FUTURE', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER')
 
+safety_exclude_re = re.compile(
+    r"/(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|_build|buck-out|build|dist|lib/python\d.\d)/"
```
jdufresne: I guess this assumes there will not be a Python 3.10. Is that documented somewhere as the case?

`\d` includes all Unicode digits so I find `[0-9]` a safer alternative.

https://docs.python.org/3/library/re.html#index-27

test_isort.py
akx: Sure thing.

isort/main.py
akx: I decided against that, because many current projects (all I currently have, more or less) have a custom `skip =` stanza in the configuration, so they wouldn't benefit from this at all.

isort/settings.py
akx: Oh, hey, yeah, good point. I didn't think about 3.10+ at all.

timothycrosley:
Looks good, thank you!

PR #813, opened 2019-02-12
Handle third-party packages ending with EXT_SUFFIX

rhwlo:
Thanks for taking the time to review my PR!

## Summary
- Adds functionality to the `PathFinder` to allow it to find Python modules ending with the `EXT_SUFFIX`
- Adds tests for the PathFinder in `test_isort.py`

## Motivation
I noticed the other day that isort was incorrectly sorting https://github.com/python-rapidjson/python-rapidjson into the "first party" section. It turns out that the compiled module doesn't exist as `.../site-packages/rapidjson.so` but has a more complicated suffix.


timothycrosley:
This is awesome! Thank you!

Issue #770, opened 2018-11-09
Editing the file in the git hook


Vayel:
Currently, the git hook ([hooks.py:51](https://github.com/timothycrosley/isort/blob/develop/isort/hooks.py#L51)) only checks whether the imports are sorted. It would be nice to have the same behaviour as with [Black](https://github.com/ambv/black/blob/master/black.py#L294):

* Return a non-zero code if there are errors (non-sorted imports)
* Fix the errors (but let the user do `git add` again)

PR #817, opened 2019-02-19
Add option for git_hook to modify files

bsamseth:
With reference to #770, this adds an option to the ``git_hook`` function to allow it to not only display warnings when sources are unsorted, but also fix the sources.

This addition is off by default, so no existing usage will be affected.

I would still argue that ``modify=True`` should be the default, to be in line with the tagline "_isort your python imports for you so you don't have to_". This way, the user doesn't even have to do the sorting, just simply re-staging the files. But I guess this will come at a cost of breaking backwards compatability (hence the per now default of ``modify=False``).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -584,10 +584,11 @@ To cause the commit to fail if there are isort errors (strict mode), include the
     import sys
     from isort.hooks import git_hook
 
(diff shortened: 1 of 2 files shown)


timothycrosley:
This looks great, thank you!

PR #832, opened 2019-02-25
Add typechecking with mypy to Travis

mkurnikov:


isort/main.py
```diff
@@ -23,6 +23,7 @@
 import os
 import re
 import sys
+from typing import Any, List  # noqa: F401
```
mkurnikov: `# noqa: F401` is here because of inability of `flake8==3.5.0` to parse type comments.

timothycrosley:
This is awesome, thank you!

PR #843, opened 2019-02-25
Added NOQA to lines to be skipped 

richardlthomas:
Expected behavior is now that all isort:skip and NOQA imports will be moved to bottom, unsorted.


timothycrosley:
This looks great! Thank you!

Issue #828, opened 2019-02-25
Update isort distribution to use twine


timothycrosley:
Currently, when pushing to PYPI, a deprecation warning is generated recommending we switch to using twine.

PR #845, opened 2019-02-26
Changed distribution function to upload via Twine per issue #828

richardlthomas:
Fixes issue #828


orlnub123:
The `--repository-url https://test.pypi.org/legacy/` should probably be removed. I don't think distributing via the PyPI test server is a good idea.

PR #850, opened 2019-02-26
Split SortImports class into smaller classes, part 1

mkurnikov:
I want to start working on a https://github.com/timothycrosley/isort/issues/818, but decided to refactor the project a bit to make it a little more extensible. 

Diff ended up being too big, and current develop won't merge cleanly in new branch at all, so I cherry-picked part of it which is mergeable.

And in general, what do you think about it? Is it a good idea? I want to split big `SortImports` object into a number of smaller ones to
1. Reuse parsing logic for function / if-block scoped imports.
2. Make it a little more maintainable, my PyCharm can't even index `isort.py` properly, it's too big and complex.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -111,12 +111,7 @@ class SortImports(object):
         if self.config['line_ending']:
             self.line_separator = self.config['line_ending']
         else:
-            if '\r\n' in file_contents:
-                self.line_separator = '\r\n'
-            elif '\r' in file_contents:
-                self.line_separator = '\r'
-            else:
-                self.line_separator = '\n'
+            self.line_separator = utils.infer_line_separator(file_contents)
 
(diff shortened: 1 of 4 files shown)


mkurnikov:
Slept on it, I think one big PR with refactoring is not a good idea. I'll prepare a number of smaller less invasive ones, with additional test coverage.

timothycrosley:
I love the ideas represented here! isort right now has way too much functionality tucked away in an an `__init__` method within a single `isort` module. I'm going to take a bit of time merging it just because I think I need to get at lost one additional hotfix in for the 2.7.x supported release before I go full steam ahead on isort 5.0.0 - as much as I want to do that!

timothycrosley:
This looks great, thank you!

PR #854, opened 2019-02-27
Fix bug with parsing WrapModes from config file

mkurnikov:
Currently `isort` crashes trying to parse `multi_line_output` from a config file due to the fact that `WrapModes(str)` is not a valid `str -> type instance` callable.


timothycrosley:
Looks great, thank you!

PR #856, opened 2019-02-27
Add LRU cache to RequirementsFinder._get_files_from_dir

Tenzer:
This caches the result of the function for future invocations with the same
input, which can massively speed up future invocations if they have the same
input as previous calls.

In the case of this function this happens a lot, as it always is tried with all
parent paths in the file system from `isort` starting point.

The "backports.functools_lru_cache" package is used for Python 2 as it doesn't
have the built in `lru_cache` decorator which was added in Python 3.2.

Some benchmarks against a folder with 182 Python files with all finders enabled:

Python 2.7.15: Before 32 seconds, after 4 seconds.
Python 3.7.2: Before 28 seconds, after 2 seconds.


isort/finders.py
```diff
@@ -35,6 +35,11 @@
 except ImportError:
     Pipfile = None
 
+try:
+    from functools import lru_cache
+except ImportError:
```
jdufresne: The `develop` branch dopped support for EOL Pythons. So maybe this should target that branch and then we wouldn't need to includ the `backports` dependency.

Tenzer: I specifically targeted this branch in order to get a fix out in the 4.3.x series which still support Python 2. The fix can easily be ported to the develop branch afterwards.

jdufresne: Gotcha. Sounds good to me.

timothycrosley:
This is great thank you!

PR #860, opened 2019-02-28
allow named multi_line_import arguments

anderson-dan-w:
Allows users to supply something like `multi-line=VERTICAL_HANGING_INDENT` rather than `multi-line=3` on the command line (and in `.isort.cfg`). Happy to add tests if required, just struggled a little to find where they might go inside `test_isort.py`.


timothycrosley:
This is awesome, thank you!

PR #863, opened 2019-03-01
tests: fix default_settings_path

blueyed:
Restore `chdir` after tests.

This is relevant for when pytest-cov would display coverage in the end,
and would use absolute paths due to this.


timothycrosley:
This is great

Thank you!

Issue #417, opened 2016-04-01
Relative import sorting order 


lee-kagiso:
Is there a setting to change the sort order of relative imports from

``` python
from .. import x
from . import y
```

to

``` python
from . import y
from ..import x
```

so that one can comply with the Google style guide as per this linter https://github.com/public/flake8-import-order#configuration?

timothycrosley:
Hi @lee-kagiso, 

Thanks for bringing this to my attention! This is not something isort currently supports, but I'll work on adding support for this in the next release.

Thanks!

~Timothy

timothycrosley:
This is fixed as the default (no config necessary) and will be pushed into a new release today.

Thanks!

~Timothy

jpgrayson:
Let me start by saying thank you for `isort`. It has made a positive impact on my personal and team's workflows!

Although the referenced `flake8-import-order` plugin may reject the descending order for relative imports, AFAICT the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html) makes no assertions about order of relative imports. In fact, it prohibits using relative imports at all:

> Do not use relative names in imports. Even if the module is in the same package, use the full package name. This helps prevent unintentionally importing a package twice.

To be clear, by descending order, I mean:
```python
from ...baz import x
from ..foo import y
from .bar import z
```

There are at least two arguments for `isort` to default to descending order for relative imports:

1. The new ascending order is a regression from `isort`'s previous behavior for imports of the form `from ..module import name`, which more common in my projects than the `from .. import module` form.
2. The spirit of PEP-8 is to order imports from "least local" to "most local". Imports from `...` are less local than those from `..`, and imports from `.` are most local. Descending relative imports achieves this ordering.

The original intent of this issue was to add an option to select the order of relative imports. I suggest that such an option remains a valid request for `isort`.

jaap3:
I agree with @jpgrayson, with the recent release of isort it suddenly started reversing imports that were perfectly fine before.

PR #867, opened 2019-03-01
Add -rr/--reverse-relative option (issue #417)

jpgrayson:
This new option changes the sort order of relative imports. The new default
sort order is from furthest to closest (most dots to least dots):

    from ... import a
    from ...foo import x
    from .. import b
    from ..bar import y
    from . import c
    from .baz import z

With the new -rr/--reverse-relative option, the order of the relative
import *levels* (inter-order) is reversed, but the ordering within each
relative import level (intra-order) remains the same (i.e. normal lexical
order).

    from . import c
    from .baz import z
    from .. import b
    from ..bar import y
    from ... import a
    from ...foo import x

Signed-off-by: Peter Grayson <pete@jpgrayson.net>


timothycrosley:
This is great, 

Thank you!

PR #871, opened 2019-03-03
ci: add codecov.yml

blueyed:
Ref: https://docs.codecov.io/docs/codecov-yaml

For comments: https://docs.codecov.io/docs/pull-request-comments

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  status:
+    project: yes
+    patch: yes
+    changes: yes
+
+comment:
+  layout: "header, diff"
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  status:
+    project: yes
+    patch: yes
+    changes: yes
```
blueyed: There might be a problem currently with codecov posting status updates currently, but in general those are useful (next to the CI statuses).

timothycrosley:
Looks great, thank you!

PR #875, opened 2019-03-04
Support integer values for `multi_line_output` from TOML files

Tenzer:
TOML is a typed file format so if you had `multi_line_output = 3` specified in
your `pyproject.toml` file you would get an error like this after the changes
in https://github.com/timothycrosley/isort/pull/860 were merged in:

> TypeError: getattr(): attribute name must be string

This change makes sure the attribute name is a string when passed to `getattr()`
to avoid errors like this going forward.


timothycrosley:
Looks great! Thank you!

PR #877, opened 2019-03-05
Lookup real path of virtualenv

### LINKED ISSUE #876: confused by symlinks pointing to virtualenv gives FIRSTPARTY not THIRDPARTY
On my box using isort 4.3.10

```
lrwxrwxrwx 1 brian brian 36 Aug  8  2018 /home/brian/.virtualenvs -> /home/brian/.local/share/virtualenvs
```

This appears to get isort confused, and it can incorrectly categorize third party imports as first party.

```
from-type place_module for django_cron returned FIRSTPARTY
```

If I redo with another virtual environment that doesn't have symlinks it shows:

```
from-type place_module for django_cron returned THIRDPARTY
```

Which I believe is correct.

I believe it is failing here: https://github.com/timothycrosley/isort/blob/f6ff653fbf149b67cef1e553940b788a9afa685e/isort/finders.py#L176

Where `self.virtual_env_src` contains `/home/brian/.virtualenvs/energy_rating/src/` and `prefix` contains `
/home/brian/.local/share/virtualenvs/energy_rating/src/django-cron`

I suspect the solution might be to apply `os.path.realpath()` on `self.virtual_env_src`, which would turn it into `/home/brian/.local/share/virtualenvs/energy_rating/src` and the condition will pass.

brianmay:
This ensures we don't get confused and incorrectly classify third party
libraries as first party libraries.

Fixes #876.

### DIFF
--- a/isort/finders.py
+++ b/isort/finders.py
@@ -140,6 +140,7 @@ class PathFinder(BaseFinder):
 
         # virtual env
         self.virtual_env = self.config.get('virtual_env') or os.environ.get('VIRTUAL_ENV')
+        self.virtual_env = os.path.realpath(self.virtual_env)
         self.virtual_env_src = False
         if self.virtual_env:
             self.virtual_env_src = '{0}/src/'.format(self.virtual_env)


timothycrosley:
This is great, thank you!

Issue #842, opened 2019-02-25
pie_slice in isort>=4.3.5 no longer sets sys.setdefaultencoding('utf-8')


ericboyangan:
Hi! We're experiencing a new error in a Python 2.7 lint test that I believe was caused by bumping our isort from 4.3.4 to 4.3.5 (by way of flake8-isort's call to SortImports with both `file_path` and `file_contents`).

I see that the 4.3.5 version of pie_slice.py no longer calls `sys.setdefaultencoding('utf-8')` in python 2; which means source code containing non-ASCII now errors with `UnicodeDecodeError` on [line 129 of isort.py](https://github.com/timothycrosley/isort/blob/a7007483856e0f06c9cfe715a335d22936e3b3c9/isort/isort.py#L129). In general, IIUC, it seems `file_contents` cannot be a string with non-ASCII characters.

Wondering what you think would be the best resolution for this issue? And thank you so much for the great tool all these years!

cleder:
``` File "/home/vagrant/venv/lib/python2.7/site-packages/isort/isort.py", line 108, in __init__
    if file_contents is None or ("isort:" + "skip_file") in file_contents:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 1988: ordinal not in range(128)```
isort==4.3.9

**Comment by timothycrosley** [comment]
This has been fixed in the latest 4.3.10 release.
https://github.com/timothycrosley/isort/releases/tag/4.3.10

Thanks!

~Timothy

**Comment by bcroq** [comment]
I still get @cleder 's error using isort 4.3.10:

```
  File "/home/bcq/.virtualenvs/.../lib/python2.7/site-packages/isort/isort.py", line 108, in __init__
    if file_contents is None or ("isort:" + "skip_file") in file_contents:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1050: ordinal not in range(128)
```

**Comment by bcroq** [comment]
@timothycrosley here is how I get this issue.

First, create a `test.py` Python file containing:
```python
# coding: utf-8

x = 'testé'
```

Then run `isort - < test.py`.

Here is what I get with isort 4.3.5 to 4.3.10:
```pytb
Traceback (most recent call last):
  File "/home/bcq/.../bin/isort", line 11, in <module>
    load_entry_point('isort', 'console_scripts', 'isort')()
  File "/home/bcq/git/isort/isort/main.py", line 331, in main
[...]

PR #883, opened 2019-03-06
work on file encoding detection

bcroq:
Here is a proposed patch for #842 that works on Python 2 and Python 3.
Tested with a UTF-8 encoded file and a latin-9 encoded file:
```console
$ isort - < test_utf8.py && echo ok
# coding: utf-8

x = 'testé'
ok
$ isort - < test_latin9.py && echo ok
# coding: latin9

x = 'testé'
ok
```
```console
$ isort test_utf8.py && echo ok
ok
$ isort test_latin9.py && echo ok
ok
```
Also tested with flake8-isort:
```console
$ flake8 --no-isort-config test_utf8.py && echo ok
ok
$ flake8 --no-isort-config test_latin9.py && echo ok
ok
```

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -109,8 +109,8 @@ class SortImports(object):
                             " or matches a glob in 'skip_glob' setting".format(file_path))
                     file_contents = None
             if not self.skipped and not file_contents:
-                    with io.open(file_path, 'rb') as f:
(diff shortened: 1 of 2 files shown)


timothycrosley:
This looks great, thank you!

PR #884, opened 2019-03-06
Do not print file names when in the quiet mode

emilmelnikov:
The current help for the `--quiet` option says that it enables "extra
quiet output, only errors are outputted".  However, isort still prints
`Fixing spam/eggs.py` messages to stdout.


timothycrosley:
Looks great, thank you!

PR #901, opened 2019-03-12
Make sort case-sensitive

ichorid:
Currently, `isort` ignores case when sorting modules. This behavior conflicts with some stricter sorting styles, like i.e. "cryptography" style. This style requires sorting to put upper-case starting modules first, lower-case modules second.
This PR fixes this issue by removing an unnecessary coercion to `.lower()` in `isort`'s core modules sorting routine.


ichorid:
Oops, someone already fixed that ))

PR #922, opened 2019-04-11
Update the pre-commit configuration for isort

asottile:
👋 hello, pre-commit maintainer here

A few small things to make this usable elsewhere :)

- `additional_dependencies: []` is the default
- `types: [python]` will limit `isort` to running on python code instead of every file in the repository

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,5 +2,5 @@
     name: isort
     entry: isort
     require_serial: true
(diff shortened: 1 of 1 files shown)


asottile:
Adding to our official list here!  I'll work on redirecting our current "mirror" of the repository here! https://github.com/pre-commit/pre-commit.github.io/pull/223

timothycrosley:
Thank you!

PR #926, opened 2019-04-15
Fix list passed from toml settings

### LINKED ISSUE #923: Using pyproject.toml with seed-isort-config results in a list, which breaks `_as_list`
I'm using isort as a pre-commit hook, in conjunction with seed-isort-config and a pyproject.toml in my root dir.

It seems that seed-isort-config has now gained support for toml, and is thus setting `known_third_party = ["dateutil", "pydantic", "requests", "urlpath"]`

I believe that isort is still expecting the value to be a comma-separated string, as I am getting the following traceback when trying to run it:

```
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/bin/isort", line 10, in <module>
    sys.exit(main())
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/lib/python3.7/site-packages/isort/main.py", line 342, in main
    config = from_path(arguments.get('settings_path', '') or os.path.abspath(file_names[0]) or os.getcwd()).copy()
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/lib/python3.7/site-packages/isort/settings.py", line 182, in from_path
    _update_settings_with_config(path, 'pyproject.toml', [], ('tool.isort', ), computed_settings)
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/lib/python3.7/site-packages/isort/settings.py", line 250, in _update_settings_with_config
    _update_with_config_file(editor_config_file, sections, computed_settings)
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/lib/python3.7/site-packages/isort/settings.py", line 297, in _update_with_config_file
    computed_settings[access_key] = union(existing_data, _abspaths(cwd, _as_list(value)))
  File "/Users/step7212/.cache/pre-commit/repou647ppru/py_env-python3.7/lib/python3.7/site-packages/isort/settings.py", line 321, in _as_list
    for item in value.replace('\n', ',').split(',')
AttributeError: 'list' object has no attribute 'replace'
```

Example `pyproject.toml`:
```toml
[tool.isort]
multi_line_output = 3
include_trailing_comma = true
force_grid_wrap = 0
use_parentheses = true
line_length = 88
combine_as_imports = true
known_first_party = ["mypackage"]
known_third_party = ["dateutil", "pydantic", "requests", "urlpath"]
```

Adding the following to the beginning of the function fixes the issue for me:

```python
def _as_list(value: str) -> List[str]:
    if isinstance(value, list):
        return value
```
Alternatively, we could strip the items like so:
```python
def _as_list(value: str) -> List[str]:
    if isinstance(value, list):
        return [item.strip() for item in value]
```
I can submit a PR but I'll wait for discussion on which method to go with, if either.

StephenBrown2:
Resolves #923 

When toml is used for config, and especially with the seed-isort-config pre-commit function, the setting gets set to a list of strings, rather than a comma-separated string. This checks that the value passed is already a list and if so strips each item to be safe, and returns that.

This does not update the type-hints in the signature, as the function isn't really supposed to expect a string, but it's just a safeguard in case.

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -316,6 +316,8 @@ def _update_with_config_file(
 
 
 def _as_list(value: str) -> List[str]:
+    if isinstance(value, list):
+        return [item.strip() for item in value]
     filtered = [
         item.strip()
         for item in value.replace('\n', ',').split(',')


timothycrosley:
Looks great, thank you!

PR #933, opened 2019-04-25
Extract some logic from __init__ into separate methods

mkurnikov:


mkurnikov:
Follow-up PRs are going to be:
1. Extract all logic related to file reading (and file skipping possibly) out of `_SortImports`. 
2. Move all file operations to `pathlib`.
3. Enable `strict_optional = True` for mypy. 
4. Instead of manually parsing and changing python file, use https://github.com/davidhalter/parso with their lossless syntax trees.

timothycrosley:
This sounds great, thank you!

PR #936, opened 2019-04-27
Move most of filesystem operations to pathlib

mkurnikov:


StephenBrown2:
This would need to import `pathlib2` [from pypi](https://pypi.org/project/pathlib2/) for Python <=? 3.4

mkurnikov:
https://docs.python.org/3/library/pathlib.html
Pathlib is there from 3.4,and `develop` dropped support for python2. 
So there shouldn't be a need for that. Only thing here's with, I guess, a bit different functionality of the library, I'll look at it today.

StephenBrown2:
Def agree with the docs, I saw that the Python 3.4 tests were failing but didn't look to see that it's throwing a FileNotFoundError error: https://travis-ci.org/timothycrosley/isort/jobs/525448405#L265

isort/format.py
```diff
@@ -27,24 +28,27 @@ def format_natural(import_line: str) -> str:
...
+        tofile=file_name + ':after',
+        fromfiledate=file_mtime,
         tofiledate=str(datetime.now())
     )
     for line in unified_diff_lines:
         sys.stdout.write(line)
 
 
-def ask_whether_to_apply_changes_to_file(file_path: str) -> bool:
+def ask_whether_to_apply_changes_to_file(fpath: str) -> bool:
```
timothycrosley: Curious, why the switch from `file_path` to `fpath`?

timothycrosley:
This is awesome thank you! For what it's worth, Python3.4 has reached end of life, and the as such the next release of isort does not need to support it.

Thanks!

~Timothy

isort/format.py
mkurnikov: I think I did it before extracting the function out to discern between variables. Doesn't really matter, can roll it back if you want.

timothycrosley: I went ahead and changed it to `file_path` just because of my personal preference to spell out words, rather than use abbreviations - but agree completely it doesn't matter much. Especially in the context of all the great and very welcome improvements in this pull request :).

Issue #735, opened 2018-08-14
The -k/--keep-direct-and-as flag does not work


user37596:
-k, --keep-direct-and-as
                        Turns off default behavior that removes direct imports
                        when as imports exist.
But when I have file `test.py`:

    from django.utils.translation import ugettext as _, ugettext

and run `isort -k test.py` or `isort --keep-direct-and-as test.py`, the result is:

    from django.utils.translation import ugettext as _

isort version is 4.3.4.

timothycrosley:
This is fixed in develop and will make it's way out in the 5.0.0 release.

Thanks!

~Timothy

Issue #908, opened 2019-03-22
isort with same imports but different alieces break imports 


tbicr:
just use `isort -rc file` for next kind of imports:

```
from datetime import datetime
import datetime.datetime as dt
```

but got next result:

```
import datetime.datetime as dt
from datetime import datetime as dt
```

eg. `datetime` now unavailable in module

timothycrosley:
This is fixed in develop and will make it's way out in the 5.0.0 release. Thanks!

~Timothy

PR #944, opened 2019-05-09
 keep-direct-and-as only inserts direct import if it was present in the original file

viourr:
This will fix issue #735 where the --keep-direct-and-as option would add a straight import, even if it wasn't in the original file.
This also fixes issue #606 (using the -k option).
It also seems to be one step closer on #908 but I wasn't concentrating on that issue.

In addition, this change allows the same object to be imported as multiple aliases; which, I agree, is poor style, but removing those aliases causes real crashes. It should probably be up to some other linter to shame the programmer about the double reference.


timothycrosley:
This is awesome, thank you!

PR #951, opened 2019-05-21
[WIP] Introduce parso as parser

mkurnikov:
It's a work-in-progress of rewriting `SortImports` object into parso-based one.

### DIFF
--- a/isort/compat.py
+++ b/isort/compat.py
@@ -3,11 +3,12 @@ import os
 import re
 import sys
 from pathlib import Path
+
 from typing import Any, Optional, Tuple
 
 from isort import settings
 from isort.format import ask_whether_to_apply_changes_to_file, show_unified_diff
-from isort.isort import _SortImports
+from isort.isort2 import _SortImports2
 
 
 def determine_file_encoding(file_path: Path, default: str = 'utf-8') -> str:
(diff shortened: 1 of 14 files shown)


mkurnikov:
1. I modified `parso` grammar file and tokenizer to make comment a first-class node via  `PythonNode(comment_stmt)`. 
2. AST is changed in-place. 

Algorithm is: 
1. Find out a position of first import to mark an end of the top-level comments. 
2. Extract all top-level imports and remove them from AST. 
3. Add new sorted imports starting from the "end of top-level comments" position. 

@timothycrosley I probably won't finish it, `isort` has too many different flags to change behaviour. I can work on different output modes, as it's very AST-transformations-heavy code, but someone else would have to bring it all together.

timothycrosley:
@mkurnikov, no worries! Thank you for the awesome work! Once you feel you've like you've gotten as far along the journey as you intend let me know, and I will do all I can to help push it over the finish line, either way I really appreciate the great work you put into this.

timothycrosley:
@mkurnikov,

I've done a lot of thinking about this and I've updated isort to support both of the requested features:

    Support sorting imports under non-global scopes.
    If there's a code inside the imports, we can disable reorganizing, and only sort alphabetically inside sections.

But I've done it without switching to AST style parsing. I realized that there are fairly strict requirements around import lines, making them very easy to identify with confidence separate from other code blocks, without using AST parsing. Doing so is also faster, and allows isort to remain immune to ast changes (such as the walrus operator, which broke other formatters if used across Python versions, but isort kept working). I'd be very happy to get your feedback and testing on this.

Thanks!

~Timothy

Issue #957, opened 2019-06-19
Long aliases and use_parentheses generates invalid syntax


mattbennett:
isort with `use_parentheses=True` inserts parens after the `as` if the alias is very long. 

Example:

``` python
from package import module as very_very_very_very_very_very_very_very_very_very_long_alias
```
results in
``` python
from package import (
    module as (
    very_very_very_very_very_very_very_very_very_very_long_alias))
```

This is not valid syntax.

EgorChernik:
The new issue comes))
<pre>from package (
    very_long_module_name as nice_alias,
)</pre>

`line_length` doesn't work anymore with that import-construction.

mattbennett:
@EgorChernik, in ad6af18 you've addressed the second point that I made in my PR: https://github.com/timothycrosley/isort/pull/958

For my usecase (isort combined with black), I actually don't want line-length to be respected in this case. I would argue that the reason `use_parentheses` exists is to avoid backslashes, so it's not completely obvious to me that this is something that needs fixing.

timothycrosley:
@mattbennett Thanks for reporting and fixing this issue!

I've pushed a hot fix release that includes your fix: https://github.com/timothycrosley/isort/releases/tag/4.3.21

@EgorChernik this is a good point, long term it may make sense to have a strict line length enforcement setting, to enable nested wrapping like this.

Thanks!

~Timothy

PR #958, opened 2019-06-19
Fix issue 957

mattbennett:
Fix for #957.

```
from package import module as very_very_very_very_very_very_very_very_very_very_long_alias
```
becomes
```
from package import (
     module as very_very_very_very_very_very_very_very_very_very_long_alias
)
```

This will be the output even if the `module as alias` line is longer than the maximum line length. There is another possibly valid output mixing backslashes and parentheses to attempt to stay below the line length:

```
from package import (
     module as \
     very_very_very_very_very_very_very_very_very_very_long_alias
)
```

I haven't implemented that here because it's not obviously better and I want to use `isort` in combination with [black](https://github.com/python/black), which does not use backslashes in this case.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -242,12 +242,15 @@ class _SortImports(object):
 
                     cont_line = self._wrap(self.config['indent'] + splitter.join(next_line).lstrip())
                     if self.config['use_parentheses']:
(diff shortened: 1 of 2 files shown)


timothycrosley:
This looks great! Thank you!

PR #963, opened 2019-07-03
Fix up typo, replace resovled with resolved

timgates42:


timothycrosley:
This looks great, thank you!

Issue #971, opened 2019-07-25
Wrap length taking number as string


jaswanth098:
Ref: https://github.com/timothycrosley/isort/blob/develop/isort/main.py#L287
I have been trying to wrap_length, but isort throws the following error and quits
```bash
TypeError: '>' not supported between instances of 'str' and 'int'
```

jaswanth098:
I have found the problem 
Ref: https://github.com/timothycrosley/isort/blob/develop/isort/main.py#L287

Zarathustra2:
@jaswanth098 @timothycrosley This can be closed as PR has been merged!

jaswanth098:
Yes, can be closed @Zarathustra2 closing it

jaswanth098:
When will be the nearest release?

PR #974, opened 2019-08-05
bug fix type int for wrap length

jaswanth098:
#971 
@timothycrosley This is totally a small fix could you check this


timothycrosley:

PR #975, opened 2019-08-08
Add NPM node_modules to default exclusion list

jdufresne:
Projects frequently have Python and JavaScript coexisting. NPM is the de
facto standard package manager for JavaScript. Ignore third party
packages by default so projects don't need configure this.


timothycrosley:
This is a good call! Thanks for adding it!

Issue #960, opened 2019-06-20
Respect .gitignore


fpuga:
To avoid (miss)configurations it will be awesome that `isort` ignore by default the patterns set in `.gitignore`.

Other projects like black [are also thinking](https://github.com/python/black/issues/475) on doing it.

mdagaro:
I'm interested in taking up this issue if it's okay @timothycrosley

timothycrosley:
@mdagaro, it would be very appreciated, thank you!

timothycrosley:
This is implemented in the develop branch (Thanks @mdagaro!) and will be deployed with the 5.2.0 minor release.

timothycrosley:
This was just deployed with version 5.2.0 of isort. Thanks! 

~Timothy

PR #976, opened 2019-08-09
Issue #960: Respect .gitignore

mdagaro:
Add flag -git/--exclude-gitignore to ignore files found in a .gitignore.

Things to consider would be the name of the flag (I chose what seemed the most reasonable to me), error checking if it is not a git repository and the phrasing of the help messages. ~~I should probably add a directory check for the gitignore.~~ I realized the code was already doing that, go me!

Another note, I have not tested this because I don't have any good directory structure set up to test this.


timothycrosley:
I love this idea! I'm gone to take some time tomorrow to test it, and assuming everything looks good I'll merge it in - a great thing to have for sure!

saroad2:
I think you can use [GitPython](https://github.com/gitpython-developers/GitPython) instead of a subprocess, but I'm not sure "check-ignore" is supported there yet.

PR #977, opened 2019-08-14
Specify targeted Python version to mypy

jdufresne:
Discovered a typing bug where a Path object was passed to
os.path.relpath(), but Path support was not added until Python 3.6.

https://docs.python.org/3/library/os.path.html#os.path.relpath

> Changed in version 3.6: Accepts a path-like object.


timothycrosley:
Thank you for finding and fixing this!

PR #982, opened 2019-08-14
Simplify subprocess use with high level run API

jdufresne:
Per the Python docs, subprocess.run() is the preferred interface for
running subprocess and replaces the older API (such as
subprocess.check_output())

https://docs.python.org/3/library/subprocess.html

> The recommended approach to invoking subprocesses is to use the run()
> function for all use cases it can handle. For more advanced use cases,
> the underlying Popen interface can be used directly.
>
>
> The run() function was added in Python 3.5; if you need to retain
> compatibility with older versions, see the Older high-level API
> section.

Allows for:

- Return a str from get_output() to decode once instead of every line of
  the result.

- Pass command arguments as a list of strings to ensure no possibility
  for unescaped shell injection.

### DIFF
--- a/isort/hooks.py
+++ b/isort/hooks.py
@@ -28,17 +28,18 @@ from typing import List
 from isort import SortImports
 
 
-def get_output(command: str) -> bytes:
+def get_output(command: List[str]) -> str:
     """
     Run a command and return raw output
 
     :param str command: the command to run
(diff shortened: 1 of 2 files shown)


timothycrosley:
Good call! Thanks for updating this!

PR #983, opened 2019-08-14
Introduce Black to automate code formatting

jdufresne:
**This should merely be considered a proposal. Let me know what you think. If there is not interest in this change, no worries.**

Use Black, the uncompromising Python code formatter, to format the project and simplify contributions. Contributors no longer need to think or care about the project's preferred style, just let the tools handle it instead.

Black is becoming more and more popular across Python projects and has recently became a project under the Python Software Foundation.

This is especially important with the adoption of mypy as type annotations create another obstacle or consideration in code formatting, but Black handles this just fine.

Many projects use both Black and isort. Using Black internally also helps ensure continued compatibility between the two popular projects.

For more details on Black, see:

https://github.com/psf/black

> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.


timothycrosley:
This proposal will 100% be accepted! Black is something I've started to incorporate in my own projects and at work. Compatibility with it will increasingly be a must.

In general I would like for isort to follow the same standard as [hug/HOPE-8](https://github.com/hugapi/HOPE/blob/master/all/HOPE-8--Style-Guide-for-Hug-Code.md).

Being consistent with that the only Black customization would be 100 line length.

Thanks for getting the ball rolling on this!

~Timothy

jdufresne:
That's great! I've rebased this PR to resolve conflicts.

PR #986, opened 2019-08-14
Use ==/!= to compare strings instead of is/is not

jdufresne:
The is/is not operators are not guaranteed to evaluate as true. They
only happen to due to CPython implementation details that exist as an
optimization.


jdufresne:
This is now included in #978.

PR #989, opened 2019-08-15
Simplify yield expressions using 'yield from' syntax

jdufresne:
https://www.python.org/dev/peps/pep-0380/


jdufresne:
This is now included as part of #991.

PR #999, opened 2019-08-17
Combine adjacent strings into one string

jdufresne:
Black has a bug such that strings that were previously multiple lines
sometimes end up as adjacent strings on one line. See:

https://github.com/psf/black/issues/26


jdufresne:
Any idea why uploading to codecov is failing on Travis? The error is:

```
HTTP 400

Please upload with the Codecov repository upload token to resolve issue.
```

jdufresne:
Looks like the error was only temporary

timothycrosley:
Not sure what caused it, but glad it's working now!

Thanks for all this great work!

PR #1016, opened 2019-09-23
fix invalid `*requirements*/*.{txt,in}` discovery

orsinium:
Now it adds files from the root when tries to discover subfolders with `requirements` in name


orsinium:
Codecov isn't ok with zeros. I guess, because of floating-point comparation :upside_down_face:

timothycrosley:
Looks great, thank you!

PR #1026, opened 2019-10-07
Fix bad sort with force_single_line and sort_within_sections

stpierre:
With both force_single_line and sort_within_sections enabled, modules
are not reliably sorted alphabetically; instead, all imports that
would be combined if force_single_line were false are instead grouped
together, even if another import should (lexicographically) be
somewhere in their midst. The new test case provides a concrete
example.

Also makes a small regex performance enhancement by precompiling
regexes used when force_single_line=True

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -66,6 +66,9 @@ if TYPE_CHECKING:
 
 
 class _SortImports:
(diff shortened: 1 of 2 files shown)


timothycrosley:
Looks great! Thank you!

PR #1028, opened 2019-10-09
Group a single uppercase character import with all uppercase imports

Siecje:
Instead of imports starting with an uppercase character.

### DIFF
--- a/isort/isort.py
+++ b/isort/isort.py
@@ -217,7 +217,7 @@ class _SortImports:
             module_name = str(module_name)
 
         if sub_imports and config["order_by_type"]:
-            if module_name.isupper() and len(module_name) > 1:
+            if module_name.isupper():
                 prefix = "A"
             elif module_name[0:1].isupper():
                 prefix = "B"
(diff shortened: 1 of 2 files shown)


timothycrosley:
Hi @Siecje!

Thanks for resurfacing this issue. I, sadly, don't think I can merge it as is, as this will break a lot of existing sorted code and there really isn't a a good way to determine the intended type of the variable (we try to group by type CONSTANT, Class, or variable - only based on casing - which will inevitably lead to some inconsistency). The current behaviour is actually semi-intentional: [https://github.com/timothycrosley/isort/issues/376](https://github.com/timothycrosley/isort/issues/376). Like the conversation in that ticket mentions, the best fix we can probably do is allow users to specify the intended type section for particular imports.

Thanks,

~Timothy

Siecje:
I'm going to keep `order_by_type=False`, so it doesn't matter to me, but you might want to add a test that expects this behaviour.

timothycrosley:
These are great suggestions! I've added a test and comment at the line in question: https://github.com/timothycrosley/isort/pull/1032/files

PR #1050, opened 2019-10-31
Issue #1051: find extension suffixes from importlib.machinery

rhwlo:
Thanks for considering my pull request!

This addresses https://github.com/timothycrosley/isort/issues/1051.

In PR https://github.com/timothycrosley/isort/pull/813 earlier this year, I tried to solve a problem with loading compiled extensions by populating an `ext_suffix` instance variable. While my solution worked for OS X and Linux, it unfortunately didn’t work for Windows, where compiled libraries will end with `.pyd` or with an architecture string and `.pyd`.

This PR changes the behavior to use the extension suffixes from Python’s `importlib.machinery` directly.

### DIFF
--- a/isort/finders.py
+++ b/isort/finders.py
@@ -1,4 +1,5 @@
 """Finders try to find right section for passed module name"""
+import importlib.machinery
 import inspect
 import os
 import os.path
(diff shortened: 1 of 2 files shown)


timothycrosley:
This looks great! Thank you!

PR #1071, opened 2019-12-19
explicitly check for first-party

honnix:
modules under pwd could be identified as first-party instead of
default configuration

@timothycrosley wdyt of this?


honnix:
```[RuntimeError]
The Poetry configuration is invalid:
  - Additional properties are not allowed ('website' was unexpected)
```

Incompatible poetry?

honnix:
This is a breaking change though because it would sort/separate imports differently than today.

To address https://github.com/timothycrosley/isort/issues/704 without installing all third party dependencies into virtual, one could configure default to be `THIRDPARTY`.

timothycrosley:
`modules under pwd could be identified as first-party instead of
default configuration`

I think this (pwd or specified src directory) is the correct approach and would like to get this in for the 5.0.0 release, thanks for the pull-request!

honnix:
Great, thanks! Should I try again to get CI fixed? That feels better to be a separated PR though.

honnix:
@timothycrosley ping

AdmiralNemo:
I came across this PR while researching how _isort_ determines whether a package is "first party" vs "third party."

Personally, I don't think this is a good way to determine the difference.  I have a lot of per-project virtual environments, which I keep inside the project directory in a `.venv` subdirectory.  This change would end up marking _everything_ as first party for these projects, since the virtual environment's "site-packages" directory path contains the current working directory path.

If the "first party" path could be made configurable (i.e. only things under `src/`), that would be fine, but please don't assume that everything in `${PWD}` is first party.

honnix:
I think this happens after checking for 3rd party.

honnix:
https://github.com/honnix/isort/blob/e3e39aaa25ab6c75be406d4fd80b43b4a5ddca80/isort/finders.py#L187

AdmiralNemo:
Ah, interesting.  There must be something else going on that I do not understand, because the only way I have ever managed to get _isort_ to classify anything as "third-party" is to put `default_section=THIRDPARTY`, and then explicitly list my own packages in `known_first_party`.  I was afraid that this change would prevent that, but it seems like there is some other problem I need to address. 🤷‍♂️

timothycrosley:
https://github.com/timothycrosley/isort/issues/1147 specifies the desired behaviour change along these lines

PR #1076, opened 2020-01-02
Make it smoother to use with `pre-commit`

glasserc:
- Add all optional requirements to the `.pre-commit-hooks.yaml` file so that users automatically get pipfile, pyproject, and requirements support.

- Add documentation of the `.pre-commit-hooks.yaml` to the README.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,3 +4,4 @@
     require_serial: true
     language: python
     types: [python]
+    additional_dependencies: [pipreqs, tomlkit, requirementslib, toml, pip-api]
--- a/README.md
+++ b/README.md
@@ -626,7 +626,19 @@ project.
 Git hook
 --------
 
-isort provides a hook function that can be integrated into your Git
+isort provides [pre-commit](https://pre-commit.com/) hook definition
+so you easily set up a hook to ensure that your imports remain
+sorted. To use it, add something like this to your
+`.pre-commit-config.yaml`.
+
+```yaml
+- repo: https://github.com/timothycrosley/isort
+  rev: 4.3.21-2
+  hooks:
+    - id: isort
+```
+
+isort also ships with a hook function that can be integrated into your Git
 pre-commit script to check Python code before committing.
 
 To cause the commit to fail if there are isort errors (strict mode),


.pre-commit-hooks.yaml
```diff
@@ -4,3 +4,4 @@
     require_serial: true
     language: python
     types: [python]
+    additional_dependencies: [pipreqs, tomlkit, requirementslib, toml, pip-api]
```
asottile: (pre-commit maintainer here) -- I would *not* suggest adding these to `additional_dependencies` by default -- this will slow down and bloat the installation when the consumer is not using these features -- the installation here should ~ideally be identical to just installing `isort`

plus there's potential for this to drift from the setuptools extras

README.md
```diff
@@ -626,7 +626,19 @@ project.
 Git hook
 --------
 
-isort provides a hook function that can be integrated into your Git
+isort provides [pre-commit](https://pre-commit.com/) hook definition
+so you easily set up a hook to ensure that your imports remain
+sorted. To use it, add something like this to your
+`.pre-commit-config.yaml`.
+
+```
```
asottile:
```suggestion
```yaml
```

README.md
```diff
@@ -626,7 +626,19 @@ project.
...
+- repo: https://github.com/timothycrosley/isort
+  rev: 4.3.21-2
+  hooks:
+    - id: isort
+```
```
asottile: you'll probably also want to mention `seed-isort-config` as per [this](https://github.com/pre-commit/mirrors-isort#classification-of-imports)

glasserc:
There's a question in my mind as to what qualifies as "bloat". It seems to me that installing all the optional features of a project is no more bloated than installing an auxiliary tool which only exists because the optional features aren't being used.

Just for fun, I benchmarked `isort`+`seed-isort-config` against `isort[pipfile,requirements,pyproject]` on a [reasonably-sized project](https://github.com/Kinto/kinto). The `seed-isort-config` combination was faster by ~10-15% (~2s vs ~2.3s).

I agree that repeating the requirements here isn't ideal. Maybe we should make these optional features mandatory instead?

timothycrosley:
@asottile is there a way for precommit users to install the optional dependencies only if they need them? If so, the solution I believe would be documentation. If not, then maybe we need multiple pre-commit versions, or I think the correct solution is to include all the dependencies.

asottile:
yep! that's the point of `additional_dependencies` for example, one could include the current configuration by doing the following:

```yaml
repos:
-   repo: https://github.com/timothycrosley/isort
    rev: ...
    hooks:
    -   id: isort
```

and if they need (for instance) the toml support, they would do this instead:

```yaml
repos:
-   repo: https://github.com/timothycrosley/isort
    rev: ...
    hooks:
    -   id: isort
        additional_dependencies: [toml]
```

so yeah, I agree -- this should be a documentation update and not an update to the provided yaml itself

timothycrosley:
@glasserc I really appreciate the thought you put into this, and I think this will lead to a better user experience. I'm closing this pull-request in favor of a documentation based solution: https://github.com/timothycrosley/isort/issues/1150. Which I think your existing work could be modified to resolve if you feel motivated to take it on.

@asottile Thanks for your help and feedback on this as well!

~Timothy

PR #1000, opened 2019-08-18
Add configuration to ensure a newline before each comment

FuegoFro:
This adds a `ensure_newline_before_comments` configuration that matches the format that `black` imposes (for better or worse) and allows `isort` to be used with `black` without any thrashing/conflicts with regards to blank lines before comments.

Should help address https://github.com/psf/black/issues/251

PR #1077, opened 2020-01-03
Add ensure_newline_before_comments to black profile

StephenBrown2:
Since #1000 was merged to address black compatibility, best include it in the profile.


timothycrosley:
This is great, thank you!

PR #1101, opened 2020-01-15
Add --ensure-newline-before-comments argument

StephenBrown2:
Also adds config option from #1000 to the argument list.

Feel free to disregard, if the args list is getting too out of hand.


timothycrosley:
This is great! Happy to have the args a little longer to keep things in sync.

Thanks!

~Timothy

PR #1110, opened 2020-01-22
Fixes #1015

### LINKED ISSUE #1015: isort line-wraps later when multiple imports, making lines too long
Should not `isort` always wrap after `import` when that is required to keep lines within the line length? When there are more than one import and Hanging Indent mode is used, that does not seem to be the case.

Or, in other words, should not `isort` always wrap after `import` in the example below?

```$ isort -w 25 -m 2 tmp.py```

Contents of `tmp.py`:
```
from package1 import first_package, \
    second_package
from package2 import \
    first_package

#                       |
#               25 characters
```

tegkhanna:
The first import was directly appended to the statement earlier(probably to prevent extra code for the first coma(',')). Checked the line length on the first import and added a new line if it goes beyond.

### DIFF
--- a/isort/wrap_modes.py
+++ b/isort/wrap_modes.py
@@ -110,8 +110,21 @@ def vertical(**interface):
 def hanging_indent(**interface):
     if not interface["imports"]:
         return ""
-
-    interface["statement"] += interface["imports"].pop(0)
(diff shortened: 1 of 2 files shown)


timothycrosley:
This is awesome! Thank you!

PR #1120, opened 2020-02-06
Remove Python 3.5 from trove classifiers

Siecje:


timothycrosley:
This is great, thank you!

Issue #1121, opened 2020-02-09
Switch to github actions


timothycrosley:
Currently isort uses .travis to run tests cases, however these tests can be very slow to run, as travis imposes limits on how many run in parallel. This ticket represents the work to move the tests to run on github actions to move past the travis limitation.

hugovk:
Please see PR https://github.com/timothycrosley/isort/pull/1122 for a start on this.

timothycrosley:
This is now merged in, and active for all testing across OSes. Thanks!

PR #1122, opened 2020-02-10
Lint and test on GitHub Actions

hugovk:
For #1121.

* Add a workflow for linting
  * Only runs on `ubuntu-latest` with Python 3.8
* Add a workflow for `poetry run pytest`
  * Runs on `ubuntu-18.04`, `ubuntu-16.04`, `macos-latest`
  * Maybe a single `ubuntu-latest` would be enough?
  * With Python 3.6-3.8

Examples:

* Lint https://github.com/hugovk/isort/runs/436539359?check_suite_focus=true
* Test https://github.com/hugovk/isort/runs/436539366?check_suite_focus=true

Not included:

* Deploy. This could be done with https://github.com/pypa/gh-action-pypi-publish
* Codecov. Tokenless upload is not yet supported, like with Travis. This either needs the token putting as plaintext in the repo (like https://github.com/pytest-dev/pytest/pull/6441), or waiting a bit longer, they're working on it now (https://github.com/codecov/codecov-action/issues/29#issuecomment-581540357)
* Windows. Omitted for now, the poetry install was failing. As this issue is about replacing Travis and not AppVeyor, that can wait.

For some reason, 3.8 on macOS is failing:

```
>           assert str(tmpdir.join("file1.py")) in out
E           AssertionError: assert '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0/file1.py' in ''
E            +  where '/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0/file1.py' = str(local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0/file1.py'))
E            +    where local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0/file1.py') = <bound method LocalPath.join of local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0')>('file1.py')
E            +      where <bound method LocalPath.join of local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0')> = local('/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pytest-of-runner/pytest-0/test_command_line_True_0').join
```

https://github.com/hugovk/isort/runs/436539509?check_suite_focus=true#step:7:218

Any ideas?

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,7 +10,7 @@ jobs:
         python-version: [3.8]
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
 
       - name: pip cache
         uses: actions/cache@v1
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,35 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: pip cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: lint-pip-${{ hashFiles('**/pyproject.toml') }}
+          restore-keys: |
+            lint-pip-
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade poetry
+          poetry install
+
+      - name: Lint
+        run: ./scripts/lint.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,51 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+        os: [ubuntu-18.04, ubuntu-16.04, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Ubuntu cache
+        uses: actions/cache@v1
+        if: startsWith(matrix.os, 'ubuntu')
+        with:
+          path: ~/.cache/pip
+          key:
+            ${{ matrix.os }}-${{ matrix.python-version }}-${{ hashFiles('**/pyproject.toml') }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.python-version }}-
+
+      - name: macOS cache
+        uses: actions/cache@v1
+        if: startsWith(matrix.os, 'macOS')
+        with:
+          path: ~/Library/Caches/pip
+          key:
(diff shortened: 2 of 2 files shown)


timothycrosley:
This all looks fantastic! Thank you! Similar to PyLint, I'd feel okay leaking the coverage key if needed for now. I'm not sure the cause of the mac os failure, but I'm going to work on securing a permanent mac testing machine to help diagnose.

Thanks!

~Timothy

hugovk:
You'll need to fetch the coverage token from https://codecov.io/gh/timothycrosley/isort/settings, but I can help set it up, if you like.

Would you like it included in this PR or a follow up? It might be better in a follow up, because then GHA will have been enabled in this repo -- it needs the config files to be merged before GHA can be used in PRs.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+
+    steps:
+      - uses: actions/checkout@v1
```
cclauss: v1 --> v2?  https://github.com/actions/checkout/releases

.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
...
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
+        os: [ubuntu-18.04, ubuntu-16.04, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v1
```
cclauss: v1 --> v2?  https://github.com/actions/checkout/releases

hugovk: Thanks, updated both!

.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --upgrade poetry
+          poetry install
+
+      - name: Test
+        shell: bash
```
MCOfficer: why specify bash here?

hugovk: There are different defaults for different platforms:

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

MCOfficer: True, but this runs on macOS and ubuntu, which both default to `bash`.

cclauss: Current releases of macOS default to `zsh`, not `bash` so I approve of the _explicit is better than implicit_ approach used here.

MCOfficer: it's a nitpick, so i feel fairly indifferent about it.

hugovk: Yeah, Windows would be a great addition for the future as AppVeyor is really slow with a single concurrent job:

> Not included:
> ...
> * Windows. Omitted for now, the poetry install was failing. As this issue is about replacing Travis and not AppVeyor, that can wait.

PR #1127, opened 2020-02-10
issue-1125 remove documentation examples of two character flags

ucodery:


timothycrosley:
This look great thank you!

PR #1138, opened 2020-02-12
Fix Config instance copy mechanism

bryant1410:


bryant1410:
For context, the problem was that the original `config` got modified as well when copying. So then the line length goes smaller with the larger line numbers in a file.

I saw this was the only usage of copying I could find in the codebase.

timothycrosley:
Thanks for catching this, adding a test case, and fixing! The whole reason to make a copy was to avoid mutating! So this is a very good catch :)! Thanks!

PR #1172, opened 2020-04-05
Fix: grid mode wrongly fall back to hanging grid

innerlee:
When the line ending is `\r\n` and the line is exact full, this grid mode will be wrongly fall back to hanging grid because \r counts for 1 character.

This conflicts with yapf and will be annoying when hit, because pre-commit will prevent you from submitting commits.


timothycrosley:
No more work is being done against the master branch in preparation for the 5.0.0 release.

Issue #1192, opened 2020-05-07
--keep-direct-and-as options should be removed


ucodery:
This setting now defaults to True, so there is no point in setting it in the shell. It should be marked deprecated and then removed after a few minor releases.

Additionally, there is no mirror to this option (--discard-direct-with-as if you will), and it is not listed as a configurable setting in the wiki (although the tests suggest that it would work if added to a config). The only sure way to get the old behavior is to use the api directly. For these reasons I would suggest also removing keep_direct_and_as_imports from isort code as a whole. IMHO this setting just has the ability to introduce bugs. The following simple script would break if keep_direct_and_as_imports was set to False, for instance in an upstream commit hook.
```python
from os import path
import os.path as pth

def foo(bar):
    return path.pathsep.join(bar)
```

timothycrosley:
Thanks for reporting! This is fixed in develop and slated to be released in the 5.1.0 minor release by the end of the month

timothycrosley:
A fix for this has been deployed to PyPI in the [5.1.0 release](https://timothycrosley.github.io/isort/CHANGELOG/#510-july-14-2020), 

Thanks for reporting!

~Timothy

PR #1193, opened 2020-05-07
Issue 1160: remove redundant as imports

ucodery:


ucodery:
I still have an open question as to how this logic would be expected to interact with "keep_direct_and_as_imports", specifically when it is False.
This behavior is changed, but to me the result is pretty obvious
```python
import isort
import isort as isort
# now produces -->
import isort
```
This behavior is also changed, and IMO makes for cleaner code, but would probably surprise me if I had used the keep_direct_and_as_imports=False option.
```python
import isort as isort
# now produces -->
import isort
```

Given that the above, simpler code, was achieved without additional special cases in output.py, and that I opened #1192, I have submitted with this new behavior.

graingert:
@ucodery this `import x as x` syntax is used in mypy with https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport

ucodery:
@graingert interesting, it is good to see some documentation why this shadowing is used in some places. Given the MyPy convention, would you still agree that the first example is correct? I still see two import lines importing the same thing, and not changing the namespace, as a lint error, and I still see defaulting to the shorter as more straightforward.

timothycrosley:
Thanks for all the input and work here! Took the ideas, and additional feedback, and this is now incorporated under this flag: https://timothycrosley.github.io/isort/docs/configuration/options/#remove-redundant-aliases

PR #1212, opened 2020-05-22
sort_file: Preserve permission bits

### LINKED ISSUE #1194: Removes executable bit from executable scripts
The isort `develop` branch (commit cf93f354610e26c47124e00ac1c4fd5b4e376ad2) fails to preserve the executable bit on scripts. This regressed somewhere between 490e135840908efd06399430c053643ac074df91 and a6ada7f39449ddac91fa94bb7e56d326c4734aac (none of the commits between those actually run).

```console
$ cat > test.py <<EOF
#!/usr/bin/env python3
import sys
import platform
sys.stdout.write(platform.python_version())
EOF
$ chmod +x test.py
$ ls -l test.py
-rwxr-xr-x 1 anders anders 95 May  7 18:24 test.py
$ ./test.py
3.8.2
$ isort test.py
Fixing /tmp/test.py
$ ./test.py
bash: ./test.py: Permission denied
$ ls -l test.py
-rw-r--r-- 1 anders anders 95 May  7 18:24 test.py
```

andersk:
isort 4 would overwrite the old file in place, so its permission bits would not be modified.  But as of commit 1920a6c47796bec4ddc2141dbaece6b44a93834e or so, isort instead writes a new file with default permission bits and renames it into place.  So we now need to explicitly copy the permission bits.

Fixes #1194.

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -1,3 +1,4 @@
+import shutil
 import sys
 import textwrap
 from io import StringIO
@@ -479,6 +480,7 @@ def sort_file(
                     with tmp_file.open(
                         "w", encoding=source_file.encoding, newline=""
                     ) as output_stream:
+                        shutil.copymode(filename, tmp_file)
                         changed = sorted_imports(
                             input_stream=source_file.stream,
                             output_stream=output_stream,


timothycrosley:
This looks great! Thanks for catching and fixing this!

PR #1227, opened 2020-06-11
Run pre-commit hook only on python files

feluelle:


feluelle:
Oh found there is already https://github.com/timothycrosley/isort/issues/1154 for that and it was fixed in the `develop` but not yet released?

PR #1266, opened 2020-07-05
Add Changelog, docs and tests to source tarball

sebix:
Same as timothycrosley/isort#531 which does no longer work, as now
poetry seems to be used for building the packages

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,14 @@ classifiers = [
     "Topic :: Utilities",
 ]
 urls = { Changelog = "https://github.com/timothycrosley/isort/blob/master/CHANGELOG.md" }
+include = [
+    "CHANGELOG.md",
+    "docs/configuration/*.md",
+    "docs/contributing/*.md",
+    "docs/major_releases/*.md",
+    "docs/quick_start/*.md",
+    "tests/*.py"
+]
 
 [tool.poetry.dependencies]
 python = "^3.6"


timothycrosley:
Thank you!

timothycrosley:
@sebix unfourtantely, I had to revert this. It looks like adding extra files is not quite as straight forward with poetry as I would have assumed. See: https://github.com/timothycrosley/isort/issues/1285. Which I confirmed locally, all the included files ended up just littering the root of my sitepackages. I couldn't easily figure out a good way around it, and the state of the issues on Poetry's issue tracker around it didn't give me much hope

sebix:
> @sebix unfourtantely, I had to revert this. It looks like adding extra files is not quite as straight forward with poetry as I would have assumed. See: #1285. Which I confirmed locally, all the included files ended up just littering the root of my sitepackages. I couldn't easily figure out a good way around it, and the state of the issues on Poetry's issue tracker around it didn't give me much hope

Ok, too bad that this is not - easily - possible with poetry. The files should have just ended up in the sdist tarball. Sorry for the inconvenience.

timothycrosley:
No problem at all! I appreciate you drawing my attention to the issue! I've reopened https://github.com/timothycrosley/isort/issues/463

sebix:
Adding python files (tests) in sdist only is possible -> https://python-poetry.org/docs/pyproject/#packages and #1313 
Adding static files (docs) in sdist is currently not possible -> https://github.com/python-poetry/poetry/issues/1338

PR #1282, opened 2020-07-06
Fix typo in profile doc description build

scop:
None of the profiles actually contain a description so this is a no-op, but happened to notice.


timothycrosley:
Thanks!

Issue #1276, opened 2020-07-06
lines after imports bug with triple quotes


hyeonjames:
Hello!

I found that isort creates only one line when it starts with decorator and triple quotes for top level function or class definition. 

( I am not sure if it actually follows PEP8 but my PyCharm and flake8 see it as an error )

I am using isort v5.0.3

```python
from module import decorator


@decorator("""
hello
""")
def a():
    pass
```

I think this code is supposed to have no problem with isort, but when I used isort, it causes the error below:
```bash
(base) james@MacBook-Pro-9 playground % isort isort_bug.py --diff --check
ERROR: /Users/james/playground/isort_bug.py Imports are incorrectly sorted and/or formatted.
--- /Users/james/playground/isort_bug.py:before 2020-07-06 18:22:34.970555
+++ /Users/james/playground/isort_bug.py:after  2020-07-06 18:22:44.025902
@@ -1,5 +1,4 @@
 from module import decorator
-
 
 @decorator
```

Besides, I think it is caused only when it starts with triple quotes

Below codes have no problem
```python
from module import decorator


@decorator("hello")
def a():
    pass
```

hyeonjames:
I think I can fix it, I am working on it.

timothycrosley:
Thanks for reporting and fixing! Your changes have been deployed with the 5.0.6 release of isort :rocket: 

Thanks!

~Timothy

PR #1299, opened 2020-07-08
Issue #1276: Fix a bug that creates only one line after triple quotes

hyeonjames:
I think these sections are to find the first line after the tail of imports and I think it does not have to parse anymore when the line is in quotes

https://github.com/timothycrosley/isort/blob/develop/isort/output.py#L180-L206


timothycrosley:
This looks great! Thank you!

PR #1312, opened 2020-07-11
Fix #1275 in the case where the --add-imports option is given

### LINKED ISSUE #1275: Regression: Blank line after docstring removed.
Thank you for the software and congratulations on a major release.

--

I have a file like the following:

```python
"""
My docstring
"""

from b import thing
from a import other_thing
```

If I run `isort example.py` I expect this file to change to:

```python
"""
My docstring
"""

from a import other_thing
from b import thing
```

However, it changes to:

```python
"""
My docstring
"""
from a import other_thing
from b import thing
```

with `isort` 5.0.3.

This works as expected with `isort` 4.3.21.

--

Also, the newline is not removed when `isort` does not make sorting changes, so this leaves my codebase inconsistent.

bbugyi200:
Fixes #1275.

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -456,6 +456,11 @@ def _sort_imports(
                         import_section += line_separator.join(add_imports) + line_separator
                         contains_imports = True
                         add_imports = []
+
+                        # HACK: This will occur on the blank line after a
+                        # module docstring. We want to preserve this blank
+                        # line.
+                        output_stream.write(line_separator)
                     else:
                         not_imports = True
                 elif (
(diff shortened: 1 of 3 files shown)


timothycrosley:
Thanks for this! I've manually merged this in develop, (just to try to find a slightly less hacky way to do it: https://github.com/timothycrosley/isort/commit/fd87001295467c87612a8cdbcaa4c3239f4989b6)

Thanks!
~Timothy

bbugyi200:
@timothycrosley No problem. I know this solution is crap. The test case is good though.

I didn't feel like really diving into that `_sort_imports()` function to figure out wth is going on. That function needs to be refactored IMO.

PR #1319, opened 2020-07-13
Suppress setuptools' warning

NeilGirdhar:
Setuptools expects distutils to be imported after it.


timothycrosley:
Thanks for raising this! I've fixed this in develop (couldn't use this exact solution due to linting failing)

NeilGirdhar:
@timothycrosley Thanks for the quick fix!

PR #1323, opened 2020-07-15
Use correct spelling for use_parentheses

dylrich:
The spelling of this configuration option was updated, but the documentation still uses the old spelling.


timothycrosley:
Thank you!

Issue #1328, opened 2020-07-19
force_sort_within_sections with multi-line file docstring causes comments to be duplicated?


EricCousineau-TRI:
In https://github.com/RobotLocomotion/drake/pull/13709#pullrequestreview-451092149, I noticed that a comment was getting duplicated.

Managed to get a min repro here, using `isort==5.1.2`:
https://github.com/EricCousineau-TRI/repro/blob/e73591778217dff2937db6eae5155e5574d184c0/bug/isort_bug/code.py

Basically, the following source code:
```py
"""
Multi-line docstring
"""
# Comment for A.
import a
# Comment for B - not A!
import b
```
gets transformed into this:
```py
"""
Multi-line docstring
"""
# Comment for B - not A!
import a
# Comment for B - not A!
import b
```

(see the linked `code.py` for comparison)

timothycrosley:
Thanks for reporting! This is fixed in the 5.1.3 patch release

EricCousineau-TRI:
Sweet! Thanks for the speedy response!

(Just an xref: bf47b311f854d25a9161e9ae6cda709002bf3f77)

EricCousineau-TRI:
BTW I'm guessin' ya may already know, but here're GitHub's "magic words" for auto-closing issues via PRs or commits:
https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

I think this would've been auto-closed if you removed the intermediate "issue", e.g. "Fixed #1328: ..."

PR #1329, opened 2020-07-19
test_isort: Reproduce comment-duplication bug

EricCousineau-TRI:
This does not fix anything, it just reproduces the bug in #1328


EricCousineau-TRI:
Er, meant to make this a draft PR - sorry!

timothycrosley:
Thanks for giving a repro test with the issue! I've added this test into isorts regression test suite:

https://github.com/timothycrosley/isort/blob/develop/tests/test_regressions.py#L432

Thanks!

~Timothy

PR #1342, opened 2020-07-23
Make sure that warnings are formatting properly

Gricha:
Hey! If there is a syntax error right now, `isort` will print something like this:
```
/Users/greg.pstrucha/.virtualenvs/bonfire/lib/python3.8/site-packages/isort/api.py:345: UserWarning: {file_path} unable to sort due to existing syntax errors
  warn("{file_path} unable to sort due to existing syntax errors")
```

I've noticed we are not formatting strings properly there, let me know if that works!

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -356,9 +356,9 @@ def sort_file(
                     except FileNotFoundError:
                         pass
         except ExistingSyntaxErrors:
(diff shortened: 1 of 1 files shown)


timothycrosley:
Thank you for finding and fixing this!

PR #1345, opened 2020-07-25
Adds lazy option to git hook.

### LINKED ISSUE #1214: git_hook() lazy option
Can we get a lazy option added, that matches flake8? i.e. It should include unstaged files if set to True.

https://gitlab.com/pycqa/flake8/-/blob/master/src/flake8/main/git.py#L22

sztamas:
Fixes #1214


timothycrosley:
:tada: This is awesome! Thank you!

PR #1349, opened 2020-07-26
Adds color to ERROR and SUCCESS in check mode.

### LINKED ISSUE #1177: Verbose -- colored result
When debugging isort issues, it would be useful to print result as ERROR, SUCCESS etc. with color.
* ERROR = red
* SUCCESS = green
etc.
It will be more convenient to find the message by color

sztamas:
Fixes #1177.

Did this with `colorama` as I think it doesn't warrant introducing something like `termcolor` or `blessings`. `colorama` is needed anyways with both of those to support Windows.

Some further considerations:

- we might want a `--no-color` option to turned the colors explicitly off. Colorama turns them off by default if you're piping the results into another command
- the diffs aren't colored when using `--diff`

_Sidenote for @timothycrosley : we're running out of issues marked with `help_wanted`, so you might want to add a few more if you have the time :smile:_


timothycrosley:
This looks great! Unfortunately, this is complicated a bit by isorts policy to never include dependencies. isort is depended on by some large projects such as pylint, so in the past when it did include or rely on dependencies the odds of conflicts were very high. See: https://timothycrosley.github.io/isort/docs/major_releases/release_policy/#packaging-guarantees. I do think that the policy could be clarified to say that will change isort file output. However, it would still be necessary I believe to make this the non default, and require a flag such as `--color` for those that want it.

sztamas:
OK, thanks for sharing the background info!

How about adding `colorama` to extras (https://python-poetry.org/docs/pyproject/#extras) then?

We can keep the existing functionality as is, add a new `--color` option which errors out with a descriptive message to install the `colors` extra if `colorama` is missing.

What do you think?

timothycrosley:
@sztamas Exactly! That sounds like a perfect solution

timothycrosley:
This looks awesome! Thank you!

PR #1351, opened 2020-07-27
Flake8 config cleanup.

### LINKED ISSUE #1311: No changelog posted for 5.0.9
I'm not sure if the website just takes longer to update, or if it was accidentally missing but 5.0.9 (latest version) is not mentioned in the docs:

https://timothycrosley.github.io/isort/CHANGELOG/

sztamas:
Fixes #1311 

- specify flake8 config only in `setup.cfg` - removed passing in the same options in the `lint.sh` scripts as well
- remove `F401` and `F403` from the global ignore list
- added per-file `F401` ignores for `__init__.py` files and per-line `F401` ignores for all other cases
- removed unused import masked by global `F401` ignores
- changed DeepSource ignore comments to `# noqa: F401`, because those work for both flake8 and DeepSource
- added links to black documentation on why we choose to ignore `W503` and `E203` - the only globally ignored rules left

### DIFF
--- a/isort/__init__.py
+++ b/isort/__init__.py
@@ -1,5 +1,5 @@
 """Defines the public isort interface"""
-from . import settings  # noqa: F401
+from . import settings
 from ._version import __version__
 from .api import check_code_string as check_code
 from .api import check_file, check_stream, place_module, place_module_with_reason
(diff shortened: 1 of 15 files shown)


timothycrosley:
This is an awesome set of improvements, thank you!

PR #1360, opened 2020-07-31
Fix max_line_length parsing issue

JiriSko:
### Bug
Invalid value in `.editorconfig` file causes crash of isort.

**Example of invalid `.editorconfig`:**
```
[*]
max_line_length = null
```

### Current behaviour:
isort crashed with error:
```
File "/var/task/python/lib/python3.8/site-packages/isort/settings.py", line 229, in _update_with_config_file
computed_settings['line_length'] = float('inf') if max_line_length == 'off' else int(max_line_length)
ValueError: invalid literal for int() with base 10: 'null'
```

### Expected behaviour:
No crash.

**Simple fix:** just check if the `max_line_length` value is correct.


timothycrosley:
Thank you for reporting and resolving this issue!

PR #1370, opened 2020-08-05
Added simple quiet config check to check_stream

quoimec:
Relates to [issue 1369](https://github.com/timothycrosley/isort/issues/1369) - simple check added to not print out to console when the quiet flag has been set.


timothycrosley:
The PR itself looks great! Closing this only because of the projects conservative approach to changing CLI options, captured on the related issue

PR #1374, opened 2020-08-11
Complete documentation for multi line modes.

sztamas:
I could swear I had these in the original PR, but I somehow managed to not commit them and couldn't find them locally either. :-)


timothycrosley:
I kept this in that back of my head, as something I needed to go back and add. Really appreciate you fixing it up and adding the missing documentation!

Issue #1321, opened 2020-07-15
--combine-as loses # noqa


scop:
```
$ isort --version | grep VERSION
                    VERSION 5.1.0
$ cat t2.py 
from foo import *  # noqa
from foo import bar as quux  # noqa
$ isort --combine-as t2.py 
Fixing /tmp/t2.py
$ cat t2.py 
from foo import *  # noqa
from foo import bar as quux
```

`# noqa` on the `from foo import bar as quux` line was lost.

timothycrosley:
This is fixed in 5.4.0

Thanks!

~Timothy

PR #1376, opened 2020-08-12
Bugfix #1321: --combine-as loses comment

sztamas:
@timothycrosley please feel free to adjust if needed, but at least it gives you a hint on what is wrong.

Actually it the `# noqa` comment isn't lost, but it is combined up (incorrectly) into the '*' import, but because both of the comments are the same (ie. `noqa`) it is shown only once.
If you changed the second comment to someone else it would should up on the star import, again incorrectly.

BTW, I find it strange that `--combine-star` works for as imports as well.
Is that by design?
Wouldn't that break code that relies on those aliases?

### DIFF
--- a/isort/output.py
+++ b/isort/output.py
@@ -301,6 +301,11 @@ def _with_from_imports(
                 new_section_output.extend(above_comments)
 
             if "*" in from_imports and config.combine_star:
+                if config.combine_as_imports:
+                    comments = list(comments or ())
+                    comments += parsed.categorized_comments["from"].pop(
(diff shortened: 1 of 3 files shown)


timothycrosley:
Thanks @sztamas!

>  I find it strange that `--combine-star` works for as imports as well.

This is not by design, but has been the behaviour since the flag was introduced. I guess I would call it more an oversight, or possibly implementation shortcut, than a deliberate design decision. 

I've added a bug for this issue here: https://github.com/timothycrosley/isort/issues/1380

Issue #1375, opened 2020-08-12
--dont-order-by-type broken


Zacharias030:
I just used `pip install isort` and now cannot use the `--dt` / `--dont-order-by-type` flag from the `--help` page.

```python
✗ isort myfile.py --dont-order-by-type
Traceback (most recent call last):
  File "/usr/local/bin/isort", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/isort/main.py", line 812, in main
    config = Config(**config_dict)
  File "/usr/local/lib/python3.7/site-packages/isort/settings.py", line 422, in __init__
    super().__init__(sources=tuple(sources), **combined_config)  # type: ignore
TypeError: __init__() got an unexpected keyword argument 'dont_order_by_type'
```

Zacharias030:
This is no problem in version 4.3.21!

timothycrosley:
This is fixed in 5.4.0

Thanks @Zacharias030  for reporting and @sztamas for fixing!

~Timothy

Zacharias030:
Thanks a lot! Great software & spectacular support!

PR #1377, opened 2020-08-12
Bugfix #1375: --dont-order-by-type broken

sztamas:


timothycrosley:
Thanks for fixing this!

PR #1386, opened 2020-08-14
Fix the first known party path expansion

kjagiello:
The first known party path expansion stopped working in 5.0.5. The test for this feature was incorrectly fixed in https://github.com/timothycrosley/isort/commit/1bef08e4a56b07301b5a88c4b79e677f3414173c#diff-3e6dbf50f49313d010ff846279b07462R920 This PR fixes that regression by relying on `Config.directory`.

Thanks @lundberg for the feature and helping out with this PR!

### DIFF
--- a/tests/test_isort.py
+++ b/tests/test_isort.py
@@ -1072,27 +1072,33 @@ def test_thirdy_party_overrides_standard_section() -> None:
     assert test_output == "import os\nimport sys\n\nimport profile.test\n"
 
 
-def test_known_pattern_path_expansion() -> None:
+def test_known_pattern_path_expansion(tmpdir) -> None:
     """Test to ensure patterns ending with path sep gets expanded
     and nested packages treated as known patterns.
     """
+    src_dir = tmpdir.mkdir("src")
+    src_dir.mkdir("foo")
+    src_dir.mkdir("bar")
     test_input = (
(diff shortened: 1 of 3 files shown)


lundberg:
Awesome @kjagiello ❤️ Thanks for raising and fixing this issue.

tests/test_isort.py
```diff
@@ -1072,27 +1072,33 @@ def test_thirdy_party_overrides_standard_section() -> None:
...
 
 
-def test_known_pattern_path_expansion() -> None:
+def test_known_pattern_path_expansion(tmpdir) -> None:
     """Test to ensure patterns ending with path sep gets expanded
     and nested packages treated as known patterns.
     """
+    src_dir = tmpdir.mkdir("src")
+    src_dir.mkdir("foo")
+    src_dir.mkdir("bar")
```
lundberg, line 1081: Great that the test case is more realistic now and mimics the use case of paths as known patterns 👍🏻

timothycrosley:
Fantastic work! Sorry the regression was introduced, and really appreciate the thoroughness in fixing it and making sure it wont reappear with the improved test case.

Thanks!

~Timothy

timothycrosley:
This has been pushed out in the 5.4.2 release of isort

PR #1390, opened 2020-08-18
Issue/1389/ensure new line before comments bug

### LINKED ISSUE #1389: ensure_newline_before_comments shouldn't rely on the default lines between sections config
The current implementation of `ensure_newline_before_comments` works only inside a section.

This bug goes undetected with the default configuration of `1` for `lines_between_sections`, but no empty line is inserted before a comment if `lines_between_sections` is changed to `0`.

sztamas:
Fixes #1389.

### DIFF
--- a/isort/output.py
+++ b/isort/output.py
@@ -138,12 +138,6 @@ def sorted_imports(
             for line in new_section_output:
                 comments = getattr(line, "comments", ())
                 if comments:
-                    if (
-                        config.ensure_newline_before_comments
-                        and section_output
-                        and section_output[-1]
-                    ):
-                        section_output.append("")
                     section_output.extend(comments)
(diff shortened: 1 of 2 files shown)


isort/output.py
```diff
@@ -172,6 +166,9 @@ def sorted_imports(
         else:
             pending_lines_before = pending_lines_before or not no_lines_before
 
+    if config.ensure_newline_before_comments:
```
sztamas, line 169: Wasn't sure whether we want to insert an empty line before on a comment above the first import as well.
It looks a bit strange to me to have it there if there are not other imports before so I opted for not having an empty line in this special case.

In case you want it just move these 2 lines below the stripping of empty lines that follows next. I tested that it works there as well.

PR #1406, opened 2020-08-25
Update timothycrosley references to pycqa

sztamas:
As a result of moving the repo to PyCQA some links were broken.
Some of them were properly redirected, so it is easier to just mechanically change all the references.

Left the following badge links in `README.md` still with old references as they need some setup in the services they are linking to:

- Maintainability badge - codeclimate
- Gitter badge
- DeepSource badge


timothycrosley:
Great work! Thanks for doing this!

PR #1408, opened 2020-08-25
Fix documentation for skip_file action comment

ndevenish:
The [documentation](https://pycqa.github.io/isort/docs/configuration/action_comments/) currently lists this as `skip-file`, whereas the default in settings.py has an underscore: https://github.com/PyCQA/isort/blob/cf44d6720f27973cdb54a48bbe8b91bae9055821/isort/settings.py#L31-L34

Am assuming people might be relying on the actual implementation, so changing the documentation is the most reasonable course (absent knowledge of intention)

This had me confused for a bit!

### DIFF
--- a/docs/configuration/action_comments.md
+++ b/docs/configuration/action_comments.md
@@ -3,7 +3,7 @@
 The most basic way to configure the flow of isort within a single file is action comments. These comments are picked up and interpreted by the isort parser during parsing.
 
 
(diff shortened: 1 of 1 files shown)


timothycrosley:
Sorry it led to confusion, and thanks for fixing it! This will help keep every future reader of the docs lest confused :rocket:

PR #1409, opened 2020-08-25
Adds colors to --diff output.

### LINKED ISSUE #1405: Add coloring to --diff output
Although I was very excited to try out the color output config option (now that pylint allows isort >= 5), it does not seem to do all that much at the moment. In particular, I have expected that diffs will be nicely colored just like they are with git diff, but that does not seem to be the case. Am I missing something or is this not implemented yet?

Thank you!

sztamas:
Works on the output of `unified_diff` and assumes that all lines starting with a single `+` or a single `-` were marked so by `unified_diff`.

I can't think of a use case when you would have lines starting with a `+` or `-` in the import section where isort operates, so I think it should be fine.

Resolves #1405

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -131,6 +131,7 @@ def sort_stream(
             file_output=_output_stream.read(),
             file_path=file_path,
             output=output_stream if show_diff is True else cast(TextIO, show_diff),
+            color_output=config.color_output,
         )
         return changed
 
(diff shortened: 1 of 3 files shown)


timothycrosley:
This is great! 

> I can't think of a use case when you would have lines starting with a + or - in the import section where isort operates, so I think it should be fine.

And, if I understand it correctly, the only consequence would be incorrect coloring, which while we certainly want to avoid wouldn't  be too major of an issue.

PR #1428, opened 2020-08-28
Bugfix: file_path in warnings in isort.api.sort_file

howeaj:
This pull request fixes the warnings reported by isort.api.sort_file, which currently do not report the same filepath that was actually used in the preceding code; incorrectly reporting "None" in some cases.


timothycrosley:
:tada: Thank you for finding and fixing this!

PR #1436, opened 2020-08-31
Get rid of warnings displayed when running pytest.

sztamas:
@timothycrosley I guess you get warnings as well when running pytest?

I don't like scrolling back through them to get to the diffs etc., so I used these filters locally for a while.

Maybe worth adding to the repo?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,6 +14,10 @@ strict_optional = False
 
 [tool:pytest]
 testpaths = tests
+filterwarnings =
+    ignore::DeprecationWarning:distlib
+    ignore::DeprecationWarning:requirementslib
+    ignore::hypothesis.errors.NonInteractiveExampleWarning:hypothesis
 
 [flake8]
 max-line-length = 100


jdufresne:
Would it be at all possible to resolve or fix these warnings rather than ignore them?

sztamas:
Unfortunately, no as they are from other packages, the only ones that could be fixed by @timothycrosley are from `hypothesis-auto`, as that is his repo, but I considered those as well warnings from "outside" packages.

timothycrosley:
These have been bugging me for a while, thanks for adding an ignore! For a permanent solution for the warnings I've spun of a separate ticket here to address @jdufresne's point. For hypothesis-auto thanks to @Zac-HD we should be able to migrate those tests to full hypothesis tests fairly easily using ghostwriter. For the others, they almost all come from dependencies that are only used for deprecated module placement, so we may be able to find a good strategy to remove the warnings from the main test suite

PR #1438, opened 2020-08-31
Replace vendored toml package with a dependency

jdufresne:
isort is normally installed in dev environments only, not production
environments. In these environments, minimizing dependencies is less
necessary.

Additionally, isort is often installed alongside other linting tools,
such as black, which already have toml as a dependency. Rather than
install the same package twice (once vendorized) allow sharing the
common installation. This reduces the overall install size of a
project's dev dependencies.


timothycrosley:
Unfortunately, I cannot accept this pull request. For better or worse I made a commitment at least for the duration of the 5.0.0 release to not include any required or even optional dependencies (with exceptions only being able to be made for cosmetic requirements that don't affect configuration loading or output and aren't utilized by default). See: https://pycqa.github.io/isort/docs/major_releases/release_policy/#packaging-guarantees. 

I do think it would be worth revisiting this policy when 6.0.0 is released likely a year from now. I will say, at least my knowledge of Python's packaging situation, makes non vendored dependencies tough while ensuring isort always plays nice with other formating and linting tools. As an example, if toml release a 0.11.0 that is faster but API incompatible, and black upgraded to it, then isort would break usage of the latest version of black until it also got upgraded. Disk space is so cheap, that I think an argument can be made that avoiding dependency issues > reduction in disk space. If there was a clean way to install multiple version of a library within one Python environment this would be a non-issue of course.

jdufresne:
No worries, thanks for the detailed response. I hope at some point in the future dependencies can be unbundled, but I don't consider it a major issue at the moment.

> isort will never have dependencies, optional, required, or otherwise.

Does that mean these optional dependencies should be removed or vendored? AFAICT, they violate the policy.

https://github.com/PyCQA/isort/blob/develop/pyproject.toml#L41-L44

> As an example, if toml release a 0.11.0 that is faster but API incompatible, and black upgraded to it, then isort would break usage of the latest version of black until it also got upgraded.

This is always the risk of using third party tools. Same as any project. We're at the mercy of the upstream project acting in good faith and provide reasonable releases with good numbering. Best we can do is pin a range of allowed versions.

On the other hand, if the vendored toml library happens to include a security vulnerability, bundling it puts isort users at risk. A keen sysadmin will regurarly upgrade third party packages, but once bundled by isort, that version is locked in and users must wait for a new isort release.

> Disk space is so cheap

Normally yes.

There are situations where that is not the case. Such as serverless computing (where I have personally seen dependencies add unnecessary overhead) and embedded systems. Luckily, isort is not normally used in these scenarios.

PR #1485, opened 2020-09-15
Exit non zero if all of passed path is broken

lntuition:
- [x] Exit non zero value if all isort attempt fails
- [x] Correct break of existing test suite
- [x] Add new test suite that test all attempt fails

### DIFF
--- a/tests/unit/test_isort.py
+++ b/tests/unit/test_isort.py
@@ -3243,13 +3243,14 @@ def test_safety_skips(tmpdir, enabled: bool) -> None:
     else:
         config = Config(skip=[], directory=str(tmpdir))
     skipped = []  # type: List[str]
+    broken = []  # type: List[str]
     codes = [str(tmpdir)]
-    main.iter_source_code(codes, config, skipped)
+    main.iter_source_code(codes, config, skipped, broken)
 
     # if enabled files within nested unsafe directories should be skipped
     file_names = {
         os.path.relpath(f, str(tmpdir))
-        for f in main.iter_source_code([str(tmpdir)], config, skipped)
+        for f in main.iter_source_code([str(tmpdir)], config, skipped, broken)
     }
     if enabled:
         assert file_names == {"victim.py"}
@@ -3266,7 +3267,9 @@ def test_safety_skips(tmpdir, enabled: bool) -> None:
     # directly pointing to files within unsafe directories shouldn't skip them either way
     file_names = {
         os.path.relpath(f, str(toxdir))
-        for f in main.iter_source_code([str(toxdir)], Config(directory=str(toxdir)), skipped)
+        for f in main.iter_source_code(
+            [str(toxdir)], Config(directory=str(toxdir)), skipped, broken
+        )
     }
     assert file_names == {"verysafe.py"}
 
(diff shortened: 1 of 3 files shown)


lntuition:
I know that we need to add a test case. but I don't know how to start that job.. 😢 I'd appreciate if you could show me the way !

anirudnits:
@lntuition the thing is your code actually works I believe but because in the test:
https://github.com/PyCQA/isort/blob/e5ab07ddcbe79d7d11fef32ee12938fc8bbca8d6/tests/unit/test_isort.py#L3717

no files are actually being scanned that's why this `subprocess.run` call:
https://github.com/PyCQA/isort/blob/e5ab07ddcbe79d7d11fef32ee12938fc8bbca8d6/tests/unit/test_isort.py#L3737

comes with non-zero exit code (that's with your changes that is) and then raises a `subprocess.CalledProcessError` error and that's why the test fails. For more info on `subprocess.run`: https://www.digitalocean.com/community/tutorials/how-to-use-subprocess-to-run-external-programs-in-python-3. I believe if you can catch this exception would solve the problem but I'm not sure because I haven't worked much on this.

timothycrosley:
The test that's failing makes me think a better strategy may be to check each file or folder that is passed in for validity (is a folder or a file that exists) I'm not sure that passing in a directory that doesn't happen to have sortable files in it yet should error out.

anirudnits:
@lntuition the `except subprocess.CalledProcessError` in a try-except clause may get the clear the given test and also removing the `check=True` in `subprocess.run` arguments. The latter is not recommended from my side and the former is also a bit weird because I guess `subprocess.run` throws this exception for other errors as well! @timothycrosley I guess one way would be give a non-zero exit code only when there is a UserWarning of no such directory for the directory scenario rather than looking in sort_attempts whether it returned true or not.

lntuition:
@anirudnits @timothycrosley 
I changed way (but purpose is equivalent) to solve this problem. I found that skipped file was filtered at `iter_source_code`, so below line shouldn't happend. Similarly check that `path` is broken at `iter_source_code`. 

https://github.com/PyCQA/isort/blob/e5ab07ddcbe79d7d11fef32ee12938fc8bbca8d6/isort/main.py#L908-L911

Also check that attempt_iterator is empty, because currently we consider only if all files are missing
as error. So by removing above checks, we can easily consider any of files are missing as error.

Anyway, i don't intended breaking exist below testcase :thinking: 

https://github.com/PyCQA/isort/blob/e5ab07ddcbe79d7d11fef32ee12938fc8bbca8d6/tests/unit/test_isort.py#L3717

Because if user pass file *a* and marked skip to *a*. I think that's what the user intended(even if no filed are attempt to sort) and isort acted correctly, so this should exited with 0.

If there's anything wrong with me or missing something. Please notice to me :smile: I'll try to fix it and find a new solution.

tests/unit/test_isort.py
```diff
@@ -3287,14 +3290,29 @@ def test_skip_glob(tmpdir, skip_glob_assert: Tuple[List[str], int, Set[str]]) ->
 
     config = Config(skip_glob=skip_glob, directory=str(base_dir))
     skipped = []  # type: List[str]
+    broken = []  # type: List[str]
```
timothycrosley, line 3293: This is not a comment on your code, just a realization, we still have type comments! We can probably switch to `broken: List[str] = []` since isort is Python 3.6+ now

timothycrosley (approved):
Echoing @anirudnits, this is awesome! Great work! :tada:  As soon as you are happy with it remove the [WIP] and I will merge it in :)

tests/unit/test_isort.py
lntuition: I thought that there was some reason  that i don't know for using old style typing in testsuite. So you mean it is just legacy, isn't it? :smile:

lntuition:
@anirudnits @timothycrosley Ready to merge :smile:  Thank you so much for your help :+1:

PR #1491, opened 2020-09-21
Check only with stdin

anirudnits:
- [x] implemented the check-only option for stdin.
- [x]  added test for the same.


timothycrosley:
@anirudnits phenomenal work :100:. Huge thank you for finding and fixing this! I made one very tiny change to make it explicit that isort would either check the file or sort it and not both: https://github.com/PyCQA/isort/pull/1491/commits/ce397f6880eca421e9a73ed87ff0d0b730053634 otherwise was perfect!

Thank you!

~Timothy

Issue #1487, opened 2020-09-17
Crash on pylint's functional test with a bogus encoding declared


Pierre-Sassoulas:
This is voluntarily a hard case to handle, here's the file : https://github.com/PyCQA/pylint/blob/master/tests/functional/u/unknown_encoding_py29.py

```
ERROR: Unrecoverable exception thrown when parsing tests/functional/u/unknown_encoding_py29.py! This should NEVER happen.
If encountered, please open an issue: https://github.com/PyCQA/isort/issues/new
Traceback (most recent call last):
  File "/usr/lib/python3.8/tokenize.py", line 342, in find_cookie
    codec = lookup(encoding)
LookupError: unknown encoding: IBO-8859-1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/bin/isort", line 8, in <module>
    sys.exit(main())
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/main.py", line 886, in main
    for sort_attempt in attempt_iterator:
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/main.py", line 875, in <genexpr>
    sort_imports(  # type: ignore
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/main.py", line 94, in sort_imports
    incorrectly_sorted = not api.sort_file(
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/api.py", line 300, in sort_file
    with io.File.read(filename) as source_file:
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/io.py", line 48, in read
    stream = File._open(file_path)
  File "/home/pierre/.cache/pre-commit/reposvah2k04/py_env-python3/lib/python3.8/site-packages/isort/io.py", line 33, in _open
    encoding, _ = tokenize.detect_encoding(buffer.readline)
  File "/usr/lib/python3.8/tokenize.py", line 381, in detect_encoding
    encoding = find_cookie(second)
[...]

PR #1496, opened 2020-09-24
[WIP] Handling Encoding Error.

anirudnits:
- [] Solves https://github.com/PyCQA/isort/issues/1487

- [x] Added tests for the same.

Its failing for the windows builds with a UnicodeEncodeError, please have a look and also review the other changes that I have made. I don't think that this is complete but still opening this PR to get suggestions and correct it accordingly :)

Thanks

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -13,6 +13,7 @@ from .exceptions import (
     FileSkipComment,
     FileSkipSetting,
     IntroducedSyntaxErrors,
+    UnsupportedEncodingError,
 )
 from .format import ask_whether_to_apply_changes_to_file, create_terminal_printer, show_unified_diff
 from .io import Empty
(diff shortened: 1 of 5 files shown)


isort/api.py
```diff
@@ -297,70 +302,76 @@ def sort_file(
...
-                                    file_output=tmp_out.read(),
-                                    file_path=actual_file_path,
-                                    output=None if show_diff is True else cast(TextIO, show_diff),
-                                    color_output=config.color_output,
-                                )
-                                if show_diff or (
-                                    ask_to_apply
-                                    and not ask_whether_to_apply_changes_to_file(
-                                        str(source_file.path)
+    try:
```
Pierre-Sassoulas, line 305: I think the try except should be elsewhere (probably on the call of this function, or , better on ` encoding, _ = tokenize.detect_encoding(buffer.readline)` in `io.py` so it's more localized and the change affect less code than that.

anirudnits: @Pierre-Sassoulas I have also added a try clause on `encoding, _ = tokenize.detect_encoding` in `isort/io.py`. The reason why I also put a try catch block here as well as `isort/main.py` is that I'm trying to throw a warning for each file with invalid encoding(if verbose flag is set) and not an error even if at least one file has valid encoding. The error is only for the case when all the files processed have invalid encodings. So I needed a way to propagate this `invalid encodings exceptions` to `isort/main.py` where I am finally deciding if to throw an error or not. `isort/io.py` and `isort/api.py` are only called per file so I couldn't come up with a way to do it there.

timothycrosley:
Closing this, as I've used your work (copying and pasting most of it) to create this PR which I've merged: https://github.com/PyCQA/isort/pull/1521 (please let me know if you see anything wrong with it, or that I missed!)

I believe the reason for the windows error, was just the default encoding being used when saving the example file with an encoding error to disk. Specifying the encoding of that file as utf8 and using write_text fixed it for me.

Thanks!

~Timothy

anirudnits:
@timothycrosley nothing wrong or missed from my side :)

PR #1501, opened 2020-09-26
Issue/1499/float to top single line multiline comment

### LINKED ISSUE #1499: Float-to-top causes imports to appear above a module's docstring
With default settings only float-to-top enabled, imports end up getting placed above a module's docstring.

For example, I don't expect the following code block to change when I run `isort` over it with float-to-top enabled.

```py3
"""I'm a module-level docstring."""

import this
```

However, the code snippet will be reformatted with the import above the docstring.

```py3
import this

"""I'm a module-level docstring."""
```

Single-line comments starting with a `#` appear to be unaffected and will remain at the top, so shebangs and encoding comments are safe.

Is this intentional behavior? If so, is there any way to disable this?

Thank you!

timothycrosley:
Fixes #1499: updates isort to correctly detect beginning of single line multi line style comment, and place imports below it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Find out more about isort's release policy [here](https://pycqa.github.io/isort/
   - Fixed #1461: Quiet config option not respected by file API in some circumstances.
   - Fixed #1482: pylama integration is not working correctly out-of-the-box.
   - Fixed #1492: --check does not work with stdin source.
+  - Fixed #1499: isort gets confused by single line, multi-line style comments when using float-to-top.
 
 ### 5.5.3 [Hotfix] September 20, 2020
   - Fixed #1488: in rare cases isort can mangle `yield from` or `raise from` statements.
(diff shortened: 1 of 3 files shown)


char101:
How do I preserve the newline between the module comment and the import lines?

timothycrosley:
@char101, the simplest answer to this, is that if you don't want imports to float to the very top - don't use that option (the initial intended usage was to use it every once in a while as a one-off after you've added imports in the middle of the file to keep from losing flow). However, I do think the behaviour could be improved in the case that there is an import at the top that already demonstrates a desire to have one extra line after the top comment - I just think that is a separate issue - I filed it here: https://github.com/PyCQA/isort/issues/1502

Thanks!

~Timothy

char101:
I haven't upgraded isort in a few months and with the _old_ isort, float to top is the default behavior, and I have been using it as a convenience to add new imports on top of a function without having to scroll to the top of the file, isort will then move the imports to the top of the file. So to me it feels like an unnecessary regression. Also with the old isort, as far as remember it keeps a single blank line between the module comment and the import lines.

PR #1510, opened 2020-09-30
Added tests for different cli flag with stdin

anirudnits:
Added tests for several CLI flags with stdin. Have a look and highlight if I missed any prominent ones, I'll add those then.


timothycrosley:
This is fantastic! Thank you so much!

PR #1516, opened 2020-10-01
Fixes bug in isort-assignments sorting.

### LINKED ISSUE #1515: Bug in isort:assignments sorting
The `isort:assignments` directive is sorting by the value of the assignments instead of the variable name.

See ``key=lambda item: item[1]`` instead of ``key=lambda item: item[0]`` in:

https://github.com/PyCQA/isort/blob/develop/isort/literal.py#L33

The tests that cover this part aren't catching the error:

https://github.com/PyCQA/isort/blob/develop/tests/unit/test_ticketed_features.py#L445 

because the values assigned are sorting in the same order as the variable names.
ie. `1, 2, x` assigned is the same order as `a, b, d`.

_Note: bug found while trying to add more tests to increase test coverage missing when enabling branch coverage._

sztamas:
Fixes #1515.

Bonus: increases branch coverage of literals.py to 100%

### DIFF
--- a/isort/literal.py
+++ b/isort/literal.py
@@ -21,17 +21,18 @@ type_mapping: Dict[str, Tuple[type, Callable[[Any, ISortPrettyPrinter], str]]] =
 
 
 def assignments(code: str) -> str:
-    sort_assignments = {}
+    values = {}
     for line in code.splitlines(keepends=True):
-        if line:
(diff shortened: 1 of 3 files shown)


timothycrosley:
Great catch and fix! Thank you!

PR #1520, opened 2020-10-03
Adds characterization tests for float_to_top and on/off.

sztamas:
Adds characterization tests for float_to_top and on/off.


timothycrosley (approved):
Looks great! Thank you!

PR #1524, opened 2020-10-04
Corrected a misplaced test and added another.

anirudnits:
Sorry I previously misplaced the `--only-modified` stdin test in the `test_unsupported_encoding` function, so just corrected that. Also to increase coverage, added another test with `--only-modified` and `--check-only` flags with modified files.


timothycrosley:
Awesome, thank you!

PR #1526, opened 2020-10-04
Fix warnings can't surpressed with quiet.

### LINKED ISSUE #1525: Some warnings can't be disabled with --quiet
There are some early warnings in `settings.py:Config.__init__` that can't be disabled with `quiet` because they conditionals use `self.quiet`, but `self.quiet` isn't set until later in the method.

Ex. deprecation warning for using `not-skip` shows up event though I used `--quiet`.

```console
▶ isort . --check  --quiet
/home/sztamas/Work/isort/isort/settings.py:429: UserWarning: W0503: Deprecated config options were used: not_skip.Please see the 5.0.0 upgrade guide: bit.ly/isortv5.
  warn(
```

Code for the example above

https://github.com/PyCQA/isort/blob/develop/isort/settings.py#L428

but there are a few other warnings as well.

sztamas:
Fixes #1525

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -266,6 +266,11 @@ class Config(_Config):
             super().__init__(**config_vars)  # type: ignore
             return
 
+        # We can't use self.quiet to conditionally show warnings before super.__init__() is called
+        # at the end of this method. _Config is also frozen so setting self.quiet isn't possible.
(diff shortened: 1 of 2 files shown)


timothycrosley:

PR #1527, opened 2020-10-04
Improve PyCharm Profile

saippuakauppias:
PyCharm by default adds 2 lines after imports using the "Optimize Imports (control+option+o)" action.


timothycrosley:
Thank you!

Issue #1536, opened 2020-10-07
Provide API and CLI to list and stream imports 


timothycrosley:
isort is now exceptional at finding all imports, even nested ones, inside of Python source files of any size quickly. I think it would be beneficial to expose this functionality for downstream use, both from the CLI and via the isort API. From the CLI instead of adding yet another additional flag, it may make sense to make an entire separate command (such as as isort_identify_imports or similar.)

timqsh:
Hi! Looking into this issue.

PR #1545, opened 2020-10-08
Provide API and CLI to list and stream imports

timqsh:
Issue #1536 
I have not yet added new CLI command mentioned in issue. Need to check how it is done in argparse before.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -1,4 +1,3 @@
-import os
 import textwrap
 from io import StringIO
 from itertools import chain
@@ -73,7 +72,12 @@ def process(
     all_imports: List[str] = []
     if imports_only:
         _output_stream = output_stream
-        output_stream = open(os.devnull, "wt")
+
+        class DevNull(StringIO):
+            def write(self, *a, **kw):
+                pass
+
+        output_stream = DevNull()
 
     if config.float_to_top:
         new_input = ""
--- a/isort/core.py
+++ b/isort/core.py
@@ -1,3 +1,4 @@
+import os
 import textwrap
 from io import StringIO
 from itertools import chain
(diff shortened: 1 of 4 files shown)


timqsh:
So, I looked into making a new CLI command. It seems like it will take a little more time for me to do it. @timothycrosley maybe we split this feature into 2 PRs, so I wouldn't worry about rebasing my current code with develop?

Can you review this feature in current state - with new commands exported only in API?

timothycrosley:
@timqsh that sounds great to me! :eyes:  reviewing now

isort/core.py
```diff
@@ -395,6 +402,11 @@ def process(
         for output_str in verbose_output:
             print(output_str)
 
+    if imports_only:
+        output_stream.seek(0)
```
timothycrosley, line 406: Will this work if output_stream is sys.stdout?

timothycrosley:
One question, overall this looks really good, thank you!

isort/core.py
timqsh: You're right, it won't work with sys.stdout. Fixed it.

isort/core.py
```diff
@@ -68,6 +70,10 @@ def process(
     stripped_line: str = ""
     end_of_file: bool = False
     verbose_output: List[str] = []
+    all_imports: List[str] = []
+    if imports_only:
+        _output_stream = output_stream
+        output_stream = open(os.devnull, "wt")
```
timqsh, line 76: DeepSource doesn't like this line because ```open``` without ```with``` block. Can change it to ```output_stream = StringIO()``` but it will be less efficient in big files: all stream would accumulate in memory.

timqsh: another possible option
```python
        class DevNull:
            def write(self, *a, **kw):
                pass

        output_stream = DevNull()
```

timothycrosley: I like that option, I'd even go so far as:

```python
class DevNull:
    def write(*a, **kw):
        pass

output_stream = DevNull
```

timqsh: This works, but mypy isn't happy about type of ```output_stream``` =)
Pushed fix, that satisfies linters and should work fine.

timothycrosley:
:rocket: Great work! Merged :)

Issue #1494, opened 2020-09-22
Cython .pxd files not being picked up


cjdsellers:
I may be doing something wrong. However, on running the command `isort .` isort is not detecting the imports inside `.pxd` files.

timothycrosley:
Currently, isort only automatically picks up .pyx files: https://pycqa.github.io/isort/docs/configuration/options/#supported-extensions. Though I think it makes sense to change this, and will make sure pxd is included by default in the next minor release. 

For now, you can also adjust this behaviour, either in your config or by specifying the desired extensions on the command line:

`isort . --ext py --ext pyi --ext pyx --ext pxd`

timothycrosley:
The default has been changed in develop and will be pushed to PyPI in release 5.6.0

timothycrosley:
Update: this has just been released to PyPI in 5.6.0 release of isort: https://pycqa.github.io/isort/CHANGELOG/#560-october-7-2020

Thanks!

~Timothy

cjdsellers:
Thanks Timothy

Much appreciated!

PR #1549, opened 2020-10-08
Let pre-commit sort pxd files too

MarcoGorelli:
In #1494, it looks like support for `.pxd` files was added. However, these are currently ignored when running isort in pre-commit because there is `types: [python]`

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,6 @@
     entry: isort
     require_serial: true
     language: python
-    types: [python]
+    files: '.pxd$|.py$'
+    types: [file]
     args: ['--filter-files']
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
     entry: isort
     require_serial: true
     language: python
-    types: [python, cython, pyi]
+    types: [python]
     args: ['--filter-files']
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,5 +4,4 @@
     require_serial: true
     language: python
     files: '.pxd$|.py$'
-    types: [file]
     args: ['--filter-files']
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,5 @@
     entry: isort
     require_serial: true
     language: python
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
Hi @timothycrosley - my understanding of `types` is that the intersection of them is taken, so if you have `types: [python, cython, pyi]` then the result will be that nothing will get sorted.

From the [pre-commit docs](https://pre-commit.com/#filtering-files-with-types)

> types and files are evaluated with AND when filtering. Tags within types are also evaluated using AND.

timothycrosley:
Yep! I came to the same conclusion myself as well. This is really unfortunate however, because it makes it impossible for isort to add seamless support for Cython files without introducing a regression on non-extensioned Python files. Looks like there is an open issue to add Or type support: https://github.com/pre-commit/pre-commit/issues/607

MarcoGorelli:
Ah, I didn't think that was possible, sorry!

```bash
$ echo '#!/usr/bin/python' > nonext
$ sudo chmod +x nonext 
$ identify-cli nonext 
["executable", "file", "python", "text"]
```

MarcoGorelli:
Would it work for you to just have `types: [files]` until pre-commit/pre-commit#607 is in, and then after that `types_or: [python, cython, pyi]`?

timothycrosley:
@MarcoGorelli,

Thanks for highlighting this area of improvement! 

> Would it work for you to just have types: [files] until pre-commit/pre-commit#607 is in, and then after that types_or: [python, cython, pyi]?

The only problem with that is than it will cause a performance regression and maybe even behave in unexpected ways / break non-Python files. That said I would be amenable to that change if we could find some larger users of isort with pre-commit and then create an additional integration test suite that ran pre-commit against those users repositories and could confirm it didn't cause any issues. I think this would be a long-term win for isorts maintainability overall as well. 

Void of that, I think the most prudent thing to do would be to update the documentation of pre-commit usage to such using a hook per a file-type as shown in this thread: https://github.com/pre-commit/pre-commit/issues/1436#issuecomment-624219819

Example: 
```
- hooks:
      - id: isort
        types: [python]
      - id: isort
        types: [cython]
      - id: isort
        types: [pyi]
```

Thanks!

~Timothy

MarcoGorelli:
That's a good idea, have updated accordingly - thanks for this and for your awesome work!

Issue #1533, opened 2020-10-06
--exclude CLI flag to exclude files and directories from being scanned.


anirudnits:
Basically like the `--exclude regex_pattern` flag in *black* and *flake8* etc, where directories and files are matched with the regular expression provided to check whether they should be excluded for recursive searches.

timothycrosley:
I *think* isort already supports this via [skip-glob](https://pycqa.github.io/isort/docs/configuration/options/#skip-glob) in which case the action item for this ticket would change a bit into:

- Improve the documentation so that skip-glob is a more obvious feature.
- Add an `--exclude` alias to the CLI for skip-glob

anirudnits:
@timothycrosley yeah I guess that the documentation could be improved for skip-glob, I at least didn't get this at first read. Also making an alias will be good as many tools like *isort*  have this `--exclude` flag to skip files/directories from the CLI.

sztamas:
:+1: for the `--exclude` alias.

Maybe also consider deprecating `skip-glob`?

timothycrosley:
I think deprecating is a good idea, long term it makes sense to only have one obvious and correct way to change any knob, so we could add a deprecation notice in 5.7.0 and then remove the old behaviour in 6.0.0

timqsh:
I found that skip-glob option is confusing, because it doesn't work like real glob. It simply treats single * as .* in regexp. Isort currently uses fnmatch with this behavior documented here [fnmatch](https://docs.python.org/3/library/fnmatch.html)

Here some examples of [user configs](https://github.com/search?p=2&q=skip_glob&type=Code) . As you can see some of them relying on single * capturing all while others using double ** like in glob.

Therefore behavior can't be changed to Unix like glob in minor release because it will break current users with configs like this https://github.com/fKunstner/qparse/blob/86a7191c4a9d474a7bad6ea637971b3625a25321/.isort.cfg#L7 .

I found this issue when integration tests on my PR were failing on apache/airflow repo. They also used ignore-glob as real glob which lead to inconsistencies with pre-commit. I made [PR](https://github.com/apache/airflow/pull/11613) in their repo to fix it.

We should consider it when implementing --exclude flag. I see 2 options:
1) Leave current behavior. Confusion might go away because there will be no glob word in --exclude flag
2) Implement --exclude differently, so that Unix style with single * and double ** is supported. [python's glob module docs for reference](https://docs.python.org/3/library/glob.html)

govindrai:
[...]

PR #1567, opened 2020-10-17
CLI to get imports from file

### LINKED ISSUE #1536: Provide API and CLI to list and stream imports

timqsh:
closes #1536

### DIFF
--- a/isort/__init__.py
+++ b/isort/__init__.py
@@ -5,6 +5,7 @@ from .api import check_code_string as check_code
 from .api import (
     check_file,
     check_stream,
+    get_imports_file,
     get_imports_stream,
     get_imports_string,
     place_module,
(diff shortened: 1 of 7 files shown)


timqsh:
It looks like current master of apache/airflow repo doesn't pass integration test

timothycrosley:
@timqsh Great work here! It does look like there was an interim issue with the integration test that has sense been resolved. The remaining errors seem to be because the test case is looking for windows newline style on windows, whereas the output is always unix style.

timqsh:
apache/airflow integration test is still broken. I wrote why in https://github.com/PyCQA/isort/issues/1533

Issue #1566, opened 2020-10-16
Import format changes based on line ending type


VakarisZ:
Import 
```
from impacket.smb3structs import (
    SMB2_CREATE, SMB2_FLAGS_DFS_OPERATIONS, SMB2_IL_IMPERSONATION, SMB2_OPLOCK_LEVEL_NONE, SMB2Create,
    SMB2Create_Response, SMB2Packet)
``` 
checked with command `python -m isort . -c -l 120 --wl 120` passes on dos line endings, but with unix line endings it fails because it expects
```
from impacket.smb3structs import (SMB2_CREATE, SMB2_FLAGS_DFS_OPERATIONS, SMB2_IL_IMPERSONATION, SMB2_OPLOCK_LEVEL_NONE,
                                  SMB2Create, SMB2Create_Response, SMB2Packet)
```
Visually, both line endings look the same, so I'd expect them to be treated the same (either as 1 or as 2 characters).

anirudnits:
From my brief inspection, I believe the issue lies here https://github.com/PyCQA/isort/blob/6807cfbaeb7d25d2b92c4e0efce41371ffff9777/isort/io.py#L44
with `newline=""`. From python docs:

`When reading input from the stream, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n',`
`'\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newlines mode is enabled,but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.`

I'm not sure if this the exact reason and I would delve a bit deeper when I get time.

areeh:
@anirudnits 
Bear in mind that setting relates to isort's history of changing newlines, so while that might fix the 1 vs 2 character issue, it might bring back the changing line endings issue. See eg: #493 #671

areeh:
For reference, here's how Black appears to solve both issues at once (edited heavily for brevity):

```
with open(src, "rb") as buf:
    src_contents, encoding, newline = decode_bytes(buf.read())

dst_contents = format_file_contents(src_contents)

with open(src, "w", encoding=encoding, newline=newline) as f:
    f.write(dst_contents)
```

```
def decode_bytes(src: bytes) -> Tuple[FileContent, Encoding, NewLine]:
    """Return a tuple of (decoded_contents, encoding, newline).
    `newline` is either CRLF or LF but `decoded_contents` is decoded with
[...]

PR #1576, opened 2020-10-23
wip #1566

areeh:
wip #1566

TODO: 
Add reproducer test, see eg: https://github.com/PyCQA/isort/issues/1566#issue-723204823
Fix `sort_stream`-related code. One issue is that out is sometimes set to `sys.stdout`

### DIFF
--- a/tests/unit/test_api.py
+++ b/tests/unit/test_api.py
@@ -8,6 +8,7 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 from isort import api
+from isort.api import detect_newline
 from isort.settings import Config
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_api.py
```diff
@@ -10,14 +11,15 @@
 from isort.settings import Config
 
 imperfect_content = "import b\nimport a\n"
-fixed_content = "import a\nimport b\n"
+fixed_content = f"import a\nimport b\n"
 fixed_diff = "+import a\n import b\n-import a\n"
```
areeh, line 15: revert

PR #1587, opened 2020-10-30
Profile: follow black behaviour with regard to gitignore

### LINKED ISSUE #1585: Make `--gitignore` or `skip_gitignore` default behaviour (for black and all other profiles)
Black ignores files not version-controlled https://github.com/psf/black/blob/master/src/black/__init__.py#L613, as such profile black should imply https://pycqa.github.io/isort/docs/configuration/options/#skip-gitignore being true here https://github.com/PyCQA/isort/blob/1e0eb0d2aba1e5384724bac410efde44cbd57fbf/isort/profiles.py#L4-L11.

sgaist:
Black for Python ignores the files listed in .gitignore.

Update the profile to follow that

Fixes #1585

### DIFF
--- a/isort/profiles.py
+++ b/isort/profiles.py
@@ -8,6 +8,7 @@ black = {
     "use_parentheses": True,
     "ensure_newline_before_comments": True,
     "line_length": 88,
+    "skip_gitignore": True,
 }
 django = {
     "combine_as_imports": True,


timothycrosley:
Thank you!

gaborbernat:
Any reason why this profile doesn't read the line length and adjusts it similarly?

timothycrosley:
I'm a little hesitant to do that - though not 100% against the idea - simply because there is no guarantee that isort and black are within the same config file, and I could see that leading to some very confusing user interaction where isort doesn't find the black setting for some reason or another. It's a bit more manual but at least comes with a consistent explanation and less magic to say that the profile simply provides the same defaults as black, anything beyond that must be set in both.

gaborbernat:
What if they both are in the toml file? 🤠

PR #1606, opened 2020-12-08
Fix description of config lookup

qsantos:
As visible in [code](https://github.com/PyCQA/isort/blob/develop/isort/settings.py#L626-L628).


timothycrosley:
Thanks for helping to clarify this in the documentation :100: !

PR #1609, opened 2020-12-14
Set pre-commit language_version to python3

jdufresne:
For details on how pre-commit chooses a default language version, see:
https://pre-commit.com/#overriding-language-version

> For each language, they default to using the system installed language

On some platforms this could be Python 2. As isort is Python-3-only,
this should be enforced in the pre-commit hook. As this isn't currently
listed, some projects have been forced to override it.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,5 +3,6 @@
     entry: isort
     require_serial: true
     language: python
+    language_version: python3
     types: [python]
     args: ['--filter-files']


MarcoGorelli:
> some projects have been forced to override it.

Just out of interest, which ones? I'm just surprised it would be necessary - isort only supports Python3, and so if you're running this hook you must also be running Python3, so the hook will using Python3.

E.g.:
```
$ python --version
2.7.18
$ mkdir mcve
$ cd mcve/
$ python3.8 -m virtualenv venv
$ . venv/bin/activate
$ nano .pre-commit-config.yaml
$ pip install pre-commit
$ git init
$ pre-commit run -a
python version...........................................................Failed
- hook id: python-version
- exit code: 1

3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0]
```

where `.pre-commit-config.yaml` is
```
repos:
    - repo: local
      hooks:
        - id: python-version
          name: python version
          language: python
          entry: python -c 'import sys; print(sys.version); sys.exit(1)'
          pass_filenames: false
          always_run: true
```

jdufresne:
pip-tools is the most recent project that prompted me to supply this PR.

https://github.com/jazzband/pip-tools/blob/bc9695c604314e62f6379b50f003dbe0f9878c9d/.pre-commit-config.yaml#L11

jdufresne:
FWIW, black also includes this configuration:

https://github.com/psf/black/blob/a522aa45c65bb896637d0add27b4c2a9fec6b976/.pre-commit-hooks.yaml#L6

timothycrosley:
Thanks for fixing this! It's important since sort can run against any version of Python targeting source files (even Python 1.0!) but now requires Python 3 to run

PR #1625, opened 2020-12-30
Update indent option doc

gpakosz:
The current rendering of the `indent` option documentation is a bit off:

image

This PR tries to improve things by using `<code>` and `&nbsp;`.
It renders correctly when viewing the `options.md` file on GitHub so I expect it's going to render correctly when visiting https://pycqa.github.io/isort/docs/configuration/options/#indent

Hope that helps

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -263,10 +263,10 @@ multi_line_output = 3
 
 ## Indent
 
-String to place for indents defaults to "    " (4 spaces).
+String to place for indents defaults to "&nbsp;&nbsp;&nbsp;&nbsp;" (4 spaces).
 
 **Type:** String  
-**Default:** `    `  
+**Default:** <code>&nbsp;&nbsp;&nbsp;&nbsp;</code>  
 **Python & Config File Name:** indent  
 **CLI Flags:**
 


timothycrosley:
Hi @gpakosz, 

Thanks for pointing out this doc deficiency! I agree the suggested change is a marked improvement, unfortunately I can't accept this PR as that document file is auto-generated so will be undone automatically on the next doc push. The fix would need to be done in some generic way in https://github.com/PyCQA/isort/blob/develop/scripts/build_config_option_docs.py to live across documentation releases.

Thanks!

~Timothy

gpakosz:
Woops sorry.

Not sure how to proceed then. Changing the help string in `main.py` line `435` doesn't sound like a good idea.
And I'm not sure how to alter `build_config_option_docs.py` so that it detects this case.

gofr:
The 4 spaces in the help of the `--indent` option in main.py are squashed into just one space by argparse as well when using `--help` because of how the default formatter works. I think they're only showing up as 4 actual spaces in options.md (when viewing it as text). It's probably easier to rephrase the argument help not to have the 4 literal spaces at all.

Using `<pre><code>` would help the "Default:" value, but not sure that would really make it better.

Issue #1624, opened 2020-12-30
force_sort_within_sections is not case-insensitive


gofr:
I want my from-imports and straight-imports to be sorted together so I'd like to use `--force-sort-within-sections`. And I want my modules to be sorted case-insensitively too, which is already the default.

Sort the following with the `--force-sort-within-sections` flag:

    import a
    from Z import b, C

With a bash command that's: `echo -e 'import a\nfrom Z import b, C' | isort - --force-sort-within-sections`

### Expected result

    import a
    from Z import C, b

### Actual result

    from Z import C, b
    import a

The [`sorting.section_key`](https://github.com/PyCQA/isort/blob/8e6d383064d337813a5b021bd597e166ccd630b3/isort/sorting.py#L54) function that gets called doesn't care about the `case_sensitive` setting. It _does_ however get `order_by_type` passed in as a parameter and that lowercases the line if it's turned off. So if I call it like this: 

    echo -e 'import a\nfrom Z import b, C' | isort - --force-sort-within-sections --dont-order-by-type

I get:

    import a
    from Z import b, C

But that's still not what I want, because I _do_ want to order by type.

I see the `order_by_type` was added in 1372970. And I also see there's already a unit test covering this: https://github.com/PyCQA/isort/blob/6e5414cb72ee822d351968626bb9a37e10425857/tests/unit/test_isort.py#L2358

Is this a deliberate choice or an oversight? Should there be some code in there that lowercases just the module names (depending on `case_sensitive`)? And should `order_by_type` only lowercase the imported names, so it doesn't affect the module order?

timothycrosley:
Closing as this was fixed via your PR, thank you!

PR #1626, opened 2020-12-30
Make force_sort_within_sections respect case

gofr:
force_sort_within_sections only looked at the order_by_type
option to determine how to order imports with different case
in a section. Whether you order import names by type or not
also affected the order of the modules.

When force_sort_within_sections is used:
* ignore case on the module name if case_sensitive is false,
* ignore case on the imported names if order_by_type is false.

### DIFF
--- a/isort/sorting.py
+++ b/isort/sorting.py
@@ -54,6 +54,7 @@ def module_key(
 def section_key(
     line: str,
     case_sensitive: bool,
+    honor_case_in_force_sorted_sections: bool,
     order_by_type: bool,
     force_to_top: List[str],
     lexicographical: bool = False,
@@ -77,7 +78,11 @@ def section_key(
         line = re.sub("^import ", "", line)
     if line.split(" ")[0] in force_to_top:
         section = "A"
-    if not case_sensitive or not order_by_type:
+    # * If honor_case_in_force_sorted_sections is true, and case_sensitive and
+    #   order_by_type are different, only ignore case in part of the line.
+    # * Otherwise, let order_by_type decide the sorting of the whole line. This
+    #   is only "correct" if case_sensitive and order_by_type have the same value.
+    if honor_case_in_force_sorted_sections and case_sensitive != order_by_type:
         split_module = line.split(" import ", 1)
         if len(split_module) > 1:
             module_name, names = split_module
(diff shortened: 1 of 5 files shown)


gofr:
Addresses issue #1624

tests/unit/test_isort.py
```diff
@@ -2358,10 +2358,10 @@ def test_alphabetic_sorting_no_newlines() -> None:
 def test_sort_within_section() -> None:
     """Test to ensure its possible to force isort to sort within sections"""
     test_input = (
-        "from Foob import ar\n"
         "import foo\n"
         "from foo import bar\n"
         "from foo.bar import Quux, baz\n"
+        "from Foob import ar\n"
```
timothycrosley, line 2364: I'm a little hesitant to make a change that would change the default styling with a single option set, as it would break isort's formatting guarantee: https://pycqa.github.io/isort/docs/major_releases/release_policy/#formatting-guarantees. Is there any chance you could restructure the change to not affect the existing test cases, possibly being behind it's own flag - and then the project can revisit in isort 6.0.0 next year?

gofr: Yeah, I realized this was a bit problematic... Not sure how to do it without a flag, so I'll add a flag.

isort/sorting.py
```diff
@@ -54,6 +54,7 @@ def module_key(
 def section_key(
     line: str,
     case_sensitive: bool,
+    honor_case_in_force_sorted_sections: bool,
     order_by_type: bool,
     force_to_top: List[str],
     lexicographical: bool = False,
```
gofr: This function is getting rather a lot of arguments now. This is only called in one place, and all arguments are part of the `config`. Should this function be changed to just take a `config` argument, like `module_key` above also does?

`module_key` also needs to be able to override the config since it's called in some different contexts. That's not the case for this function. Is there still a good reason to keep any arguments separate?

And if you think it's a good idea to fix this, should I do it in this pull request or a separate one? (I think probably a separate one.)

timothycrosley (approved):
:100: LGTM! Great work!

isort/sorting.py
timothycrosley: I think your instinct is correct that it does make sense to refactor it to use the immutable config directly. I have a preference for more smaller PRs as they are easier to review, but the project has no official stance and I will merge either as long as the tests pass and the changeset looks good. Honestly, just thrilled to have the contributions, thank you!

gofr:
Great, thanks. :) I'll create a separate pull request for the minor refactoring to `section_key`.

PR #1629, opened 2021-01-02
Simplify section_key helper function arguments

gofr:
sorting.section_key() is only used once and all its arguments
come from Config. Make it take a single config parameter instead
of a bunch of individual settings.

Follow-up to #1626


timothycrosley:
:rocket: This is a nice improvement, thank you!

PR #1649, opened 2021-02-03
Indicate type hint support with a py.typed file

### LINKED ISSUE #1648: Add py.typed to make isort PEP-561-compatible
**Problem**: mypy can't find type hints when importing `isort`:

```bash
$ mypy .
my_module.py:1: error: Skipping analyzing 'isort': found module but no type hints or library stubs
my_module.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 608 source files)
```

**Solution**: add a py.typed file to make `isort` PEP-561-compatible
**Alternatives**: add `isort` to typeshed, or worser workarounds

Shamelessly snugged the above from psf/black#1316 , the fix can be as simple as psf/black#1395

asfaltboy:
fix #1648


asfaltboy:
Note: looks like the CVA reported by safety is unrelated (might not even be [a tornado issue][1] apparently)

[1]: https://github.com/tornadoweb/tornado/issues/2981

timothycrosley:
Awesome thanks for fixing this! I've ignored the tornado warning as I agree with your, and linked assesment, that it is being wrongly flagged in this case

Issue #1640, opened 2021-01-22
Isort violates line_length with include_trailing_commas setting in special cases


micahjsmith:
It seems possible to get isort to violate `line_length` when using `include_trailing_commas`.

Consider the following example:

test.py:
```
from mylib import aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, bbbbbbbbb
```

.isort.cfg:
```
[settings]
include_trailing_comma = True
line_length = 79
multi_line_output = 4
```

Then running isort produces

```
from mylib import (
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, bbbbbbbbb,)
```

You will notice if you inspect this in your editor that the second line of the fixed file is 80 characters, violating line_length.

I would instead expect behavior where `bbbbbbbbb` is placed on line 3, such that no line is over the max length:
```
from mylib import (
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
    bbbbbbbbb,)
```

- isort version: 5.7.0
- python version: 3.7.7

Similar to #1634

gofr:
Yeah, I mentioned this case in https://github.com/PyCQA/isort/issues/1634#issuecomment-757344253. My pull request there would solve both these issues (but still needs a bit of work like tests and docs).

As I understand it this behavior is intentional, but I don't really understand why it would be. :grimacing: Well, backwards-compatibility may be an issue.

The behavior was "introduced" in 92f53fc47 (to fix some other edge cases). It did not take into account the use of `include_trailing_comma`. The code was last changed for issue #654 which was I think really a partial fix for the `include_trailing_comma` bug.

micahjsmith:
I looked over the other thread first, and yeah the example there is very similar but not the same. In this case, it actually does violate line_length, no? rather than just producing surprising behavior that may be acceptable.

Obviously related implementation issues and I'm sure that your PR there would resolve this issue also. Thanks for your work there.

gofr:
[...]

Issue #1634, opened 2021-01-08
Possible bug when length of line is exactly line_length under certain conditions


mhiley:
Hello, I think there may be a bug where under certain conditions, isort wants to wrap an import line with a length that is exactly equal to `line_length`.

Example (`line_length` is set to 100, `wrap_length` not set):

When I have an import of this form that's exactly 100 characters long:
```python
from a import aaaaaaaaaaa, bbbbbbbbbbbb, cccccccccccc, ddddddddddddddddddddddd, eeeeeeeeeeeeeeeeeeee
```
then isort leaves it alone.

However, when I have an import of this form, where the longer line is exactly 100 characters long:
```python
from a import (
    aaaaaaaaaaa, bbbbbbbbbbbb, cccccccccccc, ddddddddddddddddddddddd, eeeeeeeeeeeeeeeeeeeeeeeeeeeeee
)
```
...then isort wants to wrap the last import (`eeeeeeeeeeeeeeeeeeeeeeeeeeeeee`) onto the next line.

This seems like a bug to me, or am I interpreting the exact meaning of `line_length` incorrectly? Thank you!

Here is my full `.isort.cfg`:
```
[settings]
atomic=1
indent=4
use_parentheses=1
line_length=100
lines_between_types=0
multi_line_output=5
no_sections=1
from_first=1
```

gofr:
The bug happens here:
https://github.com/PyCQA/isort/blob/98fad1127fc0771959ab39303548e3b50a1e7b5f/isort/wrap_modes.py#L204

I think there's a bit of a mismatch between what the two "vert-grid-grouped" multi-line wrapping modes sound like they should do and what they actually do.

Given the name of mode 6-vert-grid-grouped-no-comma, mode 5-vert-grid-grouped sounds like it includes a trailing comma on the last line while mode 6 doesn't. But neither of them control commas at all. The `include_trailing_comma` option does.

In this case what happens is that mode 5 makes sure there's room for the comma, but then doesn't actually add that comma. 

If instead you use `multi_line_output=6`, but also set `include_trailing_comma=1` you get the opposite wrong behavior: A trailing comma is included but the line is not wrapped, creating a line of 101 characters even though the line length is 100.

Two potential ways of fixing this I can think of:
* Merge modes 5 and 6, dropping one of them or making them aliases. Let `include_trailing_comma` control the comma.
[...]

PR #1653, opened 2021-02-05
Issue/1634/vert grid trailing chars 2

gofr:
This is a more complete replacement for my earlier PR #1635 to address issues #1634 and #1640. It builds on top of #1651.

Also see alternative PR #1652, which only addresses #1640.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,20 +221,9 @@ from third_party import (
 )
 ```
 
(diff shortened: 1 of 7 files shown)


timothycrosley:
This looks great! Thanks for the time and care put into this!

Issue #1650, opened 2021-02-04
Quiet does not suppress skipping files messages


AustinScola:
### Description
With `quiet=True` in `.isort.cfg`, a message about the number of skipped files is still output to stdout.

### Example

File tree:
```.
├── foo.py
└── .isort.cfg
```
The contents of .isort.cfg:
```
[settings]
quiet=True
skip_glob=foo.py
```

The result of running isort:
```
$ python3 -m isort .
Skipped 1 files
```

AustinScola:
Thanks!

PR #1655, opened 2021-02-05
Don't print skipped/broken files with quiet config setting

gofr:
When using --quiet, no warnings are printed about skipped and
broken paths. But you do get warnings when using the "quiet"
setting in a config file.

Be consistent and don't print the warnings then either.

Fix for #1650.


timothycrosley:
Thanks for fixing this :tada:

Issue #1659, opened 2021-02-09
Add option to disable force_sort_within_sections for relative imports


jwodder:
It appears to be isort's opinion that statements of the form "`from . import foo`" are "straight-style" imports (like `import pathlib`) rather than "from-style" imports (like `from pathlib import Path`).  As a result, when `force_sort_within_sections` is in effect, these statements get sorted among "`from .foo import bar`" statements as though they were written "`from .import import foo`".  I don't like this; I would like to have my `force_sort_within_sections` and eat it, too.

Specifically, my goal is for isort to sort the following statements as shown:

```python
import arithmetic
from botany import flower
import chemistry

from . import foo
from .apple import bar
from .banana import baz
from .. import quux
from ..asparagus import gnusto
from ..broccoli import cleesh
```

I can get the relative imports to be sorted the way I want by setting `reverse_relative = true`, but then I also need to set `force_sort_within_sections = true` in order to get the absolute imports to sort the way I want, and that ends up changing the ordering of the relative imports.  I would thus like to request a configuration option or setting for achieving the above ordering, such as by making `force_sort_within_sections` not apply to relative imports.

gofr:
It's even more instructive if you replace `banana` by `mango` (or any other fruit that sorts alphabetically after `import` :wink:) in that example. Then with `force_sort_within_sections = true` they get sorted like this:

    from .apple import bar
    from . import foo
    from .mango import baz

That's definitely not right.

timothycrosley:
Closing this as it is fixed by @gofr's pr here https://github.com/PyCQA/isort/pull/1664 and the addition of the `sort_relative_in_force_sorted_sections` option

Issue #1395, opened 2020-08-19
conforming to flake8 appnexus


JackDra:
appnexus requires relative imports to have current directory above parent directory. E.g.

```
from .fileA import a_var
from ..fileB import b_var
```

but isort sorts them in order:

```
from ..fileB import b_var
from .fileA import a_var
```

Is there any way to reverse this condition? (i might have missed it in the docs somewhere...). 

Currently it is reversed with the current tox.ini setup:

```
[flake8]
ignore = E123,E133,E226,E241,E262,E265,W503,W504
application-import-names = ...
application-package-names = ...
import-order-style = appnexus
exclude = build

[isort]
force_sort_within_sections = true
order_by_type = false
case_sensitive = false
multi_line_output = 5
sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,APPLICATION,LOCALFOLDER
known_third_party = ...
known_first_party = ...
known_application = ...
lines_after_imports = 2
no_lines_before=LOCALFOLDER
```

timothycrosley:
Hi @JackDra,

This is indeed supported by using the `reverse_relative` config option, see: https://timothycrosley.github.io/isort/docs/configuration/options/#reverse-relative.

Thanks!

~Timothy

JackDra:
Ah, I wasn't sure if that was for that.... My bad

It seems to still not be working. I found the issue to be a conflict(?) between `force_sort_within_sections` and `reverse_relative`.
If I have both on, it ignores reverse_relative. Is this just a limitation with the logic? or a bug?

JackDra:
From what I can tell, it seem like here:

https://github.com/timothycrosley/isort/blob/66bc89ff5290d28571b1d8af620644bcba5084f9/isort/output.py#L125-L134

The ordering of the '..' and '.' imports are flipped back to the regular ordering.

adding print statements to the `sorting.naturally` function called:
```
def naturally(to_sort: Iterable[str], key: Optional[Callable[[str], Any]] = None) -> List[str]:
    """Returns a naturally sorted list"""
    if key is None:
        key_callback = _natural_keys
    else:

        def key_callback(text: str) -> List[Any]:
            return _natural_keys(key(text))  # type: ignore

    show_sort = False
    for this_sort in to_sort:
        if any(['..' in ikey for ikey in to_sort]):
            show_sort = True
    if show_sort:
        print(to_sort)
        print(sorted(to_sort, key=key_callback))
        print()
[...]

PR #1664, opened 2021-02-11
Sort relative imports correctly with force_sort_within_sections

gofr:
Relative import sort order when using force_sort_within_sections
was inconsistent with the order without that setting. Change the
force_sort_within_sections sort order to match.

This fixes the relative import ordering issues noted in #1659.

Even though all the existing tests still pass (with only one minor test change), the change to the ordering might be significant enough that this needs to be hidden behind a new setting until 6.x?

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -669,6 +669,14 @@ def _build_arg_parser() -> argparse.ArgumentParser:
         "Without this option set, `--order-by-type` decides module name ordering too.",
     )
     section_group.add_argument(
+        "--srss",
+        "--sort-relative-in-force-sorted-sections",
+        action="store_true",
+        dest="sort_relative_in_force_sorted_sections",
+        help="When using `--force-sort-within-sections`, sort relative imports the same "
+        "way as they are sorted when not using that setting.",
+    )
+    section_group.add_argument(
         "--fass",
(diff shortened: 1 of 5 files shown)


timothycrosley:
> Even though all the existing tests still pass (with only one minor test change), the change to the ordering might be significant enough that this needs to be hidden behind a new setting until 6.x?

I think this intuition is correct as it could change existing formatting - and honestly may always stay behind a flag (even if the default value is reversed) as strictly speaking when we get into these more niche combinations - there is no "correct" these are just style preferences - and the most important thing is that they stay the same for the same settings given.

gofr:
The current behavior was only added relatively recently, though, in 5.5.0 for #1395. The change that caused in the two tests I added here was about as big as the change now caused by this PR. And the test you added for #1395 still passes.

It seems to me like this is on par with that fix in that it clears up some undefined/untested behavior in a niche combination (as you rightly say!).

It feels like a bit of a waste to make this configurable. But I also see this may just be the price you have to pay for having successful software! :wink:

timothycrosley:
@gofr Thanks so much for this great work! And sorry I've been so slow to merge it. I agree long term it probably doesn't need an additional option, and I'll make sure to deprecate in the next release.

Thanks!

~Timothy

Issue #1665, opened 2021-02-12
Caveat or remove recursive glob suggestion in README.md


dpoznik:
The instructions [suggest](https://pycqa.github.io/isort/#using-isort) that running `isort .` is equivalent to running `isort **/*.py`, but I was finding this not to be the case. Rather, `isort .` fixed many files that `isort **/*.py` left intact. This is because recursive glob is not enabled by default on many systems. For the equivalence to be true, one must:
* (macOS users) [upgrade bash](https://itnext.io/upgrading-bash-on-macos-7138bd1066ba)
* [set `globstar`](https://unix.stackexchange.com/questions/49913/recursive-glob): `shopt -s globstar`

I think it would be worth either noting this or just removing "which is equivalent to: `isort **/*.py`" from the instructions. Otherwise, users may incorrectly think they've fully `isort`ed their projects when they haven't.

Thanks for the great tool!

PR #1672, opened 2021-02-17
Add clarifying note on globstar to README.md

dpoznik:
For `isort .` and `isort **/*.py` to be equivalent, one must have `globstar` enabled. This PR adds a small clarifying note to that effect.

Addresses #1665


timothycrosley:
This is great! Thanks!

PR #1673, opened 2021-02-17
Fix not replacing the source file if only literals are changed

yhlam:
When there is no change to the imports but the literal are changed (i.e. sorted by isort), isort doesn't replace the source file with the sorted file. This pull request fixes the problem.


timothycrosley (approved):
This looks great, thanks!

PR #1676, opened 2021-02-24
fix :Failed to pull configuration information from pyproject.toml

dongfangtianyu:
When I configured isort in the `pyproject.toml`, an error occurred: `Failed to pull configuration information from pyproject.toml`.

After further analysis, it is because pyproject.toml contains many non-ASCII characters, which causes the system default encoding to fail to open.
This pull request added a test case to reproduce the problem and fixes it.


timothycrosley:
Thank you for finding the root cause and fixing this!

Issue #1685, opened 2021-03-06
`--skip` does not seem to work for specified subdirectories


jack1142:
Let's say you have a tree like this:
```
+ SomeFolder
+- pyproject.toml
+- directory
 +- subdirectory
  +- file.py
```
With contents of the files:
- `pyproject.toml`
```
[tool.isort]
skip = [
    "directory/subdirectory"
]
filter_files = true
```
- `directory/subdirectory/file.py`
```
import os
import collections
```

When using a command `isort directory\subdirectory\file.py`, the file `directory/subdirectory/file.py` gets fixed by isort.

If you change `directory/subdirectory` in `pyproject.toml` to be just `directory`, isort does not touch the `directory/subdirectory/file.py`, but I was unable to get isort to ignore `directory/subdirectory`.

As a workaround, you can use `skip_glob` with `directory/subdirectory/*` pattern, but it does seem like a bug to me, considering that this work fine for top-level directories.

gofr:
This is where that logic is implemented (and the next 10 lines):
https://github.com/PyCQA/isort/blob/f76ed04404ed7e814237564297d7b5e1f3358654/isort/settings.py#L524

Files are skipped using `skip` in two ways:
1. The normalized, absolutized path of a file matches the normalized, absolutized path of a `skip` item. E.g. you could specify `directory/subdirectory/file.py` to skip just that file.
2. A single component in the relative path of a file matches a `skip` item. This is why `directory` works, not because it's a top-level directory. `subdirectory` would also work, but this skips all directories with that name if you have more than one.

I guess there are two ways to go here:
* Improve the documentation to clarify that only the above two things work, and if you want something else you should use `skip_glob`, since that can already do what you need.
* Expand (1) above to also match paths to directories, not just files.

gofr:
One pull request for each of the two options I mentioned. The one that changes the implementation would need some tests at least, and some Windows fixes, but leaving them for now...

jack1142:
Seems like this has been fixed with documentation in #1689 so I'm gonna close this.

PR #1689, opened 2021-03-11
Clarify documentation for --skip option

gofr:
The --skip option only handles full file paths or names of
individual path components.

For example, "foo/bar/baz.py" can be used to skip just that file.
And "foo" would skip any files with that name and any files nested
in directories with that name.

On the other hand, "foo/bar" does *not* skip everything in the
"foo/bar" directory. --skip-glob can be used to achieve this.

Attempt to clarify the documentation to say that.

Potential fix for #1685. Alternative to #1690.


jack1142 (commented):
This needs to be updated in the `docs/configuration/options.md` as well

gofr:
> This needs to be updated in the `docs/configuration/options.md` as well

That file is actually generated (by https://github.com/PyCQA/isort/blob/develop/scripts/build_config_option_docs.py) :slightly_smiling_face:

PR #1695, opened 2021-03-22
Fix bug regarding multiline docstrings

jonafato:
If a multiline docstring has its closing quotes at the end of a line
instead of on a separate line, isort fails to properly add imports using
the `add_imports` config option; instead, it adds the desired imports
into the middle of the docstring as illustrated below. While PEP 257
(and other guides) advises that closing quotes appear on their own line,
`isort` should not fail here.

This change adds a check for closing docstrings at the end of a line in
addition to the existing line start check for all comment indicators. A
new section of the `test_add_imports` test explicitly tests multiline
imports and this failure scenario specifically.

---

A working example:

```python
"""My module.

Provides example functionality.
"""

print("hello, world")
```

Running `isort --add-import "from __future__ import annotations"`
produces the following as expected:

```python
"""My module.

Provides example functionality.
"""

print("hello, world")
```

---

The failure behavior described:

```python
"""My module.

Provides example functionality."""

print("hello, world")
```

Running `isort --add-import "from __future__ import annotations"` as
above produces the following result:

```python
"""My module.

Provides example functionality."""

print("hello, world")
```

Subsequent executions add more import lines into the docstring. This
behavior occurs even if the file already has the desired imports.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -13,7 +13,8 @@ from .settings import FILE_SKIP_COMMENTS
 
 CIMPORT_IDENTIFIERS = ("cimport ", "cimport*", "from.cimport")
 IMPORT_START_IDENTIFIERS = ("from ", "from.import", "import ", "import*") + CIMPORT_IDENTIFIERS
-COMMENT_INDICATORS = ('"""', "'''", "'", '"', "#")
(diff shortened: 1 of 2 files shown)


timothycrosley:
Thanks for finding and fixing this issue!

PR #1699, opened 2021-03-31
Require setuptools, isort/settings.py imports pkg_resources

hroncok:


timothycrosley:
Hi @hroncok, Thanks for raising this! You are correct it should be listed as a dependency, but an optional one. It's only imported when external plugins are utilized. I've addressed this here: https://github.com/PyCQA/isort/commit/3896dae9df9fb084634763f62485d7cd5808c2b2

Thanks!

~Timothy

Issue #1628, opened 2020-12-30
Faster gitignore support


timothycrosley:
Currently, while isort offers gitignore support, it performs too slowly to be used on anything but smaller projects. The current solution effectively calls a git sub process for every single file it encounters. This ticket represents the work to improve the performance to an acceptable point for large repos.

timothycrosley:
This has been fixed by @legau's PR :tada:

PR #1703, opened 2021-04-07
Faster gitignore

legau:
This PR addresses #1628 
The method I used is to keep in memory every gitignored files of checked folders. This reduces ``git -check-ignore`` calls which should be enough to improve significantly the ``--gitignore`` speed.
For now I haven't really tested it and there's room for optimization.

Let me know if you agree with the concept.

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -520,7 +520,7 @@ class Config(_Config):
             files = glob.glob(str(git_folder) + "/**/*", recursive=True)
             files_result = (
                 subprocess.check_output(  # nosec # skipcq: PYL-W1510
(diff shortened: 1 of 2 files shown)


timothycrosley:
I think this general approach is promising! If it could be updated to pass CI I would likely merge

legau:
So it seems Windows builds are still failing with respect to the subprocesses being handled differently. Maybe you have an idea what's wrong ? I don't have access to windows to debug this

legau:
@timothycrosley it should be ok now

timothycrosley:
@legau This is fantastic! Thanks for the great work and sticking with it, this improvement will make a lot of people happy

PR #1710, opened 2021-04-20
Fix UserWarnings

### LINKED ISSUE #1709: UserWarnings: "Likely recursive symlink detected" and "X was skipped as it's listed in ..."
This issue is about 2 `UserWarning`s. The first one seems to be a false-positive, the second one IMO should be a print instead of a warning.

1) [This line](https://github.com/PyCQA/isort/blob/99faae897f9da95cfd74ef9914c70313ed9abac3/isort/files.py#L30) in `isort/files.py` is possibly wrong as it seems to be triggered by any valid symlink, not only recursive ones. Now that I'm thinking about it, is there actually any reason to give such a warning even if this code did what it says it does and detected recursive symlinks? Why is a recursive symlink considered an exceptional condition if it's clearly supported by the very code that's giving the warning?

2) [This line](https://github.com/PyCQA/isort/blob/99faae897f9da95cfd74ef9914c70313ed9abac3/isort/main.py#L1182) in `isort/main.py` reports the skipped files as warnings (if the `--verbose` setting is enabled). I would argue that this should be a regular print. It's skipping a file, because it was specified to be skipped by the user. Therefore, it's an informational message, not a warning. It should go to `stdout` instead of `stderr` and it shouldn't be formatted as a warning (i.e. the user shouldn't care which line in which source file of `isort` this message was triggered).
   
   There is already a precedent for `print`ing informational messages if `--verbose` is specified. For example, [this line](https://github.com/PyCQA/isort/blob/f074239e5e39c2dd9e4d373e3c6ce528dae3f78c/isort/parse.py#L437) and [this line](https://github.com/PyCQA/isort/blob/f074239e5e39c2dd9e4d373e3c6ce528dae3f78c/isort/deprecated/finders.py#L393) (the second one is arguably a better candidate to be a warning actually)

I would be willing to create a PR that removes the first warning and/or changes the second warning to a print, if you agree with the above changes.

RuRo:
See #1709 for more info.

Fixes #1709.

### DIFF
--- a/isort/files.py
+++ b/isort/files.py
@@ -1,7 +1,6 @@
 import os
 from pathlib import Path
 from typing import Iterable, Iterator, List, Set
-from warnings import warn
 
 from isort.settings import Config
 
@@ -26,8 +25,6 @@ def find(
                         dirnames.remove(dirname)
                     else:
                         if resolved_path in visited_dirs:  # pragma: no cover
-                            if not config.quiet:
-                                warn(f"Likely recursive symlink detected to {resolved_path}")
                             dirnames.remove(dirname)
                     visited_dirs.add(resolved_path)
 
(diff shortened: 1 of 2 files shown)


timothycrosley:
Thanks for identifying and fixing these! :rocket:

PR #1725, opened 2021-05-09
Initialize `colorama` with `strip=False` to force color for non-tty output.

### LINKED ISSUE #1602: --color output does not seem to work in CI
While the color output works on a normal console, it fails in CI. In the attached screenshot I would expect the highlighted lines to be colored considering that isort is run with the `--color` flag.

My guess is that it does not detect the output as a "real interactive shell" and assumes that it does not support escape codes. But in this case it does. It would be nice (and sufficient) to have a `--force-color` flag to disable this auto-detection.

image

RuRo:
Fixes #1602.

### DIFF
--- a/isort/format.py
+++ b/isort/format.py
@@ -11,7 +11,7 @@ except ImportError:
     colorama_unavailable = True
 else:
     colorama_unavailable = False
-    colorama.init()
+    colorama.init(strip=False)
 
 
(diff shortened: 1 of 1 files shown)


RuRo:
👍 @timothycrosley do you have an ETA for the next release of `isort`?

RuRo:
@timothycrosley sorry to bother you, but can you please create a new PyPI release from the current master.

PR #1738, opened 2021-05-29
Add check-before-apply option to improve performance.

shotat:
## What is this change?

Add the `--check-before-apply` option.
If this option is enabled, `isort` skips actual sorting when the file is already formatted.
It improves performance in a typical use case.

## Why is this needed?

I've noticed that it took seconds or minutes to run `isort` for my Python project.
In a typical use case, `isort` is executed so many times to keep projects clean.
So usually most of the files are already formatted and no need to format repeatedly.

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -78,6 +78,7 @@ def sort_imports(
     file_name: str,
     config: Config,
     check: bool = False,
+    check_before_apply: bool = False,
     ask_to_apply: bool = False,
     write_to_stdout: bool = False,
     **kwargs: Any,
@@ -85,12 +86,15 @@ def sort_imports(
     incorrectly_sorted: bool = False
     skipped: bool = False
     try:
-        if check:
+        if check or check_before_apply:
             try:
                 incorrectly_sorted = not api.check_file(file_name, config=config, **kwargs)
             except FileSkipped:
                 skipped = True
-            return SortAttempt(incorrectly_sorted, skipped, True)
+            if check:
+                return SortAttempt(incorrectly_sorted, skipped, True)
+            if check_before_apply and not incorrectly_sorted:
+                return SortAttempt(incorrectly_sorted, skipped, True)
 
         try:
             incorrectly_sorted = not api.sort_file(
@@ -296,6 +300,14 @@ def _build_arg_parser() -> argparse.ArgumentParser:
         dest="ask_to_apply",
         action="store_true",
         help="Tells isort to apply changes interactively.",
+    )
+
+    general_group.add_argument(
+        "--check-before-apply",
+        dest="check_before_apply",
+        action="store_true",
+        help="Run checks before applying changes. "
+        "This can improve performance if most of the files are correctly formatted.",
     )
 
     target_group.add_argument(
@@ -1047,6 +1059,7 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
     ask_to_apply = config_dict.pop("ask_to_apply", False)
     jobs = config_dict.pop("jobs", None)
     check = config_dict.pop("check", False)
+    check_before_apply = config_dict.pop("check_before_apply", False)
     show_diff = config_dict.pop("show_diff", False)
     write_to_stdout = config_dict.pop("write_to_stdout", False)
     deprecated_flags = config_dict.pop("deprecated_flags", False)
@@ -1135,6 +1148,7 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
                     config=config,
                     check=check,
                     ask_to_apply=ask_to_apply,
+                    check_before_apply=check_before_apply,
                     write_to_stdout=write_to_stdout,
                     extension=ext_format,
                 ),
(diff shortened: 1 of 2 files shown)


anirudnits:
@shotat thanks for the PR, two questions on this:

1. Isn't this same as somebody running `isort . --check` and if it gets something before running isort directly? I ask this because I personally never felt any real performance issue with isort, I can't speak for all but we don't have a lot of people raise this issue either and believe this would happen in rare cases and so does this flag need to be added for them.
2. Is there any way to generate some metrics for this enhancement? I would love to see some numbers for small/medium/large projects and compare. Because this would slow down the execution for instances for cases where a lot of files need to be sorted.

shotat:
@anirudnits thank you for your quick response.

> Isn't this same as somebody running isort . --check and check if it gets something before running isort directly? I ask this because I personally never felt any real performance issue with isort, I can't speak for all but we don't have a lot of people raise this issue either and believe this would happen in rare cases and so does this flag need to be added for those cases.

My project contains 1000+ Python files and isort takes a few tens of seconds.
Actually, I wrote the following script to implement this feature for my project. As you mentioned, we can extract files from the output of `isort --check-only`.
So instead of adding `--check-before-apply` flag, it might be helpful if isort had another flag like `--file-only` or [flake8 -q](https://flake8.pycqa.org/en/latest/manpage.html)
 (which reports only file names) because extracting stderr message is a little bit awkward.
```sh
#!/bin/bash

targets=("$@")
if [ ${#targets[@]} -eq 0 ]; then
    exit 0
fi

files=$(poetry run isort --check-only "${targets[@]}" 2>&1 | awk '{ print $2 }' || true)
if [ -z "$files" ]; then
    exit 0
fi

poetry run isort $files
```

> Is there any way to generate some metrics for this enhancement? I would love to see some numbers for small/medium/large projects and compare. Because this would slow down the execution for instances for cases where a lot of files need to be sorted.

Here is a quick benchmark script, which generates 10 to 100000 files and run `isort` twice.
First execution with `--check-before-apply` has some performance penalties because of additional steps to format a lot of files.
Second execution, however, only took half the time. So this feature can bring a great improvement in a certain situation.


**scripts**

```sh
#!/bin/bash

file_counts=(10 1000 100000)

for num in "${file_counts[@]}"; do

    echo "# file count is ${num}"
    echo

    rm -rf isortbench
    mkdir isortbench
    content='import b\nimport a\n'
    for i in $(seq 1 $num); do
        echo "$content" > "isortbench/${i}.py"
    done

    # first isort
    echo 'running the first isort: files are not formatted.'
    time isort isortbench >/dev/null 2>&1
    echo
    # second isort
    echo 'running the second isort: files are already formatted.'
    time isort isortbench >/dev/null 2>&1

    rm -rf isortbench
    mkdir isortbench
    content='import b\nimport a\n'
    for i in $(seq 1 $num); do
        echo "$content" > "isortbench/${i}.py"
    done

    # first isort
    echo 'running the first isort --check-before-apply: files are not formatted.'
    time isort isortbench --check-before-apply >/dev/null 2>&1
    echo

    # second isort
    echo 'running the second isort --check-before-apply: files are already formatted.'
    time isort isortbench --check-before-apply >/dev/null 2>&1
done

```


**results**

```
# file count is 10

running the first isort: files are not formatted.

real    0m0.983s
user    0m0.731s
sys     0m0.180s

running the second isort: files are already formatted.

real    0m0.841s
user    0m0.679s
sys     0m0.153s

running the first isort --check-before-apply: files are not formatted.

real    0m0.820s
user    0m0.669s
sys     0m0.147s

running the second isort --check-before-apply: files are already formatted.

real    0m0.777s
user    0m0.636s
sys     0m0.137s

# file count is 1000

running the first isort: files are not formatted.

real    0m1.993s
user    0m1.388s
sys     0m0.599s

running the second isort: files are already formatted.

real    0m1.944s
user    0m1.342s
sys     0m0.504s

running the first isort --check-before-apply: files are not formatted.

real    0m2.229s
user    0m1.593s
sys     0m0.627s

running the second isort --check-before-apply: files are already formatted.

real    0m1.552s
user    0m1.232s
sys     0m0.300s

# file count is 100000

running the first isort: files are not formatted.

real    2m22.790s
user    1m17.039s
sys     0m56.474s

running the second isort: files are already formatted.

real    2m7.400s
user    1m16.900s
sys     0m44.322s
running the first isort --check-before-apply: files are not formatted.

real    2m50.169s
user    1m41.887s
sys     0m59.007s

running the second isort --check-before-apply: files are already formatted.

real    0m58.692s
user    0m46.005s
sys     0m12.446s
```

timothycrosley:
Intuitively, it doesn't make a ton of sense that such an option could possibly have any impact but a negative one, since check does all the same operations as normal running, but then also tries to compare the output. Literally the only action that doesn't take place is a rename, to swap in the new sorted version of the file. If it helps it's likely an indicative of a disk performance issue or really small files. I think the better approach to improve the speed (that would actually have a very measurable impact) would be to cache success operations - similar to black - so that on the second run nothing at all happens if the file and the config haven't changed.

shotat:
@timothycrosley thanks for your advice.
As you mentioned, too many disk IOs can degrade performance.
I noticed that isort has already `--overwrite-in-place` option.
It seems that this might be helpful to reduce redundant disk IOs.
However, `--overwrite-in-place` option actually does not reduce any disk IOs because isort always create `<source_file>.isorted` and write out the result to the file.
This looks like an unnecessary process so I would like to write another patch not create `.isorted` file if `--overwrite-in-place` option is enabled.

I'd like to close this pull request anyway.
Thank you!

PR #1756, opened 2021-06-19
1732 adds sort_order allowing pythonic sort as an option

bobwalker99:
recreated branch from https://github.com/PyCQA/isort/pull/1752


timothycrosley (approved):
:tada: Great work here! Thanks so much for your contribution!

PR #1759, opened 2021-06-21
Use a pipe to pass filenames to git check-ignore

### LINKED ISSUE #1758: New code to check ignored files trivially leads to command-line overload
5.9.0 made changes to the way gitignore is handled (see 927cf7efed0c526e8aee9ce5e1ff7a7682ec7d57 and 4f32c446f9de8bb10a402a58420a6b6b32e57f08), but unfortunately doesn't take into account that a typical project could easily have many thousands of files. The following line then blows up:
https://github.com/PyCQA/isort/blob/0ccefec5a394a77a8a03e6c2f517b484765abcda/isort/settings.py#L546
with:
```python
  File "/.../bin/isort", line 8, in <module>
    sys.exit(main())
  File "/.../site-packages/isort/main.py", line 1130, in main
    if config.is_skipped(Path(file_name)):
  File "/.../site-packages/isort/settings.py", line 586, in is_skipped
    git_folder = self._check_folder_gitignore(str(file_path.parent))
  File "/.../site-packages/isort/settings.py", line 550, in _check_folder_gitignore
    subprocess.check_output(  # nosec # skipcq: PYL-W1510
  File "/.../subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/.../subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/.../subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/.../subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: 'git'
```

In our project, git tracks 214 files. But, running `glob.glob()` on the same directory gives me:
```
$ python3 -c 'import glob; print(sum(1 for _ in glob.glob("./**/*", recursive=True)))'
8677
```

8677 is way too many arguments to pass to a single command. Our project happens to include a Node.js component and so has a `node_modules` subfolder (ignored), accounting for 7.3k files in that total.

The code should either be reverted, or updated to handle a large number of files in batches, or use the `--stdin` to pipe through the filenames over stdin.

E.g. the following code works for our project without breaking command-line length:

```python
            files_result = subprocess.check_output(
                ["git", "-C", str(git_folder), "check-ignore", "--stdin"],
                encoding="utf-8",
                env={"LANG": "C.UTF-8"},
                input="\n".join(files),
            ).splitlines()
```

Note: in addition to switching to `--stdin`, I also added `LANG=C.UTF-8` to the environment (should make the `codecs.escape_decode()` call obsolete on Windows), and used `.splitlines()` to remove the need to handle an empty last line result.

mjpieters:
The number of files in a project could be larger than the limit on command-line arguments. Using a pipe avoids running into this issue.

Fixes #1758

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -2,6 +2,7 @@
 
 Defines how the default settings for isort should be loaded
 """
+import codecs
 import configparser
 import fnmatch
 import glob
@@ -543,12 +544,16 @@ class Config(_Config):
             git_folder = Path(topfolder_result.decode("utf-8").split("\n")[0])
 
             files = glob.glob(str(git_folder) + "/**/*", recursive=True)
-            files_result = subprocess.check_output(  # nosec # skipcq: PYL-W1510
-                ["git", "-C", str(git_folder), "check-ignore", "--stdin"],
-                encoding="utf-8",
-                env={"LANG": "C.UTF-8"},
-                input="\n".join(files),
-            ).splitlines()
+            files_result = (
+                codecs.escape_decode(  # type: ignore
+                    subprocess.check_output(  # nosec # skipcq: PYL-W1510
+                        ["git", "-C", str(git_folder), "check-ignore", "--stdin"],
+                        input="\n".join(files).encode(),
+                    )
+                )[0]
+                .decode("utf-8")
+                .splitlines()
+            )
 
             self.git_ignore[git_folder] = {Path(f.strip('"')) for f in files_result}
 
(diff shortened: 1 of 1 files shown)


timothycrosley:
Thank you so much for identifying and fixing this issue!

PR #1761, opened 2021-06-21
Avoid git octal escaping

### LINKED ISSUE #1760: Use null-delimiting when interacting with git check-ignore and disable git octal quoting
Currently, the `git check-ignore` command result is parsed with using an unsupported legacy function in `codecs`, because git can output `\OOO` octal escape codes for any 'unusual' characters.

Instead of using undocumented functions, isort should simply disable the git quoting features:

- Disable [`core.quotePath`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corequotePath) (pass `-c core.quotePath=`).
- Use null delimiting instead of linefeed delimiting (pass [`-z`](https://git-scm.com/docs/git-check-ignore#Documentation/git-check-ignore.txt--z), read and write filenames with `\0` as the delimiter)

This would also allow isort to just use the `subprocess` encoding and decoding support, provided git is explicitly configured to expect and produce UTF-8 (set the `LANG=C.UTF-8` environment variable).

mjpieters:
- Disable core.quotePath
- Use -z to switch from linefeed to null-byte delimiting

This fixes #1760

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -2,7 +2,6 @@
 
 Defines how the default settings for isort should be loaded
 """
-import codecs
 import configparser
 import fnmatch
 import glob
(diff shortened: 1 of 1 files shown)


mjpieters:
Right, so `scripts/done.sh` doesn't run all the tests. Now running `poetry run pytest tests/unit/ -s` locally and will see if I can't get these tests to pass properly, later today.

mjpieters:
> Right, so `scripts/done.sh` doesn't run all the tests. Now running `poetry run pytest tests/unit/ -s` locally and will see if I can't get these tests to pass properly, later today.

Turns out I was just being a numpty. Force-pushed an amended commit, should be better.

timothycrosley:
:tada: thanks for this nice improvement!

Issue #1753, opened 2021-06-18
Documentation not working correctly


Lars147:
Hi there,

I noticed that some links on the isort documentation pages are broken. For now, I noticed it only for the page https://pycqa.github.io/isort/docs/major_releases/introducing_isort_5.html.

The listed links on this page are redirecting to pages ending with a trailing "/". 

However, the links should end with ".html" in order to work.

anirudnits:
@Lars147 thanks for pointing this out, I've corrected the links in the docs. Hope I got them all!

PR #1764, opened 2021-06-24
Fix URLs for 5.0.0 upgrade guide

JohnHBrock:
Related to #1753, I fixed some more broken URLs.


timothycrosley (approved):
:tada: Thanks for fixing these!

PR #1773, opened 2021-07-05
Symlinked paths are excluded from gitignore checks

ucodery:


ucodery:
This fixes #1773

ucodery:
Our company uses a build tool in CI that link a bin dir from the root of the repo to a global private bin dir on the host machine. Before this change, runs of isort had ballooned to a run time of at least 7 minutes. After this change at least 1000 less files were passed to git to check for ignoring and runtime went down to almost 10 seconds.
Also, FYI we have multiple .gitignores at different levels, so a self-contained .gitignore like in #1585 would not have helped.

timothycrosley (approved):

Issue #1772, opened 2021-07-05
skip-gitignore will check files not in the git repository


ucodery:
When isort asks git to check which files it ignores, isort can ask git about files not under revision tracking, due to symlinks. Even though isort gets the root of the git repository from a prior git execution, it then does a recursive glob which can resolve symlinked directories and add files and sub-directories that are not part of the repository.

When this happens, the call to find out what files should be ignored gets very slow, slower even than just running isort over a larger code base. git will spew out `fatal: pathspec '/some/external/path' is beyond a symbolic link` for all of the extra paths.

ucodery:
Also, some food for thought, black will ignore any file that is on the other side of a symlink that points outside of the project's root. It might be good for isort to take the same line and always ignore these sorts of files.
isort does have an older userbase, and works in more than just git repositories, so this might break some workflows though.

timothycrosley:
Closing this as your fix has been merged in the main branch, and will be released soon (in 5.9.2) :tada: - thank you! 

>  It might be good for isort to take the same line and always ignore these sorts of files.

I'm open to doing this in the future!

adamjstewart:
I just ran into this same issue! Unfortunately, isort 5.9.2 doesn't seem to solve the problem for me.

Specifically, I have a [Spack](https://spack.io) environment in my project directory. The `.gitignore` contains `.spack-env/`, and git correctly ignores everything in this directory. However, when I run `isort`, it tries to reformat files in a completely different root directory because it follows symlinks in `.spack-env/`.

I thought #1773 was designed specifically to address this issue? Is my case somehow slightly different?

timothycrosley:
@adamjstewart Another option would be to use the  [dont_follow_links](https://pycqa.github.io/isort/docs/configuration/options.html#dont-follow-links) option.

adamjstewart:
Yes, I have `skip_gitnore = true` in my `pyproject.toml`. I also tried specifying `--skip-gitignore` on the command-line.

ucodery:
It's possible that this fix is incomplete. If you wouldn't mind could you send the full config you are using, as well as the complete output from your run of `isort`? (Also `isort --show-files` would be helpful).

adamjstewart:
Here is a minimal working reproducer. First, we create some files:
```console
$ cd
$ cat > a.py
# Incorrectly ordered imports
import numpy
import os
$ mkdir foo
$ cd foo
$ cp ../a.py b.py
$ ln -s ../a.py c.py
[...]

PR #1789, opened 2021-07-20
Check gitignore more intelligently

ucodery:
This is building on the original fix for #1772


ucodery:
Hey thanks for merging this. I was not able to repro any failures running integration tests locally and couldn't figure out what was wrong on the PR checks. Hopefully this doesn't regress anything.

PR #1806, opened 2021-08-16
Don't use multiprocessing with --jobs 1

scop:
A pool with size 1 is quite a bit slower than not using multiprocessing
here at all.

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -1150,7 +1150,7 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
         if config.verbose:
             print(ASCII_ART)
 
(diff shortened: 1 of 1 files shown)


timothycrosley:
Why not just avoid specifying `--jobs`? I can see the diagnostic value in having a single multiprocess job, and since the default is what you suggest, I can't immediately see the value in trying to optimize it

scop:
I just found it weird. Could also be a special wrinkle when scripting something like "run isort with $number_of_cores minus one" so that it doesn't hog all cores, but that'd cause a slowdown with a 2-core box unless one knows 1 is a magic value to avoid.

But not a biggie if you'd rather not change this.

Issue #1807, opened 2021-09-10
ERROR: Unrecoverable exception thrown when parsing tests/fixtures/__init__.py!This should NEVER happen.


A143865:
Traceback (most recent call last):
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/bin/isort", line 8, in <module>
    sys.exit(main())
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/main.py", line 1185, in main
    for sort_attempt in attempt_iterator:
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/main.py", line 1170, in <genexpr>
    sort_imports(  # type: ignore
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/main.py", line 92, in sort_imports
    incorrectly_sorted = not api.sort_file(
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/api.py", line 407, in sort_file
    changed = sort_stream(
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/api.py", line 210, in sort_stream
    changed = core.process(
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/core.py", line 384, in process
    parsed_content = parse.file_contents(import_section, config=config)
  File "/Users/a143865/.cache/pre-commit/reposrxo0one/py_env-python3/lib/python3.9/site-packages/isort/parse.py", line 450, in file_contents
    root = imports[placed_module][type_of_import]  # type: ignore
KeyError: 'LOCALFOLDER'

anirudnits:
Hello @A143865, can you share the isort configurations you are using? You can fetch them by running `isort . --show-config`.

A143865:
> Hello @A143865, can you share the isort configurations you are using? You can fetch them by running `isort . --show-config`.

Certainly. I should also note that this error did resolve though it did so without any effort from me to try and debug. 
[isort_configs.txt](https://github.com/PyCQA/isort/files/7154799/isort_configs.txt)

anirudnits:
[...]

PR #1810, opened 2021-09-14
More extensive exception handling

anirudnits:
Based on issue #1807 where the error message is directly printed on the console, rather than raising the `MissingSections` exception with the more descriptive error message.


timothycrosley (approved):

timothycrosley:
This is great! Thank you for fixing this!

PR #1818, opened 2021-09-30
feature: implement lines before imports

### LINKED ISSUE #1646: Feature Request: Lines Before Imports
There is already a `lines_after_imports` option.  Would it be possible to also have a `lines_before_imports` option?  I have a project where half of the files have a space after the module doc string and before the first import and the other half do not have a blank line.  Ideally isort could be used to make these consistent the same way it can to make the spacing after the imports consistent.

```
"""Module docstring."""

import os
```

```
"""Module docstring."""
import os
```

sgaist:
This feature complements the `lines_after_imports` option by allowing to specify a number of space before the imports.

This can be useful for example to ensure that the same amount of space can be found between a module docstring and the imports.

Closes #1646


MNastri:
Hello. This does not check if the specified number of lines before imports already exist. It keeps adding lines every time isort is ran and the file imports get fixed.

sgaist:
Hi,

I added the following to `test_custom_lines_before_import_section`  in `test_isort.py`

```python
    output = isort.code(test_input, lines_before_imports=2)
    assert isort.code(output, lines_before_imports=5) == 5 * ln + test_input
```

The original input is thus updated to have 2 lines before and then it is run again to have 5 lines which yield the expected result.

Can you share a detailed description of how you have your issue triggered on your side ?

PR #1833, opened 2021-10-27
only skip real comments

chrisdecker1201:
I had the issue that I wrote a test for darker (https://github.com/akaihola/darker/pull/230) where I get an error because isort is handling every # isort: skip_file even if this is not a real comment.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -1,6 +1,7 @@
 import textwrap
 from io import StringIO
 from itertools import chain
+from re import match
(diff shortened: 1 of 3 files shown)


timothycrosley:
I think the correct thing todo here, is just to be clever about how you define this token in other places, like isort itself is, see: https://github.com/PyCQA/isort/blob/dbd18429cf566e9554842c9322515237bd8efb49/isort/settings.py#L58

Issue #1840, opened 2021-11-03
skip_file ignored when on the first docstring line


romaingz:
This does not occur on version `5.9.3`.

Since version `5.10.0`, the following `skip_file` directive is ignored:

```python
"""isort:skip_file"""
```

This is also ignored:

```python
"""isort:skip_file

some other comment.
"""
```

It needs to be on a new line:

```python
"""
isort:skip_file
"""
```

Omegastick:
`skip_file` in a comment is also ignored.

```
#  isort:skip_file
````

**Comment by ahopkins** [comment]
FWIW, this is also ignored:

```
"""
   isort:skip_file
"""
```

max-sixty:
Here's an example of 5.10.0 ignoring directives (unless there's some intentional change I'm missing?): https://github.com/pydata/xarray/pull/5949

timothycrosley:
Sorry for the noise!  This should be fixed in https://github.com/PyCQA/isort/releases/tag/5.10.1, a contribution meant to avoid false flagging of non-comment isort: skip-files, was to narrow and has been reverted. In the meantime additional test cases will be added to catch this going forward.

Thanks!

~Timothy

PR #1845, opened 2021-11-07
Revert "only skip real comments"

timothycrosley:
Reverts PyCQA/isort#1833


timothycrosley:
Reverting this as it breaks many existing skips. See: https://github.com/PyCQA/isort/issues/1840

Resolves #1840

PR #1858, opened 2021-11-25
Remove redundant `multi_line_output = 3` from "Compatibility with black"

jdufresne:
According to the profiles documentation:
https://pycqa.github.io/isort/docs/configuration/profiles.html

The black profiles contains: "multi_line_output = 3". Avoid confusion on
the "Compatibility with black" page by not adding a redundant option.
The reader may believe the extra "multi_line_output" is required.

This came up in a real world scenario:
https://github.com/pypa/pep517/pull/137#discussion_r756648644


timothycrosley:
Thanks for fixing this! I could see how that could be confusing

PR #1860, opened 2021-12-01
Fix `FileNotFoundError` raised by `_check_folder_gitignore`

tyong920:
Current version raised `FileNotFoundError ` when `git` is not installed.

E.g.
CI using docker image: python:slim which has no git installed by default.
Screen Shot 2021-12-01 at 10 57 57

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -555,7 +555,10 @@ class Config(_Config):
         env = {**os.environ, "LANG": "C.UTF-8"}
         try:
             topfolder_result = subprocess.check_output(  # nosec # skipcq: PYL-W1510
-                ["git", "-C", folder, "rev-parse", "--show-toplevel"], encoding="utf-8", env=env
+                ["git", "-C", folder, "rev-parse", "--show-toplevel"],
+                encoding="utf-8",
+                env=env,
+                shell=True,
             )
         except subprocess.CalledProcessError:
             return None


timothycrosley:
Hi @tyong920,

Thank you for taking the time to send this improvement to the project! I'm a little uncertain how this changes the behavior from just the screen shot above. What did it do before? What does it do with this change in place?

tyong920:
> Hi @tyong920,
> 
> Thank you for taking the time to send this improvement to the project! I'm a little uncertain how this changes the behavior from just the screen shot above. What did it do before? What does it do with this change in place?

Hi @timothycrosley,

Appreciate for the detailed reply!

From the current main branch implementation, `git` is a runtime dependency for the function `isort/settings.py::_check_folder_gitignore`.
i.e. in a Environment which has no `git` installed, it raise FileNotFoundError which make `isort` result in a none-zero exit code.
But in my perspective, it might be avoid since the function seems designed to tolerant with None as a return value when executing `git` related commands failed.
By providing `shell=True` it can make sure to raise `subprocess.CalledProcessError` thus can be catch by the later code block. 

Some discussions on this topic, hope it helps:
https://github.com/anapaulagomes/pytest-picked/issues/16
https://github.com/madjar/nox/issues/36

tyong920:
😂
After reviewing more implementation details, I think it might be a better choice to close the PR and keep the code as is.

Also, declare `git` as a running dependency explicitly.

PR #1863, opened 2021-12-04
Add a command-line flag to sort __all__ re-exports (#1862)

### LINKED ISSUE #1862: A command-line flag to sort __all__ re-exports
Generally speaking, imports in typed projects have a list (or tuple) of \_\_all\_\_ exports that should be made. #1358 introduces a way to do this using comments, but this can look quite messy and cluttered. For such a common use case, I would suggest a command-line flag that can automatically find and sort these \_\_all\_\_ exports.

Ideally this would look something like this:
```sh
isort . --sort-reexports
```
And in a config file (specifically pyproject.toml in this case):
```toml
[tool.isort]
sort_reexports = true
```

The name for the flag is taken from Mypy's "implicit re-exporting is disabled" message, but if there's anything more suitable obviously that should be used instead.

This should also support the notion described in #1815 when that's done.

parafoxia:
This PR adds a `--sort-reexports` CLI flag and the respective cfg and toml options to isort, which allows for similar behaviour to `# isort: [list|tuple|dict]` without the need to have comments scattered around. See #1862 for more information.

This PR makes no attempt to document this feature at present, but I'm happy to do this if you'd like me to.

Closes #1862.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -24,6 +24,11 @@ CODE_SORT_COMMENTS = (
     "# isort: unique-tuple",
     "# isort: assignments",
 )
+LITERAL_TYPE_MAPPING = {
+    "(": "tuple",
+    "[": "list",
+    "{": "dict",
+}
 
 
 def process(
@@ -73,6 +78,8 @@ def process(
     end_of_file: bool = False
     verbose_output: List[str] = []
     lines_before: List[str] = []
+    auto_reexporting: bool = False
+    line_index: int = 0
 
     if config.float_to_top:
         new_input = ""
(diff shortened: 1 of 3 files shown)


parafoxia:
Also, for some reason, the `test_importable` test fails when localling running pytest. I'm not really sure why this is, but I haven't attempted a fix as I believe it's outside the scope of this PR, and the failure is likely not caused by it.

timothycrosley:
Thanks for this improvement! The tests pass in CI and locally for me, so I'm sure its just because of the state of local venv, maybe not using poetry or not using ./scripts/done.sh?

parafoxia:
I was using Poetry, but didn't realise that other script was there, so probably was something to do with that. I did initially try to install everything without Poetry cos I can't for the life of me get it to work on 3.10, but pip wouldn't install everything, so not sure what's going on there.

And no worries! Lib has been super useful to me for a while so thought I'd give a little back ^-^

PR #1876, opened 2021-12-29
Add magic trailing comma option

### LINKED ISSUE #1683: Support for magic trailing comma
Black auto-explodes imports, collections, function calls, etc. when they contain a trailing comma:
https://github.com/psf/black/blob/master/docs/the_black_code_style.md#the-magic-trailing-comma

This means that even though this would fit into one line:
```py
from seven_dwwarfs import (
    Grumpy,
    Happy,
    Sleepy,
    Bashful,
    Sneezy,
    Dopey,
    Doc,
)
```
Black will NOT convert it into this:
```
from seven_dwwarfs import Grumpy, Happy, Sleepy, Bashful, Sneezy, Dopey, Doc
```
As long as the import list contains a trailing comma.

`isort` currently does not have the magic trailing comma handling as far as I can tell and will collapse the import list if it can fit within the line length limit.

legau:
Resolves #1683, it adds the `split_on_trailing_comma` option. 
There's an ambiguous case when from imports are split into 2 parts like this :
```python
from lib import (a, b, c,)
from lib import d
```
In this case I made the choice to always make it split whenever one of the from imports has a trailing comma.

The integration test on airflow won't work until https://github.com/apache/airflow/pull/20556 is passed

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -889,6 +889,17 @@ Includes a trailing comma on multi line imports that include parentheses.
 
 - --tc
 - --trailing-comma
+## Split on Trailing Comma
+
+Split imports list followed by a trailing comma into VERTICAL_HANGING_INDENT mode. This follows Black style magic comma.
+
+**Type:** Bool  
+**Default:** `False`  
+**Config default:** `false`  
+**Python & Config File Name:** split_on_trailing_comma
+**CLI Flags:**
+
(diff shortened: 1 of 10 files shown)


jack1142:
> The integration test on airflow won't work until https://github.com/apache/airflow/pull/20556 is passed

It looks like it's been merged now so perhaps you should try rerunning tests now?

PR #1896, opened 2022-02-12
Document shared profiles

matthewhughes934:
These were added with 2d76984f1345763f31aff8c1fc940000700cd46e, with
tests and an example added with
57eb70a4c70311f6b1748062853ded38f7fec366. Add some documentation on how
these can be created, including a link to the example profile in this
repo.

Add a link to these docs from where shared profiles are mentioned for
the `profile` config

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -1132,7 +1132,10 @@ Inserts a blank line before a comment following an import.
 
 ## Profile
 
-Base profile type to use for configuration. Profiles include: black, django, pycharm, google, open_stack, plone, attrs, hug, wemake, appnexus. As well as any shared profiles.
+Base profile type to use for configuration. Profiles include: black, django,
+pycharm, google, open\_stack, plone, attrs, hug, wemake, appnexus. As well as
+any [shared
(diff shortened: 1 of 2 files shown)


docs/configuration/options.md
```diff
@@ -1132,7 +1132,10 @@ Inserts a blank line before a comment following an import.
 
 ## Profile
 
-Base profile type to use for configuration. Profiles include: black, django, pycharm, google, open_stack, plone, attrs, hug, wemake, appnexus. As well as any shared profiles.
+Base profile type to use for configuration. Profiles include: black, django,
+pycharm, google, open\_stack, plone, attrs, hug, wemake, appnexus. As well as
+any [shared
+profiles](https://pycqa.github.io/isort/docs/howto/shared_profiles.html).
```
matthewhughes934, line 1138: I'm not sure this will link where I want/expect them to link

docs/howto/shared_profiles.md
```diff
@@ -0,0 +1,18 @@
+# Shared Profiles
```
matthewhughes934, line 1: I added this under `docs/howto` because I'm not sure where else might be best.

Issue #1895, opened 2022-02-11
skip-gitignore walks all files, including ignored ones


bmalehorn:
Hi, I'm looking into a performance issue with isort.

It seems that when you pass in `--skip-ignore` and a directory, it will:

1. enumerate *every* file, including ignored files, in the repo
2. query git to see which of those files should be ignored
3. iterate through the folder
4. ignore files from step 2

https://github.com/PyCQA/isort/blob/c6a41965247a858a0afd848fbebfca18b8983917/isort/settings.py#L555-L586

In my use case, my subdirectory has 8 files. But before checking those, isort will first enumerate all 8 *million* untracked files in the entire repo, which takes several minutes! 

There are a couple potential fixes:
1. stop calling `os.walk`, and instead run `git ls-files`. This gets all *tracked* files (which there's a lot less of) instead of getting all *untracked* files and filtering
1. only call `os.walk` in the subdirectory you're running in, instead of the entire git repo

I'd be happy to prep a PR fixing this issue, but I want to see if that makes sense or if there's other considerations I'm not taking into account.

# Reproduction

1. modify isort to print each file it's walking in [`_check_folder_gitignore`](https://github.com/PyCQA/isort/blob/c6a41965247a858a0afd848fbebfca18b8983917/isort/settings.py#L572-L573):
```
            for git_file in git_files:
                print("walking: ", os.path.join(root, git_file))
                files.append(os.path.join(root, git_file))
```

2. run this script:
```bash
#!/bin/bash

set -xeuo pipefail

cd ~/tmp/
rm -rf isort-repro/
git init isort-repro
cd isort-repro
echo 'build/' > .gitignore
mkdir build/ src/
touch build/1.py build/2.py build/3.py # these should be ignored
touch src/main.py                      # this should be included
isort src/ --skip-gitignore
```
Actual output:
```
walking:  /Users/brian.malehorn/tmp/isort-repro/.gitignore
walking:  /Users/brian.malehorn/tmp/isort-repro/build/2.py
walking:  /Users/brian.malehorn/tmp/isort-repro/build/3.py
walking:  /Users/brian.malehorn/tmp/isort-repro/build/1.py
walking:  /Users/brian.malehorn/tmp/isort-repro/src/main.py
```

Expected output:
```
walking:  /Users/brian.malehorn/tmp/isort-repro/.gitignore
walking:  /Users/brian.malehorn/tmp/isort-repro/src/main.py
```
or maybe even this, since I only passed in `src`:
```
[...]

PR #1900, opened 2022-02-16
skip-gitignore: use allow list, not deny list

bmalehorn:
Improve the performance of `--skip-gitignore` by enumerating all **tracked** files instead of all **ignored** files.

Before:
- call `os.walk` for every target directory
- for each git repo it encounters:
  - call `os.walk` to walk over all tracked and ignored files
  - call `git check-ignore` on all those files to narrow it to only **ignored** files
- for each file it encounters, skip it if it's in the list of ignored files for that repo

After:
- call `os.walk` for every target directory
- for each git repo it encounters:
  - call `git ls-files` to get a list of all **tracked** files
- for each file it encounters, skip it if it's not in the list of tracked files for that repo

Closes https://github.com/PyCQA/isort/issues/1895.

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -238,7 +238,7 @@ class _Config:
     reverse_sort: bool = False
     star_first: bool = False
     import_dependencies = Dict[str, str]
-    git_ignore: Dict[Path, Set[Path]] = field(default_factory=dict)
+    git_ls_files: Dict[Path, Set[str]] = field(default_factory=dict)
     format_error: str = "{error}: {message}"
     format_success: str = "{success}: {message}"
     sort_order: str = "natural"
@@ -552,7 +552,7 @@ class Config(_Config):
         else:
             return bool(_SHEBANG_RE.match(line))
 
-    def _check_folder_gitignore(self, folder: str) -> Optional[Path]:
+    def _check_folder_git_ls_files(self, folder: str) -> Optional[Path]:
         env = {**os.environ, "LANG": "C.UTF-8"}
         try:
             topfolder_result = subprocess.check_output(  # nosec # skipcq: PYL-W1510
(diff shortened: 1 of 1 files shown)


He3lixxx:
Struggeling with the same problem you described in #1895 (~90s run time an hour ago for ~10k lines of code across 107 files), thus this post: This reduces runtime a bit, but doesn't fully solve it for me.

On my codebase, with `skip_gitignore` enabled in a config file, after cleaning out some old venv directories, 5.10.1 takes ~16s for `isort --show-files .`
Your branch here takes ~7.2s
However, doing `isort --show-files . > isort_files.txt` and then `cat isort_files.txt | time xargs isort` only takes 1.1s, which is what I had hoped to be isort's overall performance.
For comparison, `black` runs in 3.3 seconds, and has arguably more stuff to do.

Using `strace isort --show-files . 2>&1 | grep stat`, I still get a huge amount of file status queries, mostly on files that are in folders ignored in `.gitignore`. I noticed, however, that I no longer get queries on `node_modules`, which is in the defaults for `skip`.

It looks to me as if the following execution flow doesn't ignore a .gitignored directory:
* The generator returned from `files.find` in [`main.py` at L1173](https://github.com/bmalehorn/isort/blob/1e22ced0ed25a11b4c95733af48e9402e3c9ce2b/isort/main.py#L1173) will be iterated directly below in 1175.
* [`files.find`](https://github.com/bmalehorn/isort/blob/bm/skip-gitignore-ls-files/isort/files.py#L8) is called with `paths=["."]`, and does `for dirpath, dirnames, filenames in os.walk(path, topdown=True, followlinks=whatever)`. A directory [will be removed from `dirnames` to skip it if `config.is_skipped` evaluates to true](https://github.com/bmalehorn/isort/blob/bm/skip-gitignore-ls-files/isort/files.py#L25). However, this will never be the case for a gitignored directory, because the git logic in `is_skipped` [currently is](https://github.com/bmalehorn/isort/blob/bm/skip-gitignore-ls-files/isort/settings.py#L592):
```python
            # git_ls_files are good files you should parse. If you're not in the allow list, skip.
            if (
                git_folder
                and not file_path.is_dir()
                and str(file_path.resolve()) not in self.git_ls_files[git_folder]
            ):
                return True
        return False
```
so `file_path.is_dir()` is true, so this returns false, and `os.walk` will still descend into the directory.

bmalehorn:
Yup @He3lixxx you're right that this PR does not completely fix the problem.

I wrote this fix as a performance improvement that specifically targeted my use case:
a. tracked files in whole repo: ~30k
b. untracked files in whole repo: ~2m
c. tracked files in directory I'm running `isort` on: ~100
d. untracked files in directory I'm running `isort` on: ~1k

Before my PR, the runtime was O(a + b). With this PR it's O(a + c + d). That's a lot faster for me! But in an ideal world, it would only be O(c). And if you're running it on the whole repo like you are it's not much faster at all.

---

> It looks to me as if the following execution flow doesn't ignore a .gitignored directory:

You're right. The reason I wrote it that way is that git does not have an easy way to get a list of ignored / tracked directories, only tracked files.

A more correct way would be:
- save a sorted list of all tracked files
- if you're checking a directory, binary search the list of tracked files
- if the file right after this directory has the same prefix, descend:
```
   foo/c/bar.txt
-> foo/c/ would go here
   foo/d/buzz.txt
```

- Otherwise, skip it:
```
   foo/b/bar.txt
-> foo/c/ would go here
   foo/d/buzz.txt
```

The most correct solution would be to remove `os.walk` entirely in favor of running `git ls-files`, but that might conflict with other custom rules around isort skipping.

---

@He3lixxx does that make sense? I could update this PR to include that. But I kind of want to get the maintainers to take a look at it & be ok with the general idea before trying to optimize (& obfuscate) it further.

timothycrosley (approved):

PR #1919, opened 2022-04-13
Infinite loop for unmatched parenthesis

### LINKED ISSUE #1917: Infinite loop on unclosed parenthesis
If I run isort (5.10.1 or `main` at c6a41965247a858a0afd848fbebfca18b8983917) on this invalid file with an unclosed parenthesis, it goes into a 100% CPU infinite loop:

```python3
from os import (
```

anirudnits:
- [x] Resolves #1917
- [x] Add tests

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -7,7 +7,7 @@ import isort.literal
 from isort.settings import DEFAULT_CONFIG, Config
 
 from . import output, parse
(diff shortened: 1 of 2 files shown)


anirudnits:
The pipeline is failing for outdated dependencies, I'll work to update them. Otherwise the PR seems to be working right.

PR #1961, opened 2022-08-03
Fix Rich compatibility

### LINKED ISSUE #1960: Importing isort breaks Rich
````python
from rich.console import Console
from rich.markdown import Markdown

MARKDOWN = """\
```python
from pathlib import Path

import httpx


def download(path: Path, *args, **kwargs):
    with httpx.stream('GET', *args, **kwargs) as response:
        response.raise_for_status()

        with path.open(mode='wb', buffering=0) as f:
            for chunk in response.iter_bytes(16384):
                f.write(chunk)
```
"""


def main():
    # Comment this out
    import isort

    Console().print(Markdown(MARKDOWN))


if __name__ == '__main__':
    main()
````

ofek:
Resolves #1960

### DIFF
--- a/isort/format.py
+++ b/isort/format.py
@@ -11,7 +11,6 @@ except ImportError:
     colorama_unavailable = True
 else:
     colorama_unavailable = False
-    colorama.init(strip=False)
 
 
 ADDED_LINE_PATTERN = re.compile(r"\+[^+]")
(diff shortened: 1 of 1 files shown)


ofek:
cc @timothycrosley if you have time 🙂

timothycrosley (approved):

Issue #676, opened 2018-02-15
check-only false positive if file is missing closing newline


mattbennett:
`isort --check-only` will complain about import whitespace if the _file_ is missing a closing newline.

``` python
def test_strict_whitespace_no_newline(capsys):
    test_input = ('import os\n'
                  '\n'
                  'from django.conf import settings\n'
                  '\n'
                  'print(1)')
    SortImports(file_contents=test_input, check=True)
    out, err = capsys.readouterr()
    assert out == ''  # valid imports, but AssertionError
```

timothycrosley:
This is fixed in develop, thanks for reporting!

~Timothy

PR #1966, opened 2022-08-21
Fix #1855: fix check if file altered

### LINKED ISSUE #1855: `lines_before_import` seems to not always work on files
Here is the minimal example is created:
```python
import isort

str_in = """# comment
from foo import bar


def main():
    ...
"""

str_out = isort.code(
    str_in,
    lines_before_imports=1
)

with open("tmp2.py", "w") as fd:
    fd.write(str_in)
isort.file(
    "tmp2.py",
    lines_before_imports=1
)

print("In:")
print(str_in)
print("-" * 79)
print("Out String:")
print(str_out)
print("-" * 79)
print("Out file:")
with open("tmp2.py", "r") as fd:
    print(fd.read())
```
Which yields
```
In:
# comment
from foo import bar


def main():
    ...

-------------------------------------------------------------------------------
Out String:
# comment

from foo import bar


def main():
    ...

-------------------------------------------------------------------------------
Out file:
# comment
from foo import bar


def main():
    ...

```
So interestingly enough, it seems like `lines_before_import` does not affect files in this case. Is there a config option that I did not find?
Another observation I made, If I change the input sting in a way that `isort` wants to change it for other reasons than `lines_before_import`, files and strings behave the same way. E.g., changing `from foo import bar` to `from foo import (bar,)` makes the outputs be the same.

This might or might not be related to https://github.com/PyCQA/isort/issues/1854

robsdedude:
When checking if a file got altered, leading and trailing white space is infact
relevant. Especially considering the existence of config options like
`lines_before_imports` and `lines_after_imports`.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -506,4 +506,4 @@ def _has_changed(before: str, after: str, line_separator: str, ignore_whitespace
             remove_whitespace(before, line_separator=line_separator).strip()
             != remove_whitespace(after, line_separator=line_separator).strip()
         )
-    return before.strip() != after.strip()
+    return before != after


robsdedude:
@timothycrosley it'd be great if you could find the time review this.

robsdedude:
@staticdev I'm sorry for the cold ping but I'd really like this to be moved forward :)

robsdedude:
Ah well. The `.strip()` calls I proposed to remove were introduced to fix #676. However, I'm pretty certain that simply stripping the output when checking for changes is not an appropriate fix for the issue as it introduces other issues the like one I'm trying to fix here. However, to get this PR to pass, I'd need to come up with a proper fix for 676 which turns out to be way more involved that I tough it was. I'm therefore abandoning this fix.

I think a proper fix for the linked issue would be to make sure that `core.py::process` properly splits import sections by not mixing them with other instructions (like the `print` statement in the issue's example).

PR #1992, opened 2022-11-11
Derive `settings_path` from `--filename`

### LINKED ISSUE #1989: `isort` does not search for config when using `--filename`
If you have files in this structure:
```
project
   |- sub-package
        |- pyproject.toml
        |- something.py       
```
say that `pyproject.toml` configures for profile black.

Then, this works (picks up the profile black):
```
C:\project> python -m isort C:\project\sub-package\something.py
```
but this does not:
```
C:\project> type C:\project\sub-package\something.py | python -m isort - --filename C:\project\sub-package\something.py
```
but this does (notice the CWD)
```
C:\project\sub-package> type C:\project\sub-package\something.py | python -m isort - --filename C:\project\sub-package\something.py
```

In VS Code extension for `isort` we are using the stdin way to perform sorting. This is need to support the organize import on save capability. I was able to simulate the same problem using `type` (on windows) and `cat` (on linux). 

reference issue: https://github.com/microsoft/vscode-isort/issues/138

kaste:
Fixes #1989

`--filename` is a shortcut setting used only in the stdin case to set
the `settings_path`.
Generally, `--settings-path=<file_path>` == `--filename=<filename>`
should hold.

This functionality was removed in 09734212 (Reuse config when items
passed in through stdin as used when items passed in explicitly).

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -1087,7 +1087,7 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
         return
     if "settings_path" not in arguments:
         arguments["settings_path"] = (
-            os.path.abspath(file_names[0] if file_names else ".") or os.getcwd()
+            os.path.abspath(file_names[0] if file_names else ".")
         )
         if not os.path.isdir(arguments["settings_path"]):
             arguments["settings_path"] = os.path.dirname(arguments["settings_path"])
--- a/isort/main.py
+++ b/isort/main.py
@@ -1087,6 +1087,8 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
         return
     if "settings_path" not in arguments:
         arguments["settings_path"] = (
+            arguments.get("filename", None) or os.getcwd()
+            if file_names == ["-"] else
             os.path.abspath(file_names[0] if file_names else ".")
         )
         if not os.path.isdir(arguments["settings_path"]):


isort/main.py
```diff
@@ -1087,7 +1087,9 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] =
         return
     if "settings_path" not in arguments:
         arguments["settings_path"] = (
-            os.path.abspath(file_names[0] if file_names else ".") or os.getcwd()
+            arguments.get("filename", None) or os.getcwd()
```
karthiknadig, line 1090:
```suggestion
            os.path.abspath(arguments.get("filename", None) or os.getcwd())
```
Should `--filename` argument be wrapped with `abspath` as well?

kaste: I think that's confusing because then it depends on the working dir again.  The working dir is not as easy to get right for editors; it might be the topmost open folder, or the first folder with .git in it.  And then we need to construct a relative filename of course to set it as the argument.

LukasVik:
I have checked out this branch and tried it in VSCode, and this does indeed solve all issues I had with https://github.com/PyCQA/isort/issues/1989 and https://github.com/microsoft/vscode-isort/issues/53.

Would it be possible to merge this soon?

I unfortunately have no input regarding the `abspath` or not.

karthiknadig (approved):

kaste:
@karthiknadig I just did the PR here to fix the bug.  I have no merge rights whatsoever.  🤷

karthiknadig:
@timothycrosley would it be possible for you to take a look at this?

PR #2017, opened 2022-12-06
Fix `black` compatibility for `.pyi` type stub files

### LINKED ISSUE #1976: `black` compatibility: ignore `lines_after_imports` for `.pyi` typing stubs
`isort` and `black` are two famous quality control tools. And usually, they are used together. [Black Compatibility](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html)

The user may set custom `lines_after_imports` in their configuration file (e.g. `pyproject.toml`):

```toml
[tool.black]
safe = true
line-length = 88
skip-string-normalization = true
target-version = ["py37", "py38", "py39", "py310"]

[tool.isort]
atomic = true
profile = "black"
indent = 4
line_length = 100
lines_after_imports = 2
multi_line_output = 3
```

For `.py` files, the maximum empty lines for `black` are 2 and everything works fine. However, for `.pyi` stubs files, `black` only allows a single blank line at the top level, while `isort` adds an extra blank line when setting `lines_after_imports = 2`.

https://github.com/psf/black/blob/b60b85b234d6a575f636d0a125478115f993c90c/src/black/lines.py#L485-L487

For `.pyi` files, if the user runs `black` first and then `isort`, there will be 2 empty lines after imports. If the user runs `isort` before `black`, there will be only 1 empty line after imports. Both `black` and `isort` will modify the `.pyi` files. And command:

```bash
isort --check *.pyi && black --check *.pyi
```

always fails if set `lines_after_imports = 2`.

The user have to add `isort: off` at the top of all `.pyi` files.

XuehaiPan:
Ignore `lines_after_imports` settings for `.pyi` type stub files if profile `black` is set. The `black` formatter does not allow `>= 2` line breaks in `.pyi` type stub files.

Fixes #1976


staticdev (changes requested):
Thanks for the improvement @XuehaiPan, would you mind also adding tests since you added new logic to the code?

XuehaiPan:
@staticdev I update the tests accordingly and they have successfully passed on my local machine.

staticdev (approved):

staticdev:
Thanks @XuehaiPan LGTM.

PR #2026, opened 2022-12-11
Add release workflow

staticdev:
Add release workflow - for it to work properly Github settings must have 2 secrets:
* PYPI_TOKEN - to send releases to pypi.org (when version on pyproject.toml changes)
* TEST_PYPI_TOKEN - to send non-release versions to test.pypi.org

Superseeds #2004

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,79 @@
+---
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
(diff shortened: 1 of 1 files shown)


staticdev:
@timothycrosley looks like the token for pypi PYPI_TOKEN is not working.. I run for the first time a release but I get forbidden error:
https://github.com/PyCQA/isort/actions/runs/3716321197/jobs/6302512647

PR #2028, opened 2022-12-12
Prepare release 5.11.0

staticdev:


timothycrosley-stripe (changes requested):
Hi! I would recommend we release this as 5.11.0, as major releases are discouraged unless there is major API breaking changes: https://pycqa.github.io/isort/docs/major_releases/release_policy.html major versions are just very disruptive at this point

staticdev:
> Hi! I would recommend we release this as 5.11.0, as major releases are discouraged unless there is major API breaking changes: https://pycqa.github.io/isort/docs/major_releases/release_policy.html major versions are just very disruptive at this point

You are right... will do that.

PR #2061, opened 2022-12-22
Remove unnecessary parentheses, use f-string, rm "object"

mwtoews:
These changes were discovered using:

    pyupgrade --py37-plus `find . -name "*.py"`


staticdev (changes requested):
@mwtoews create finding, thanks. I would like to add pyupgrade permanently and execute this command on our `scripts/lint.sh` as another `poetry run` command. Could you add that to your PR? For that you will also need to do `poetry add --dev pyupgrade>=3.3.1`.

DanielNoord:
This has been fixed on `main`.

Issue #2062, opened 2022-12-27
Incorrectly turned an import into a cimport in a cython 3 pure python file


jtoledo1974:
With the current version of cython it is possible to use [valid python syntax](https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#cimports) to do what used to need a cimport statement.

So isort turns 
```
import cython
from cython.cimports.libc import math


def use_libc_math():
    return math.ceil(5.5)
````

into 

```
import cython

from cython.cimports.libc cimport math


def use_libc_math():
    return math.ceil(5.5)
````

The resulting file cannot be run by the python interpreter, breaks tests and breaks the cython compilation
```
(venv311) C:\Users\Juan\cruji-enaire>cython -a test.py
C:\Users\Juan\venv311\Lib\site-packages\Cython\Compiler\Main.py:345: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: C:\Users\Juan\cruji-enaire\test.py
  tree = Parsing.p_module(s, pxd, full_module_name)

Error compiling Cython file:
------------------------------------------------------------
...
import cython

from cython.cimports.libc cimport math
                          ^
------------------------------------------------------------

test.py:3:26: Expected 'import' or 'cimport'
```

isort should be able to recognize the new `from cython.cimports` syntax and don't turn the import into a cimport.

PR #2063, opened 2022-12-27
Added test case and fix for cython pure python import

jtoledo1974:
Here is a test case demostrating the issue. I'll try to produce a patch as well.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -326,7 +326,10 @@ def process(
                             or " cimport " in import_statement
                             or " cimport*" in import_statement
                             or " cimport(" in import_statement
-                            or ".cimport" in import_statement
+                            or (
+                                ".cimport" in import_statement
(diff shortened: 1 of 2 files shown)


jtoledo1974:
Added fix for the test case that fixes #2062

staticdev:
@jtoledo1974 thank you for the fix. Just curious, in your first commit adding the test it does not fail in CI. It should, right?

jtoledo1974:
> @jtoledo1974 thank you for the fix. Just curious, in your first commit adding the test it does not fail in CI. It should, right?

In my fork of isort from master, the first commit that added the test was failing (it was turning the import into a cimport). With the second commit that adds the fix the test does not fail.

I'm not sure if that is what you were asking, though.

staticdev (approved):

PR #2074, opened 2023-01-18
Add more docs about filter-files option

micahjsmith:
The skip-related options (skip, extend-skip, skip-glob, extend-skip-glob, skip-gitignore) all interact with `filter-files` but don't mention it in their own documentation, and filter-files from its name wouldn't come up in text searches for "skip". Adds a mention of `filter-files` usage to the skip-related options docs.


timothycrosley (approved):
Thanks for these improvements :tada:

PR #2096, opened 2023-02-15
fix multi_line_output_modes docs

xinydev:
- There should not be a comma after the last import in the Vertical Hanging Indent format.
- Removing extra spaces.

The reason why I noticed this error is that I was using both isort and black, which can cause conflicts. The last import in the code formatted by black has a comma. After consulting the documentation, I found that the configuration for isort's Vertical Hanging Indent is the same as the code formatted by black. However, after configuring isort with Vertical Hanging Indent, I found that there was still a conflict between the two tools. Finally, I discovered that the isort documentation contained this error, and the last import should not have a comma.
I ultimately resolved the compatibility issue by using profile="black", but I do think that the error in the documentation should be fixed. Some people may encounter the same issue as me and be misled by the extraneous comma in the final import statement.


DanielNoord (approved):

staticdev (approved):
Thanks @xinydev and @DanielNoord, LGTM.

PR #2109, opened 2023-03-19
Resolve issue when non-indented lines get stripped with isort

### LINKED ISSUE #2087: isort uncomments code when odd indentation is next to an import
I came across a bug where isort strips off the leading characters (in my case a comment) when there's indentation next to an import. I found this in an older version of isort, but still see it in 5.12.0
```
#!/usr/bin/env python3

if True:
# this comment breaks
    import sys
```
when invoked with `isort break_isort.py` results in
```
#!/usr/bin/env python3

if True:
is comment breaks
    import sys
```

anirudnits:
- [ x ] Resolves #2087 
- [ x ] Added tests

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -332,10 +332,12 @@ def process(
                         ):
                             cimport_statement = True
 
-                        if cimport_statement != cimports or (
+                        if cimport_statement != cimports:
+                            indent = new_indent
+                        if (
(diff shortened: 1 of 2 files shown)


DanielNoord:
Superseded by https://github.com/PyCQA/isort/pull/2459

PR #2115, opened 2023-03-30
chore: move configurations to `pyproject.toml`

SauravMaheshkar:
This PR aims to minimize the number of configuration files for the project. Instead of a separate `.coveragerc`, and `.isort.cfg`, we now would have all the configurations within the `pyproject.toml` file leading to a overall minimal code structure.


staticdev (approved):

pyproject.toml
```diff
@@ -90,16 +90,32 @@ vulture = ">=1.0"
...
+source = [
+    "isort/",
+    ".tox/*/lib/python*/site-packages/isort/",
+    ".tox/*/lib/site-packages/isort/"
+]
 tests = ["tests", "*/tests"]
 
 [tool.coverage.run]
 branch = true
 source = ["git_portfolio", "tests"]
```
staticdev, line 102:
```suggestion
source = ["isort", "tests"]
```

staticdev: @SauravMaheshkar now I see we have a bug here.

staticdev (changes requested):

pyproject.toml
```diff
@@ -90,16 +90,32 @@ vulture = ">=1.0"
...
+source = ["isort", "tests"]
+omit = [
+    "isort/_vendored/*",
+    "tests/*",
+    "isort/deprecated/*",
+]
 
 [tool.coverage.report]
 show_missing = true
 fail_under = 100
```
staticdev, line 111:
```suggestion
fail_under = 99
```

staticdev (changes requested):

pyproject.toml
staticdev: @SauravMaheshkar I checked also that it was not failing before with old config but coverage was before 99.

staticdev (approved):

staticdev:
Thanks @SauravMaheshkar

PR #2119, opened 2023-04-04
Fix config overriding with --resolve-all-configs

hotpxl:
Without `--resolve-all-configs`, isort merges the options provided on the command line with the options found in config files. However this behavior is broken when `--resolve-all-configs` is specified. This change fixes that issue.


hotpxl:
bump - this is a fairly simple fix

DanielNoord:
@hotpxl Sorry for not getting to this sooner. Would it be possible to add a test for this?

DanielNoord:
Let's close for now as the author seems unresponsive.

PR #2125, opened 2023-04-21
Adds fuzz tests

jesslatimer:
This PR adds a basic fuzz test harness. It appears to have found at least one bug so far, so I hope this will be of help to you.


DanielNoord:
@jesslatimer Are you still interested in this? If so, we should probably add `atheris` as a dependency via `poetry`.

DanielNoord:
Let's close this for now!

PR #2137, opened 2023-05-20
Fix profile black: missing split_on_trailing_comma=True

janosh:
Came up in https://github.com/charliermarsh/ruff/issues/4153#issuecomment-1555986378.

### DIFF
--- a/isort/profiles.py
+++ b/isort/profiles.py
@@ -8,6 +8,7 @@ black = {
     "use_parentheses": True,
     "ensure_newline_before_comments": True,
     "line_length": 88,
+    "split_on_trailing_comma": True,
 }
 django = {
     "combine_as_imports": True,


johnthagen:
I think this is probably the right change to make, but I wanted to advocate that I personally find the current behavior optimal due to editors being able to automatically remove imports and not wanting to leave imports that might otherwise be able to be folded only a single line exploded and consuming vertical space. I personally see imports in a different class (mostly boilerplate) differently than what the black comma rule was more intended for.

But, at the same time I can see the argument to simply be consistent everywhere.

Thankfully the `split-on-trailing-comma` configuration still exists as an override.

- https://github.com/charliermarsh/ruff/issues/4153#issuecomment-1556254673

janosh:
@johnthagen @timothycrosley Would be great to get a final merge/reject decision here.

janosh:
@johnthagen @timothycrosley Bump

PR #2142, opened 2023-06-06
Make src_paths behave as expected when using --resolve-all-configs and improve performance

### LINKED ISSUE #2045: inconsistencies of known_first_party
I have a repo with this tree:
```
directory_root

    packageA
        setup.cfg
        module1.py
        module2.py

    packageB
        module3.py
```

With `packageA/module1.py` having the following contents:
```python
import module2
import packageB
import numpy
import os
```

and `packageA/setup.cfg` having the following contents:
```
[isort]
known_first_party = bla
```

If I run isort with:
`python -m isort --verbose --check-only packageA/module1.py packageA/module2.py packageB/module3.py`
I get the following output:
```

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.11.2

else-type place_module for module2 returned FIRSTPARTY
else-type place_module for packageB returned THIRDPARTY
else-type place_module for numpy returned THIRDPARTY
else-type place_module for os returned STDLIB
ERROR: /directory_root/packageA/module1.py Imports are incorrectly sorted and/or formatted.
SUCCESS: /directory_root/packageA/module2.py Everything Looks Good!
SUCCESS: /directory_root/packageB/module3.py Everything Looks Good!
```
which is what I expected. So if I run isort without *--check-only* option, the original `packageA/module1.py` file is reformatted to:
```python
import os

import numpy
import packageB

import module2
```

However, I want to start having [support for multiple config files in a single isort run](https://pycqa.github.io/isort/docs/configuration/config_files.html#supporting-multiple-config-files-in-single-isort-run), so I have to use the `--resolve-all-configs` argument:
`python -m isort --verbose --resolve-all-configs --check-only packageA/module1.py packageA/module2.py packageB/module3.py`
which gives me the following output:
```
                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.11.2

./packageA/setup.cfg used for file packageA/module1.py
ERROR: /directory_root/packageA/module1.py Imports are incorrectly sorted and/or formatted.
./packageA/setup.cfg used for file packageA/module2.py
default used for file packageB/module3.py
```

***First Question:*** 
Why is the verbose debug information different when using the `--resolve-all-configs` argument? 
The `else-type place_module for <some_import> returned <SECTION_NAME>` messages are gone, as well as the `SUCCESS: <some_file> Everything Looks Good!` message.

Now, let's see the reformatting result when I run with argument `--resolve-all-configs` and without *--check-only* option:
```python
import os

import module2
import numpy

import packageB
```

To me this is quite unexpected, *packageB*  went from THIRDPARTY to FIRSTPARTY and *module2* went from FIRSTPARTY to THIRDPARTY.

***Second Question:*** 
(related to known_first_party inconsistencies!)
Why is the sorting result different when using or not the the `--resolve-all-configs` argument, if the config .cfg file used is the same?
If the same `packageA/setup.cfg` is used, I would expect the same sorting result.

Furthermore, I noticed  using `--resolve-all-configs` argument makes isort much slower, taking a couple seconds for such a simple repo. 
However, I also noticed that adding more folders/packages makes isort even slower, even if the request is exactly the same (sorting module1/2/3 only in package A and B).
For example, if I added 20 more packages (`packageC`, `packageD`, ..., `packageZ`), with plenty of folders inside each one, the isort run with the `--resolve-all-configs` argument took almost 20 seconds, even if only 3 modules in 2 packages are being reformatted.

***Third Question:*** 
Why does isort with `--resolve-all-configs` argument become slower when adding more folders/packages, even if it is still sorting the same number of files?
Shouldn't the search for the closest config file happen only for the files that are requested to sort, thus making isort run time only dependent on the number of files to check/reformat instead of the repo size?

Finally, I went back to testing the isort standard behavior when using only a single config file (*without* the `--resolve-all-configs` argument).
I changed the `packageA/setup.cfg` file by removing the known_first_party entry, thus making it empty, which I expect to make it resort to default configuration:
```
[isort]

```
If I run isort again with:
`python -m isort --verbose packageA/module1.py packageA/module2.py packageB/module3.py`
I get the following output:
```

                 _                 _
                (_) ___  ___  _ __| |_
                | |/ _/ / _ \/ '__  _/
                | |\__ \/\_\/| |  | |_
                |_|\___/\___/\_/   \_/

      isort your imports, so you don't have to.

                    VERSION 5.11.2

else-type place_module for module2 returned THIRDPARTY
else-type place_module for packageB returned FIRSTPARTY
else-type place_module for numpy returned THIRDPARTY
else-type place_module for os returned STDLIB
Fixing /directory_root/packageA/module1.py
```
which reformats the original file in this way:
```python
import os

import module2
import numpy

import packageB
```

Again, compared to the original configuration, *packageB*  went from THIRDPARTY to FIRSTPARTY and *module2* went from FIRSTPARTY to THIRDPARTY.

***Fourth (and Final) Question:*** 
(related to known_first_party inconsistencies!)
Why do both *packageB* and *module2* change sections if I have a default configuration vs `known_first_party = bla`, where `bla` was not even a package existing in my repo?

For `module1` formatting, I expected always to have `module2` as FIRSTPARTY because it's part of the same python package and lives in the same directory, and `packageB` as THIRDPARTY because it's a separate package from `packageA` where `module1` lives.
And why setting known_first_party to some specific name would have an impact? Does it replace the default FIRST_PARTY, or extend?

Additionally and probably the most important question in this issue, why does `--resolve-all-configs` argument change the sorting result if exactly the same config file is applied?

sudowork:
When using `--resolve-all-configs`, there is unexpected behavior in that `src_paths` ends up resolving relative to the project root, which defaults to the current working directory. This results in first-party modules being marked as third-party modules in the default case.

Under the previous implementation, one possible workaround would be to specify the relative path to config directory (e.g. `relative/path/to/configdir/src`). However, assuming that the most common use of `--resolve-all-configs` is to support multiple sub-projects in the same repository/overall directory, this workaround would now require each sub-project to understand where it lives in the filesystem.

This change proposes a fix that sets `directory` on the `config_data` to be the directory containing the used configuration file if not already set. Downstream, this directory is then used to resolve the absolute paths specified by `src_paths`.

This change also introduces performance improvements to `find_all_configs` by pruning as we walk the filesystem and other smaller performance enhancements.

Fixes #2045

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -19,6 +19,7 @@ from typing import (
     Dict,
     FrozenSet,
     Iterable,
+    Iterator,
     List,
     Optional,
     Pattern,
@@ -806,26 +807,41 @@ def find_all_configs(path: str) -> Trie:
     """
     trie_root = Trie("default", {})
 
-    for dirpath, _, _ in os.walk(path):
-        for config_file_name in CONFIG_SOURCES:
-            potential_config_file = os.path.join(dirpath, config_file_name)
-            if os.path.isfile(potential_config_file):
-                config_data: Dict[str, Any]
-                try:
-                    config_data = _get_config_data(
-                        potential_config_file, CONFIG_SECTIONS[config_file_name]
-                    )
-                except Exception:
-                    warn(f"Failed to pull configuration information from {potential_config_file}")
-                    config_data = {}
-
-                if config_data:
-                    if "directory" not in config_data:
-                        config_data["directory"] = dirpath
-                    trie_root.insert(potential_config_file, config_data)
-                    break
+    config_sources_set = set(CONFIG_SOURCES)
+    for potential_config_file in _scanwalk_files(
+        path, exclude_fn=lambda entry: entry.name in DEFAULT_SKIP
+    ):
+        if potential_config_file.name not in config_sources_set:
+            continue
+        try:
+            config_data = _get_config_data(
+                potential_config_file.path, CONFIG_SECTIONS[potential_config_file.name]
+            )
+            if "directory" not in config_data:
+                config_data["directory"] = os.path.dirname(potential_config_file.path)
+            trie_root.insert(potential_config_file.path, config_data)
+        except Exception:
+            warn(f"Failed to pull configuration information from {potential_config_file.path}")
 
     return trie_root
+
+
+def _scanwalk_files(
+    root_path: str, exclude_fn: Callable[[os.DirEntry], bool] = None
+) -> Iterator[os.DirEntry]:
+    # depth-first walk of file system starting with root_path
+    stack: List[str] = [root_path]
+    while stack:
+        dir_path = stack.pop()
+        with os.scandir(dir_path) as it:
+            for entry in it:
+                # Avoid processing excluded files/dirs and their descendants
+                if exclude_fn and exclude_fn(entry):
+                    continue
+                if entry.is_dir():
+                    stack.append(entry.path)
+                elif entry.is_file():
+                    yield entry
 
 
 def _get_config_data(file_path: str, sections: Tuple[str, ...]) -> Dict[str, Any]:
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -820,6 +820,8 @@ def find_all_configs(path: str) -> Trie:
                     config_data = {}
 
                 if config_data:
+                    if "directory" not in config_data:
+                        config_data["directory"] = dirpath
                     trie_root.insert(potential_config_file, config_data)
                     break
 
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -11,7 +11,7 @@ import re
 import stat
 import subprocess  # nosec # Needed for gitignore support.
 import sys
-from collections.abc import Callable, Iterable
+from collections.abc import Callable, Iterable, Iterator
 from dataclasses import asdict, dataclass, field
 from importlib.metadata import EntryPoints
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


sudowork:
One thing of note: This would be backwards incompatible for anyone depending on the behavior of resolving first-party `src_paths` from the project root or config root. However, my assumption is that the behavior with this change is the most expected given the types of config files we look for (.isort.cfg, pyproject.toml, setup.cfg, tox.ini, .editorconfig). The only odd one out in my opinion would be `.editorconfig`, which supports having a non-root configuration; however, given the behavior of the `--resolve-all-configs` flag to only choose the closest config file, I opted to not handle this case for consistency and to avoid further complicating the code.

isort/settings.py
```diff
@@ -806,28 +807,43 @@ def find_all_configs(path: str) -> Trie:
...
+        path, exclude_fn=lambda entry: entry.name in DEFAULT_SKIP
+    ):
+        if potential_config_file.name not in config_sources_set:
+            continue
+        try:
+            config_data = _get_config_data(
+                potential_config_file.path, CONFIG_SECTIONS[potential_config_file.name]
+            )
+            if "directory" not in config_data:
+                config_data["directory"] = os.path.dirname(potential_config_file.path)
```
sudowork, line 821: This is the main logic to fix the `src_paths` resolution.

sblask:
I just ran into this problem. I have a mono repo with multiple python projects in sub folders that each have an .isort.cfg. When running isort from the root on a file in directory one, everything works as expected, same on a file in directory two. But when I run on a file from directory one and two at the same time (when using pre-commit), only one of the files is formatted correctly. I thought `--resolve-all-configs` would solve this, but quite the opposite, when using it, neither works. I'd expect the folder with the config file to be set as root for the files in them. Sounds like this PR would solve this?

DanielNoord:
This needs a rebase unfortunately. Since I can't use Copilot on another repository I can't use it to fix the PR. Unfortunately I'll close it for now. Sorry! We should have gotten to this earlier, but maintenance was a bit low in previous years.

sudowork:
@DanielNoord I can go ahead and rebase this if you re-open. Or alternatively, I can just open a new PR.

sudowork:
@DanielNoord rebased and also addressed Copilot's feedback + fixed a edge case where config priority wasn't being respected.

### Performance

Benchmarked the current implementation vs the previous `os.walk` version on identical trees (`find_all_configs`, warmup + timed runs, median):

| Tree | `os.walk` (old) | `os.scandir` (new) | Speedup |
| --- | --- | --- | --- |
| Synthetic, with `.venv`/`node_modules`/`build`/`dist` (518 dirs / 8.1K files) | 31.5 ms | 13.0 ms | ~2.4× |
| Synthetic, no skip dirs (500 dirs / 8K files) | 42.6 ms | 40.8 ms | ~1.07× |
| Large real-world monorepo | 8.75 s | 0.36 s | **~24×** |

Most of the benefit comes from pruning `DEFAULT_SKIP`.

DanielNoord:
@sudowork I think the review from Copilot correctly hints to the fact that the speed optimization might better fit a separate PR as there is a lot of "gotcha's". Can we split this into a PR that does the fix and a PR that does the speed improvement?

Manny7717 (approved):
7cd7cfa1ccc6d1f5c3e463a8bc3c76636af7d43f

Verified locally against head `7cd7cfa1ccc6d1f5c3e463a8bc3c76636af7d43f` (isort main + this PR):

- **Bug reproduced:** on main, with `--resolve-all-configs`-style trie resolution, a subdir config `src_paths = ["src"]` resolves against the invocation CWD (e.g. `<isort-repo>/src`), not the config's own directory. With this PR it resolves to `<subdir>/src` (verified end-to-end via `find_all_configs` → `Config(**data)` for two subdirs with different `src_paths`).
- **Refactor equivalence checked:** per-directory config precedence now enforced via `config_source_priority` (scandir order is arbitrary — correct call); broken-config fall-through preserved; symlinked dirs not descended (matches `os.walk(followlinks=False)`); `DEFAULT_SKIP` dirs now excluded (intended, covered by the new `.venv` case in `test_find_all_configs`).
- **Tests:** updated `test_find_all_configs` + new assertions pass; full suite on head == full suite on main: **624 passed / 6 failed / 2 errors on BOTH**, and the failure sets are byte-identical (pre-existing env noise: benchmark `FileNotFoundError`, shared-profiles tests, skip-paths issue tests) — no regression introduced.
- **Docs:** options.md updated to document the new relative-to-config resolution.

Two non-blocking notes: (1) `if "directory" not in config_data` means a user config containing an explicit `directory` key would win over the injected path — `directory` is not a documented setting, so acceptable, but worth being aware of; (2) the `directory` key now flows into trie results where it previously didn't — `Config` accepts it (existing field), and no other consumer of `find_all_configs` output appears affected.

Solid fix with the right test coverage — approving.

DanielNoord (changes requested):
Let's indeed split out the performance changes into a different PR.

PR #2153, opened 2023-07-07
Add --exclude and --exclude-glob flags

idofran:
Both _flake8_ and _black_ have `--exclude` flag which works the same as `--skip-glob` in _isort_.

This is highly confusing because `--skip-glob` isn't a real glob, and works using `fnmatch`.

aliasing `--exclude` to `--skip-glob` and creating new real glob option called `--exclude-glob` will make `isort` compatible with `--exclude` option in other utilities and add a new glob option if anyone really meant to use glob.

In my opinion, it is best to deprecate `--skip-glob` and replace it completely with `--exclude`, however I'm not sure what is the policy here for doing it.

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -126,6 +126,7 @@ Files that isort should skip over. To even skip matching files that have been sp
 
 - --sg
 - --skip-glob
+- --exclude
 
 **Examples:**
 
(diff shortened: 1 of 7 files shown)


DanielNoord:
Going to close for now as it has grown stale and there is also no issue for this.

PR #2196, opened 2023-11-06
Fix assertions in `test_git_hook`

mgorny:
Fix `called_once()` assertions in `test_git_hook` to use the correct `assert_called_once()` method.  The former does not exist, so it evaluates to a mocked method in Python < 3.12, making the assert meaningless, and it triggers an error in Python 3.12+.

While at it, split the mock into two because otherwise the test would fail because two `hooks.git_hook()` calls imply two mock calls.


staticdev (approved):
Thanks for the fix @mgorny

mgorny:
Thanks!

PR #2202, opened 2023-12-01
use the standard library tomllib on sufficiently new python

eli-schwartz:
In python 3.11 it is no longer necessary to vendor (?) a toml implementation. The vendored one is now part of the standard library. Simply use that instead.

Benefits include:
- micro-optimization when using the isort library API in an application that already uses tomllib, to avoid reading yet another one from on-disk
- seamless support for automatically dropping tomli once the minimum python version is upgraded, via tools such as pyupgrade
- distributors who patch out isort to use the system tomli, do not actually have to package tomli for versions of python that have tomllib


isort/settings.py
```diff
@@ -45,9 +45,12 @@
...
 
 if TYPE_CHECKING:
-    tomli: Any
+    tomllib: Any
 else:
-    from ._vendored import tomli
+    if sys.version_info >= (3, 11):
+        import tomllib
+    else:
+        from ._vendored import tomli
```
staticdev, line 53:
```suggestion
        from ._vendored import tomli as tomllib
```

this will not work.. you will probably need this to keep working with older versions

staticdev (changes requested):

PR #2208, opened 2023-12-10
Fix integration tests

bp72:
The following projects don't use isort anymore and use ruff instead: 
- tmuxp - https://github.com/tmux-python/tmuxp/blob/master/pyproject.toml#L151
- fastapi - https://github.com/tiangolo/fastapi/blob/master/pyproject.toml#L130
- websockets - https://github.com/python-websockets/websockets/blob/main/pyproject.toml#L85
- typeshed - https://github.com/python/typeshed/blob/main/pyproject.toml#L29
- poetry - https://github.com/python-poetry/poetry/blob/master/pyproject.toml#L113
- pillow - https://github.com/python-pillow/Pillow/blob/main/pyproject.toml#L103
- attrs - https://github.com/python-attrs/attrs/blob/main/pyproject.toml#L231
- DataDog - https://github.com/DataDog/integrations-core/pull/14212/files

### DIFF
--- a/tests/integration/test_projects_using_isort.py
+++ b/tests/integration/test_projects_using_isort.py
@@ -42,53 +42,14 @@ def test_pandas(tmpdir):
     run_isort((str(tmpdir / "pandas"), "--skip", "__init__.py"))
 
 
-def test_fastapi(tmpdir):
-    git_clone("https://github.com/tiangolo/fastapi.git", tmpdir)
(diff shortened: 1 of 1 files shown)


staticdev:
Thanks a lot @bp72 I was checking exactly that. Let's see, maybe it makes sense to join forces with ruff.

bp72:
@staticdev thank you for the review!

Issue #2234, opened 2024-02-15
[self tests] test_isort_should_warn_on_empty_custom_config_issue_1433 fails against Pytest 8


stanislavlevin:
```python
______________ test_isort_should_warn_on_empty_custom_config_issue_1433 ______________

tmpdir = local('/usr/src/tmp/pytest-of-builder/pytest-3/test_isort_should_warn_on_empt0')

    def test_isort_should_warn_on_empty_custom_config_issue_1433(tmpdir):
        """Feedback should be provided when a user provides a custom settings file that has no
        discoverable configuration.
        See: https://github.com/PyCQA/isort/issues/1433
        """
        settings_file = tmpdir.join(".custom.cfg")
        settings_file.write(
            """
    [settings]
    quiet = true
    """
        )
        with pytest.warns(UserWarning):
            assert not Config(settings_file=str(settings_file)).quiet
    
        settings_file.write(
            """
    [isort]
    quiet = true
    """
        )
>       with pytest.warns(None) as warning:  # type: ignore

tests/unit/test_ticketed_features.py:632: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = WarningsChecker(record=True), expected_warning = None, match_expr = None

    def __init__(
        self,
        expected_warning: Optional[
            Union[Type[Warning], Tuple[Type[Warning], ...]]
        ] = Warning,
        match_expr: Optional[Union[str, Pattern[str]]] = None,
        *,
        _ispytest: bool = False,
    ) -> None:
        check_ispytest(_ispytest)
        super().__init__(_ispytest=True)
    
        msg = "exceptions must be derived from Warning, not %s"
        if expected_warning is None:
>           warnings.warn(WARNS_NONE_ARG, stacklevel=4)
E           pytest.PytestRemovedIn8Warning: Passing None has been deprecated.
E           See https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests for alternatives in common use cases.

/usr/lib/python3/site-packages/_pytest/recwarn.py:279: PytestRemovedIn8Warning
```

With Pytest 7:
    
https://docs.pytest.org/en/stable/changelog.html#pytest-7-0-0rc1-2021-12-06
    
> #8645: pytest.warns(None) is now deprecated because many people used
[...]

PR #2235, opened 2024-02-15
Drop deprecated pytest.warns(None)

stanislavlevin:
With Pytest 7:

https://docs.pytest.org/en/stable/changelog.html#pytest-7-0-0rc1-2021-12-06

> #8645: pytest.warns(None) is now deprecated because many people used
  it to mean “this code does not emit warnings”, but it actually had the
  effect of checking that the code emits at least one warning of any
  type-like pytest.warns() or pytest.warns(Warning)

With Pytest 8 it's the error.

Changed according to the documentation:
https://docs.pytest.org/en/latest/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

Fixes: https://github.com/PyCQA/isort/issues/2234

### DIFF
--- a/tests/unit/test_ticketed_features.py
+++ b/tests/unit/test_ticketed_features.py
@@ -3,6 +3,7 @@ it fully works as defined in the associated ticket.
 """
 from functools import partial
 from io import StringIO
+import warnings
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


DanielNoord:
Included in https://github.com/PyCQA/isort/pull/2329

Issue #2128, opened 2023-04-30
Isort is disregarding the multiply line import


Cheukting:
Found in version 5.12.0 of isort from solving https://github.com/Zac-HD/shed/pull/94

Multiline import will automatically be put in one line.

Here is an example:
```
from os import (
    waitpid,
    waitstatus_to_exitcode,
    walk,
    write,
    writev,
)

print(
    waitpid,
    waitstatus_to_exitcode,
    walk,
    write,
    write,
)
```

Becomes

```
from os import waitpid, waitstatus_to_exitcode, walk, write, writev

print(
    waitpid,
    waitstatus_to_exitcode,
    walk,
    write,
    write,
)

```

CC @Zac-HD

Cheukting:
This is starting to give me some pain. I am happy to look into a possible solution and attempt a PR.

Zac-HD:
More specifically, the problem is that isort doesn't respect the "magic trailing comma" which Black uses, even when using profile=black.

MrMino:
It doesn't do that even when given `--split-on-trailing-comma` option. Looks like a regression of the feature introduced in #1876?

staticdev:
Closed by @2236

PR #2236, opened 2024-02-16
Black profile: enable magic comma

MrMino:
When `black` encounters a trailing comma in the `from ... import` list, it expands it vertically. Currently, `isort --profile=black` doesn't do that, which confuses people (including me :)), see #2128.

This PR enables `split_on_trailing_comma`, which is the default Black's behavior.

_I wasn't sure if I should create a new changelog version header, so i omitted it, since that would require deciding whether this is PATCH or MINOR._

---

What Black does:

```
$ echo "from x import (a, b, c,)" | black -

from x import (
    a,
    b,
    c,
)
reformatted -

All done! ✨ 🍰 ✨
1 file reformatted.
```

What isort does _currently_:

```
$ echo "from x import (a, b, c,)" | isort --profile=black -

from x import a, b, c
```

After this change:

```
$ echo "from x import (a, b, c,)" | isort --profile=black -

from x import (
    a,
    b,
    c,
)
```

### DIFF
--- a/docs/configuration/profiles.md
+++ b/docs/configuration/profiles.md
@@ -12,6 +12,7 @@ To use any of the listed profiles, use `isort --profile PROFILE_NAME` from the c
 
  - **multi_line_output**: `3`
  - **include_trailing_comma**: `True`
+ - **split_on_trailing_comma**: `True`
  - **force_grid_wrap**: `0`
  - **use_parentheses**: `True`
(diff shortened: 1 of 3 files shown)


DanielNoord (approved):
Seems sensible.

barrelful (approved):

staticdev (approved):

PR #2238, opened 2024-02-20
Add `--required-version`

archont94:
Adding functionality to verify (with CLI or config) if currently used isort version matches required version.

Similar to [black implementation](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#required-version).

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -554,8 +554,8 @@ class Config(_Config):
                 line = fp.readline(100)
         except OSError:
             return False
-        else:
-            return bool(_SHEBANG_RE.match(line))
+
+        return bool(_SHEBANG_RE.match(line))
 
     def _check_folder_git_ls_files(self, folder: str) -> Optional[Path]:
         env = {**os.environ, "LANG": "C.UTF-8"}


archont94:
@staticdev I fixed [Unnecessary else / elif used after return
PYL-R1705](https://app.deepsource.com/gh/PyCQA/isort/run/79be935a-3907-4b61-8b1d-6703c51fcc8e/python/PYL-R1705?listindex=0) but remaining error is related to [`main` cyclomatic complexity higher than threshold](https://app.deepsource.com/gh/PyCQA/isort/run/5f9c07c6-7e99-41a6-b86d-591ae9246631/python/PY-R1000?listindex=0) and I'm not gonna touch it as it might mess with a lot of other things.

Please let me know how to proceed.

DanielNoord:
This really feels like a thing you should fix in your CI/environment with correct tooling that is meant for dependency management.

Up to @staticdev but I would suggest closing this as it just adds unnecessary complexity to the tool.

DanielNoord:
I still feel this should not be done on a per tool level but by using `uv` or friends.

PR #2241, opened 2024-02-28
`wemake` has 80 chars hard limit, not 79

sobolevn:
Proof: 
- https://github.com/wemake-services/wemake-python-styleguide/blob/298721659dfdcb93144aff394576a04fdb5fd03e/setup.cfg#L20
- https://github.com/wemake-services/wemake-python-styleguide/actions/runs/8076865527/job/22066047320?pr=2864

Refs:
- https://github.com/PyCQA/isort/pull/2183
- https://github.com/PyCQA/isort/issues/2182
- https://github.com/wemake-services/wemake-python-styleguide/pull/2864

PR #2256, opened 2024-04-14
Ignore `bandit`'s CVE reported by `safety`

sobolevn:
It is a dev-dependency and there's no real vulnerability. This is why a job in my CI fails: https://github.com/PyCQA/isort/pull/2241

Link to CVE: https://data.safetycli.com/v/64484/f17

Or you can bump `bandit` to 1.7.8: https://pypi.org/project/bandit/1.7.8/


DanielNoord:
We should indeed probably just bump `bandit` as that is more future proof. Are you up for making a PR?

sobolevn:
I think that the issue is gone, see the recent CI runs: they are green.

PR #2264, opened 2024-05-14
Speed up exists_case_sensitive calls

correctmost:
These changes save ~35ms when running isort on the yt-dlp codebase on Linux (~3.2secs -> 3.165secs, ~1% of the total time).

The savings should be a bit greater on Windows and macOS because repeated `os.listdir` calls are avoided too.

The main motivation behind these changes, however, was to speed up pylint, which relies on isort for import-order checks.  When running pylint on a private codebase with ~2200 Python files, I see a reduction of ~10,000 stat calls after applying this PR.


DanielNoord (approved):
Seems sensible indeed!

staticdev (approved):

staticdev:
LGTM. Thanks @correctmost @DanielNoord @akaFM

PR #2284, opened 2024-08-04
fix: CLI --src flag glob patterns

### LINKED ISSUE #2001: src_path glob patterns do not work from command line flag
Running `isort --src '*/src/' .` results in 
```
  File "/Users/pgorniak/.pyenv/versions/3.8.13/bin/isort", line 8, in <module>
    sys.exit(main())
  File "/Users/pgorniak/.pyenv/versions/3.8.13/lib/python3.8/site-packages/isort/main.py", line 1110, in main
  File "/Users/pgorniak/.pyenv/versions/3.8.13/lib/python3.8/site-packages/isort/settings.py", line 470, in __init__
    for path in full_paths:
  File "/Users/pgorniak/.pyenv/versions/3.8.13/lib/python3.8/pathlib.py", line 1136, in glob
    drv, root, pattern_parts = self._flavour.parse_parts((pattern,))
  File "/Users/pgorniak/.pyenv/versions/3.8.13/lib/python3.8/pathlib.py", line 75, in parse_parts
    drv, root, rel = self.splitroot(part)
  File "/Users/pgorniak/.pyenv/versions/3.8.13/lib/python3.8/pathlib.py", line 294, in splitroot
    if part and part[0] == sep:
TypeError: 'PosixPath' object is not subscriptable
```
I think because paths are resolved first using pathlib at https://github.com/PyCQA/isort/blob/main/isort/main.py#L1117.

If run via `isort .` but with a config file entry like
```
[settings]
src_paths=*/src/
```
the pattern works fine because the `src_paths` entry doesn't exist yet at the line above and is only read from the config file later, meaning it remains a string and the glob works fine.

marcelo-souzaf:
# Context

Resolves #2001.
A set is created with the resolved paths to avoid multiplicity, but paths can't be used as glob expressions, so the code fails when initializing the `Config` instance.
Given that there is logic to keep the src paths unique after evaluating them, the changes fix the bug harmlessly.

# Overview

* Change the type of the config value `"src_paths"` from `pathlib.Path` to `str`.

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -1116,9 +1116,7 @@ def main(argv: Optional[Sequence[str]] = None, stdin: Optional[TextIOWrapper] = 
         config_trie = find_all_configs(config_dict.pop("config_root", "."))
 
     if "src_paths" in config_dict:
-        config_dict["src_paths"] = {
-            Path(src_path).resolve() for src_path in config_dict.get("src_paths", ())
-        }
+        config_dict["src_paths"] = set(config_dict.get("src_paths", ()))
 
     config = Config(**config_dict)
     if show_config:


DanielNoord (commented):
This feels like it could do with a small regression test. Do you think you could add one?

marcelo-souzaf:
I've added one, but I'm not sure if this is the best way to test the CLI behavior.

DanielNoord (commented):
I missed this the first time I reviewed this, but changing the type of `src_paths` could be problematic. @staticdev What do you think?

DanielNoord:
Superseded by https://github.com/PyCQA/isort/pull/2497, thanks for the initial pr!

PR #2279, opened 2024-07-18
Added _collections_abc to standard library modules

mtrd3v:
This PR fixes issue #2250 by adding _collections_abc to the known standard library modules in mkstdlibs.py.

PR #2295, opened 2024-10-08
Use `sys.stdlib_module_names` to complement known stdlib modules

### LINKED ISSUE #2250: isort does not recognize _collections_abc as part of the standard library
Hello,

In my repository:
https://github.com/LLyaudet/python-repeatable-iterable/blob/main/src/python_repeatable_iterable/__init__.py
I have the following imports:

```python3
from typing import Iterable, Iterator, TypeVar, cast
from _collections_abc import dict_items, dict_keys, dict_values

from python_none_objects import NoneIterable
```

But isort does not recognize that 
```python3
from _collections_abc import dict_items, dict_keys, dict_values
```
is part of the standard library and regroup it like that:

```python3
from typing import Iterable, Iterator, TypeVar, cast

from _collections_abc import dict_items, dict_keys, dict_values
from python_none_objects import NoneIterable
```

Best regards,
     Laurent Lyaudet

devdanzin:
This PR attempts to fix #2250 by adding the names from `sys.stdlib_module_names` (or `sys.builtin_module_names` for versions below 3.10) to the known stdlib modules lists (and as such is an alternative to #2279).

It still uses Sphinx to download known module names, adding the names from `sys` to the downloaded lists. Some of the new names (`antigravity`, `this`) are excluded, please let me know if any others should be too.

A test for some of the new known modules is included.

Happy to apply any suggestions to the code.


DanielNoord:
With https://github.com/PyCQA/isort/pull/2306 we will be using the `stdlibs` package.

However, as they say themselves in the [README](https://github.com/omnilib/stdlibs) on newer versions this shouldn't be necessary. It would be good to rebase this after we merge the linked PR and replace where possible. If all is well that shouldn't create a diff in the actual stdlib list but only in the generating code.

staticdev (changes requested):
@devdanzin thanks a lot by your contribution. Unfortunately this one was a bit forgoten and got outdated. Do you mind updating it and if possible also adding 3.13?

devdanzin:
> Unfortunately this one was a bit forgoten and got outdated. Do you mind updating it and if possible also adding 3.13?

Sure, I'll do it if we decide it's worth it, but I see main is using the `stdlibs` package to gather module names since #2306, that might be as good as or even better than using `sys.stdlib_module_names`. It's up to you whether we keep it like that or use the internal list of module names from `sys`.

Also, I see that the new code skips "private" stdlib modules that start with an underscore, while my original patch included them, because https://github.com/PyCQA/isort/issues/2250 is about such a module. So we should decide whether to keep it without "private" modules, or add them back (`stdlibs` has all of them). I think having all possible module names listed wouldn't hurt and might avoid further bug reports for incorrect sorting of "private" modules.

devdanzin:
Aha, and my patch is incorrect anyway: it uses `sys.stdlib_module_names` of the running Python version instead of the target version. It'd be possible to make a dance to launch the target versions and fetch the module names from them, but using the `stdlibs` package seems cleaner.

With that, this PR should only focus on whether to include private modules and whether to remove some entries, like `antigravity`.

DanielNoord:
I think we should include these private modules as being part of the `stdlib`. Is that a change you could make in this PR?

devdanzin:
> Is that a change you could make in this PR?

Sure, I've done that. Sorry for the late reply.

The PR currently removes the `__future__`, `__main__`, `antigravity` and `this` modules from the set of known modules, but leaves `__hello__` and `__phello__`. Do you have any preferences about which should be excluded/included?

DanielNoord:
I don't really but I'd like @staticdev to weigh on this

isort/stdlibs/py310.py
```diff
@@ -6,10 +6,105 @@
...
+    "_warnings",
+    "_weakref",
+    "_weakrefset",
+    "_winapi",
+    "_xxsubinterpreters",
+    "_xxtestfuzz",
+    "_zoneinfo",
     "abc",
     "aifc",
-    "antigravity",
```
staticdev, line 12: I would prefer not to remove `antigravity` and other modules that are shipped with python.. they are valid anyway.

staticdev (changes requested):

isort/stdlibs/py310.py
devdanzin: Done.

devdanzin:
Sorry, I let this one slip by. The removal of `antigravity` and `this` is reverted, the private modules starting with underscores are added, and `mkstdlibs.py` still uses the `stdlibs` package. I think this might be done.

devdanzin:
Hm, integration tests are failing because pandas recorded private modules like `_csv` and `_strptime ` as non-stdlib. Not sure how to address this.

There's also a Django failure that I don't understand why it would happen:
```diff
-from django.db.models.sql.compiler import SQLUpdateCompiler
+from django.db.models.sql.compiler import (
+    SQLUpdateCompiler,
+)
```

Can this last one be unrelated to this PR?

DanielNoord:
@devdanzin Are you still interested in pursuing this?

devdanzin:
Sure, please let me know how you'd like me to handle the open issues.

DanielNoord:
@staticdev This is ready for review. It creates an issue for `pandas` but honestly I think this change is fine (and is also in line with how `ruff` sorts imports). These private modules are part of the `stdlib` and should be considered as such :)

Shall we merge this? :)

Issue #2299, opened 2024-10-15
CI uses deprecated Github actions


Rotzbua:
### Problem

CI uses deprecated Github actions.

### Output

Following warning is thrown on every GH action run:

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/cache@v2, actions/setup-python@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

### Solution

* Update GH actions.
* Add dependabot to keep track of updated actions.

### Reference

https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

kurtmckee:
@DanielNoord This was fixed by recent PRs.

PR #2300, opened 2024-10-15
feat: add dependabot for GH action update

Rotzbua:
### Advantage

Dependabot can automatically look for updated GH actions and rise a PR.

Since dependabot can be very annoying and spam a lot of PRs the GH actions related PRs are grouped into one single PR. Also the update frequency is limited to monthly.

### Reference

#2299

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


kurtmckee (approved):
@timothycrosley @staticdev Please review and merge this PR.

Fixes #2299

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
DanielNoord, line 4: Should we also add `pip` here so we get updates for stuff defined in `poetry`?

kurtmckee: It might be worthwhile, though almost all dependencies in `pyproject.toml` are dev dependencies.

Would it be an option to merge this as-is and then add `pip` later if still desired?

DanielNoord (approved):

Issue #2216, opened 2023-12-29
It seems `--jobs` and `--diff` cannot be used together


ian-h-chamberlain:
I couldn't find any documention or other issues on this, so I'm not sure if this is intended behavior (and undocumented) or just a bug. 

Tested with `isort 5.13.2`, running `isort --diff --jobs=2` acts the same as `isort --jobs=2` (that is, the `--diff` option is ignored). I could imagine there might be a concern about interleaving diff output from different jobs, in which case I think there are a couple options:

1. Disallow `--diff --jobs` (and equivalent config) entirely and display an error (and document this behavior)
2. Implement some kind of locking to serialize the diffs from different jobs
3. Do nothing to prevent interleaving and hope for the best (maybe document this as a caveat?)

The cause of the current behavior seems pretty clear: `show_diff` is not plumbed into the executor if `jobs` is used: https://github.com/PyCQA/isort/blob/main/isort/main.py#L1193-L1223 

From a quick test, adding `show_diff=show_diff` to the `executor.imap` call seems to work in simple cases (effectively option 3) but I haven't tested it extensively with lots of files at once or high job counts.

mnakama:
Thanks for documenting this. I just hit the same thing. Since there's no documentation or warning, I'm 80% sure it's just a bug. Since no maintainers chimed in, it's probably worth a PR.

staticdev:
Closed by #2302

PR #2302, opened 2024-10-24
Allow `--diff` to be used with `--jobs`

mnakama:
Fix the bug reported in https://github.com/PyCQA/isort/issues/2216 where `--diff` is ignored when `--jobs` is non-zero.

All credit for this fix goes to @ian-h-chamberlain, who reported the bug and found where the fix needed to be applied.


DanielNoord (approved):
LGTM!

Up to @staticdev whether a test for this is needed, but it seems like a nice fix to get in.

staticdev (approved):

staticdev:
Thanks @mnakama and @DanielNoord

Issue #2305, opened 2024-11-22
Python 3.13 support


mayty:
Python 3.13 was [released](https://www.python.org/downloads/release/python-3130/) on Oct. 7, 2024 and is not supported as `py_version` as of now:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "D:\WinFolders\Documents\PyCharmProjects\asemu\.venv\Scripts\isort.exe\__main__.py", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "D:\WinFolders\Documents\PyCharmProjects\asemu\.venv\Lib\site-packages\isort\main.py", line 1123, in main
    config = Config(**config_dict)
  File "D:\WinFolders\Documents\PyCharmProjects\asemu\.venv\Lib\site-packages\isort\settings.py", line 531, in __init__
    super().__init__(sources=tuple(sources), **combined_config)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 105, in __init__
  File "D:\WinFolders\Documents\PyCharmProjects\asemu\.venv\Lib\site-packages\isort\settings.py", line 264, in __post_init__
    raise ValueError(
    ...<3 lines>...
    )
ValueError: The python version 313 is not supported. You can set a python version with the -py or --python-version flag. The following versions are supported: ('all', '2', '27', '3', '310', '311', '312', '36', '37', '38', '39')
```

DanielNoord:
Fixed with https://github.com/PyCQA/isort/pull/2306

PR #2306, opened 2024-11-22
Python 3.13 support

mayty:
Related issue: [2305](https://github.com/PyCQA/isort/issues/2305)


king-phyte:
Since 3.8 is EOL, shouldn't we drop support for that too?

mayty:
I think that Python 3.8 doesn't conflict with adding Python 3.13 support and dropping it is outside of this PR's scope

king-phyte:
Fair point :)

isort/stdlibs/py312.py
```diff
@@ -14,7 +14,9 @@
     "argparse",
     "array",
     "ast",
+    "asynchat",
```
hugovk, line 17: Please remove these.

They were all removed in 3.12:

https://docs.python.org/3/whatsnew/3.12.html#asynchat-and-asyncore

isort/stdlibs/py313.py
```diff
@@ -0,0 +1,224 @@
...
+stdlib = {
+    "_ast",
+    "_thread",
+    "_tkinter",
+    "abc",
+    "aifc",
+    "argparse",
+    "array",
+    "ast",
+    "asynchat",
```
hugovk, line 17: Similarly, this list contains lots of modules that have been removed in 3.12 or earlier.

Was this list created using the mkstdlibs.py script?

hugovk (changes requested):

isort/stdlibs/py313.py
mayty: all changes to `stdlibbs/py*.py` were generated by `mkstdlibs.py` scripts

isort/stdlibs/py312.py
mayty: I can remove it, but it was added by the scripts. So it will automatically appear on the next run because it's still present in the corresponding documentation: https://docs.python.org/3.12/library/asynchat.html#module-asynchat

Mb it would be better to use [stdlibs](https://github.com/omnilib/stdlibs) package or adopt it's approach to modules generation?

hugovk: Aha, the problem was caused by the docs re-adding module pages which had bee removed, but as stubs:

https://github.com/python/cpython/pull/126622

> Mb it would be better to use [stdlibs](https://github.com/omnilib/stdlibs?rgh-link-date=2024-12-27T18%3A09%3A16Z) package or adopt it's approach to modules generation?

Yeah, sounds like a good idea to use stdlibs 👍

mayty: updated the script to use `stdlibs` and ran it

hugovk (approved):
Thank you!

DanielNoord:
Awesome! Could you rebase this so we can get this merged?

mayty:
I had to drop python3.8 support to make builds pass on python3.13 because some packages required a version, which do not support python3.8 (not sure how it passed before the `make actions happy` PR merge)

At the same time, the prod package could have been installed with both 3.8 and 3.13, so I'm open to rolling back this commit if 3.8 support is considered more important than builds with python3.13 in PRs

hugovk:
I think it's fine to drop 3.8, as mentioned before, it's already EOL:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0569/

We could merge this as-is and do a more thorough 3.8 cleanup in another PR.

DanielNoord:
See https://github.com/PyCQA/meta/issues/64#issuecomment-2585135841 I have stopped merging anything.

staticdev (changes requested):
@mayty thanks a lot for your contribution. Your changes look good to me. I kindly ask to rebase again and generate a new poetry.lock. The integration tests were failing and I just merged https://github.com/PyCQA/isort/pull/2329.

Also, since you are supporting a higher version on Python, I also suggest to increase the integration tests version up to 3.12 on `.github/workflows/integration.yml`.

poetry.lock
```diff
@@ -1455,135 +1458,131 @@ files = [
 
 [[package]]
 name = "pydantic"
-version = "2.5.2"
+version = "2.10.5"
```
DanielNoord, line 1461: Note that this PR does a bit more than just bump 3.13, but also bumps some dependencies in the `poetry.lock` file.

I'm not sure if that is intentional.

pyproject.toml
```diff
@@ -80,6 +81,7 @@ types-setuptools = ">=70.0.0.20240523"
 types-colorama = ">=0.4.2"
 types-toml = ">=0.1.3"
 vulture = ">=1.0"
+stdlibs = ">=2024.10.21.16"
```
DanielNoord, line 84: It seems as if `[tool.poetry.dev-dependencies]` is almost in alphabetical order. Not sure if we want to add a commit to do so.

DanielNoord (commented):
@staticdev Pushed the changes you requested to help move this along :)


There are two more remaining questions from my side. Other than that I think also think this looks good.

poetry.lock
mayty: It is intentional, previous versions couldn't be installed on 3.13

pyproject.toml
mayty: I'm all for alphabetical order, but only if there is gonna be a commitment to maintaining it in the future (at least a mention in the [coding-standart](https://github.com/PyCQA/isort/blob/main/docs/contributing/2.-coding-standard.md) doc)

staticdev: +1 for alhabetical order.. I normally have in all project I maintain.

mayty: Sorted dev dependencies

DanielNoord (approved):
Awesome!

staticdev:
Thanks @mayty @DanielNoord @hugovk

PR #2322, opened 2025-01-09
Eliminate `actions/cache@v2` usage

kurtmckee:
[GitHub is shutting down this old action version on February 1st, 2025](https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/).

In addition, its usage here is no longer needed, because new versions of `actions/setup-python` can do this caching.

@DanielNoord As noted above, this is a time-sensitive PR because `actions/cache@v2` will be removed from GitHub soon.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -16,20 +16,13 @@ jobs:
         python-version: ["3.11"]
 
     steps:
-      - uses: actions/checkout@v2
-
-      - name: pip cache
-        uses: actions/cache@v2
-        with:
-          path: ~/.cache/pip
-          key: integration-pip-${{ hashFiles('**/pyproject.toml') }}
-          restore-keys: |
-            integration-pip-
+      - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -16,20 +16,13 @@ jobs:
...
-          restore-keys: |
-            integration-pip-
+      - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: "pip"
```
DanielNoord, line 25: Shouldn't this be `poetry`?

kurtmckee: [The docs for the `actions/setup-python` caching.](https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies)

DanielNoord (approved):
Thanks, didn't notice that we were not caching poetry previously.

PR #2325, opened 2025-01-09
Dependency updates round 1

matthewhughes934:
First round of upstreaming some improvements, mostly around
dependencies, from https://github.com/matthewhughes934/isort-fork

- Remove some `flake8` exceptions

    Mostly as a matter of consistency. For `isort/__init__.py` this
    exception looks to not be needed since
    be0fbd07817abcd5d2e62db9233b6633715d3405

    An exception is kept in pace for an example file in tests as the
    contents of the file are just for testing.

- Restore `setuptools` as optional dependency

    This looks to have been accidentally removed with
    f7a6b0eea57e87155a367e2490b49b40f83c3944, `setuptools` is still required
    in `isort.setuptools_commands`

- Drop some unused dependencies

    * `ipython` was removed with 49b41a5112ca3e8305dc4836e871ad9ab7f66792
    * `smmap2` looking at the history with `git log -G smmap --name-only
      --oneline` suggests this was only ever defined in
      `poetry.lock`/`pyproject.toml`. Given this is a library and not
      providing something extra like e.g. a `flake8` plugin it should be
      safe to remove


DanielNoord:
@matthewhughes934 This can be rebased on `master` again :) Sorry for the churn! Just taking things slow.

DanielNoord (approved):
Awesome, thanks!

Issue #2307, opened 2024-11-26
Drop Python 3.8 support due to EOL


eirnym:
Python 3.8 has EOL on 2024-10-07. I propose to drop support of it for next isort version.

kurtmckee:
The project hasn't seen updates for a while. I recommend closing this to reduce the backlog of issues to be addressed when the maintainers have time. Thanks!

eirnym:
@kurtmckee I don't see any value of closing issues for that reason.

eirnym:
Btw, some issues, such as updating pre-commit names and as well as few others, require this issue to be fulfilled first

PR #2327, opened 2025-01-09
Remove support for Python 3.8

DanielNoord:
Closes https://github.com/PyCQA/isort/issues/2307

Deciding to take it a bit slower and merge stuff in smaller chunks.

@matthewhughes934 I cherry picked your change from https://github.com/PyCQA/isort/pull/2325 as without it `poetry lock --no-update` doesn't work.
I need some more time to look into the other changes, which I hope to do tomorrow after merging this.

@Mayty You mentioned that you couldn't get stuff to work without removing Python 3.8 in https://github.com/PyCQA/isort/pull/2306/ so let's first merge that and worry about Python 3.13 afterwards. That seems like a better procedure :)

Perhaps any of you or @hugovk could have a look at this? :)

Note that I'm taking it slow as some concern was raised in https://github.com/PyCQA/meta/issues/64 that my efforts to merge some PRs might not be welcomed by the original maintainers of the library/package. Therefore I want to make extra sure I don't merge any bugs/issues and only merge stuff that is definitely okay.


DanielNoord:
> Perhaps `sys.version_info.minor >= 12` should also be `sys.version_info.minor >= 13` now? I didn't look up the intent of this check.

I did, and I think we can just use `py_version = f"{sys.version_info.major}{sys.version_info.minor}"`. That is the exact same result. Pushed a change that does this, as well as setting the target version.

> Plus we could also run pyupgrade:
> 
> ```shell
> pip install pyupgrade
> pyupgrade **/**.py --py39-plus
> ```
> 
> And this will upgrade a lot of typing things, like using `dict` and `tuple` directly instead of `typing.Dict` and `typing.Tuple`.

I'd prefer to keep that out of scope for now. Running `pyupgrade` is not necessary to unblock all the other PRs and then we can do one big effort to update all linters :)

hugovk (approved):

PR #2329, opened 2025-01-10
Updates round 2

matthewhughes934:
Second round of updates, split out to keep PRs reasonable size

- Drop `py` dependency

    This package is in maintenance only mode and only had two uses. Replace
    their use with the `tmp_path` fixture as suggested in the docs[1]

    Link: https://docs.pytest.org/en/8.3.x/reference/reference.html#tmpdir [1]

- Update `pytest`

    There was one new error:

        TypeError: exceptions must be derived from Warning, not <class 'NoneType'>

    Fix this following the docs[1]

    Link: https://docs.pytest.org/en/8.3.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests [1]

- Update `black`

- Update `flake8`

    This required fixes an error in the config (I also updated the relevant
    link, the old one would 404):

        ValueError: Error code '#' supplied to 'extend-ignore' option does not match '^[A-Z]{1,3}[0-9]{0,3}$'

    And some errors in the code:

        isort/literal.py:59:8: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
        tests/unit/test_isort.py:13:1: F401 'typing.Dict' imported but unused

    The second test conflicted with `mypy`, since `Dict` was only used in
    commented annotations, to make them both happy move one of these
    annotations out of a comment (in the future we should probably move all
    such annotations out of comments)

- Update `hypothesis`

    This required addressing some errors from `mypy`:

        tests/integration/test_setting_combinations.py:20: error: "SearchStrategy" expects no type arguments, but 1 given  [type-arg]
        tests/integration/test_hypothesmith.py:36: error: "SearchStrategy" expects no type arguments, but 1 given  [type-arg]


tests/unit/test_main.py
```diff
@@ -1178,39 +1184,40 @@ def main_check(args):
...
+        gitignore.write_text(gitignore.read_text() + "ignore_link.py\nignore_link")
 
         out, error = main_check(
             [str(git_project0), "--skip-gitignore", "--filter-files", "--check"]
         )
 
         should_check = ["/git_project0/has_imports.py"]
 
-        assert all(f"{str(tmpdir)}{file}" in error for file in should_check)
+        assert all(f"{str(tmp_path)}{file}" in error for file in should_check)
```
DanielNoord, line 1216: Using both `str` and `f"` shouldn't be necessary for `Path`

tests/unit/test_isort.py
```diff
@@ -1629,13 +1629,13 @@ def test_multiline_import() -> None:
     assert isort.code(test_input) == ("from pkg import more_stuff, other_suff, stuff\n")
 
     # test again with a custom configuration
-    custom_configuration = {
+    custom_configuration: Dict[str, Any] = {
```
DanielNoord, line 1632: Can you use `dict[str, Any]`?

DanielNoord (commented):
The changes due to `py` should probably be refactored a bit, see my comments.

This also supersedes https://github.com/PyCQA/isort/pull/2235/ and closes https://github.com/PyCQA/isort/pull/2235/

tests/unit/test_isort.py
matthewhughes934: > Can you use `dict[str, Any]`?

yep, now that we don't support Python 3.8 that will work

tests/unit/test_main.py
matthewhughes934: > Using both `str` and `f"` shouldn't be necessary for `Path`

what do you think about making this a separate change: I see there are several points in this file (including parts not touched by this change) that could do with the same update? I'm happy to include it in this patch, but my main goal was to just get things working, without changing things too much

DanielNoord: Is it possible to exclude the `py` change and do that together with the changes in a separate PR?

If not, let's at least make sure the code we're touching now in this diff is correct and worry about the rest later.

matthewhughes934: > If not, let's at least make sure the code we're touching now in this diff is correct and worry about the rest later.

:+1: [830eca7](https://github.com/PyCQA/isort/pull/2329/commits/830eca7567e136d146e5a1e8dd33d81f1459628e)

tests/unit/test_isort.py
matthewhughes934: done with [cbd07e0](https://github.com/PyCQA/isort/pull/2329/commits/cbd07e010ebb94b37306e26d7d21cf85c2c05ae8)

matthewhughes934:
I've pushed some fixup commits, so let me know just before merge and I'll rebase to squash them back in to the correct commits

matthewhughes934:
Lint is unhappy

> tests/unit/test_isort.py:13:1: F401 'typing.Dict' imported but unused

becuase that file uses type comments like `type: Dict[str, Any]`, so that needs to be imported when `TYPE_CHECKING`, or that file could be updated to replace those comments with type annotations (preferable, but happy to have that as a separate change)

DanielNoord:
See https://github.com/PyCQA/meta/issues/64#issuecomment-2585135841 I have stopped merging anything.

staticdev (approved):
LGTM, thanks a lot @matthewhughes934.

PR #2331, opened 2025-01-12
Add OSError handling in find_imports_in_file

### LINKED ISSUE #2330: isort-identify-imports fails if there are unreadable files in the directory
Steps to reproduce:

1. mkdir new_directory
2. cd new_directory
3. ln -s not_exist bad_link.py
4. isort-identify-imports .

isort-identify-imports fails with FileNotFoundError.  The same problem occurs when the files is unreadable due to permission issues.

In the case of isort, it is not an error but UserWarning.  So I believe isort-identify-imports should do the same.

Emacs creates a lock file with a name like `.#some_file.py` as a symbolic link to a nonexistent target when there are unsaved changes in the buffer.  As a result, running isort-identify-imports with an unsaved buffer in Emacs will result in an error.

kobarity:
This PR fixes #2330.

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -588,15 +588,18 @@ def find_imports_in_file(
     - **top_only**: If True, only return imports that occur before the first function or class.
     - ****config_kwargs**: Any config modifications.
     """
-    with io.File.read(filename) as source_file:
-        yield from find_imports_in_stream(
-            input_stream=source_file.stream,
-            config=config,
-            file_path=file_path or source_file.path,
-            unique=unique,
-            top_only=top_only,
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

staticdev (approved):
Thanks @kobarity and @DanielNoord

DanielNoord:
@staticdev You're not going to believe it, but I think this broke the tests main again...

`.mksymlinkto` is not available on Windows.

kobarity:
Sorry, I did not test on Windows.
How about #2369?

PR #2339, opened 2025-01-17
Fix regressions for black

### LINKED ISSUE #2338: Fix regressions from #2236
New behavior of black breaks old tests.

staticdev:
Closes #2338

### DIFF
--- a/tests/unit/profiles/test_black.py
+++ b/tests/unit/profiles/test_black.py
@@ -447,8 +447,8 @@ def sub(a: np.ndarray, b: np.ndarray) -> np.ndarray: ...
 
 def test_black_trailing_comma():
     black_test(
-    "from x import (a, b, c,)\n",
-    """\
+        "from x import (a, b, c,)\n",
+        """\
 from x import (
     a,
     b,
--- a/tests/unit/test_regressions.py
+++ b/tests/unit/test_regressions.py
@@ -408,6 +408,7 @@ from many_stop_words import (
 """,
         show_diff=True,
         profile="black",
+        split_on_trailing_comma=False,
     )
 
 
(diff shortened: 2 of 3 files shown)


matthewhughes934:
The change from https://github.com/PyCQA/isort/pull/2236 looks to not play nice when using `with` imports, here's a patch that adds a test that will fail when parsing the output:

```patch
diff --git a/tests/unit/profiles/test_black.py b/tests/unit/profiles/test_black.py
index 8f750100..027ebce4 100644
--- a/tests/unit/profiles/test_black.py
+++ b/tests/unit/profiles/test_black.py
@@ -456,3 +456,15 @@ def test_black_trailing_comma():
 )
 """,
     )
+
+def test_black_trailing_comma2():
+    black_test(
+        """\
+from x import ( a as b,)
+""",
+        """\
+from x import (
+    a as b,
+)
+""",
+    )
```

matthewhughes934:
maybe there's an existing bug with `split_on_trailing_comma`, here's a test that reproduces the issue on 6550a74d362cf3f27f531bab22c784fb8f86181a (picking a random commit that was from a few months ago)

```patch
diff --git a/tests/unit/test_isort.py b/tests/unit/test_isort.py
index ddfa4e24..9d237128 100644
--- a/tests/unit/test_isort.py
+++ b/tests/unit/test_isort.py
@@ -5559,11 +5559,9 @@ def seekable(self):
 
 
 def test_split_on_trailing_comma() -> None:
-    test_input = "from lib import (a, b, c,)"
+    test_input = "from lib import (a as b,)"
     expected_output = """from lib import (
-    a,
-    b,
-    c,
+    a as b,
 )
 """
 
```

DanielNoord:
Created a different PR in https://github.com/PyCQA/isort/pull/2340

I think that is the actual "fix" for the configuration option itself, rather than "fixing" the black profile.

staticdev:
Superseeded by https://github.com/PyCQA/isort/pull/2340

PR #2340, opened 2025-01-18
Ensure that split_on_trailing_comma works with `as` imports

DanielNoord:
Closes https://github.com/PyCQA/isort/issues/2338

Took the tests as recommended in https://github.com/PyCQA/isort/pull/2339

I don't think we should fix the tests by disabling the new profile flag, that would mean that our tests work but anybody using the profile would still get broken imports.
Instead, let's try to fix `config.split_on_trailing_comma` itself. I believe this does this.

### DIFF
--- a/tests/unit/test_isort.py
+++ b/tests/unit/test_isort.py
@@ -5587,6 +5587,7 @@ def test_split_on_trailing_comma_wih_as() -> None:
     output = isort.code(expected_output, split_on_trailing_comma=True)
     assert output == expected_output
 
+
 def test_infinite_loop_in_unmatched_parenthesis() -> None:
     test_input = "from os import ("
 
--- a/tests/unit/test_isort.py
+++ b/tests/unit/test_isort.py
@@ -5576,6 +5576,17 @@ def test_split_on_trailing_comma() -> None:
     assert output == expected_output
 
 
+def test_split_on_trailing_comma_wih_as() -> None:
+    test_input = "from lib import (a as b,)"
+    expected_output = """from lib import a as b
+"""
+
+    output = isort.code(test_input, split_on_trailing_comma=True)
+    assert output == expected_output
+
+    output = isort.code(expected_output, split_on_trailing_comma=True)
+    assert output == expected_output
+
 def test_infinite_loop_in_unmatched_parenthesis() -> None:
     test_input = "from os import ("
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_isort.py
```diff
@@ -5576,6 +5576,18 @@ def test_split_on_trailing_comma() -> None:
     assert output == expected_output
 
 
+def test_split_on_trailing_comma_wih_as() -> None:
+    test_input = "from lib import (a as b,)"
+    expected_output = """from lib import a as b
+"""
```
matthewhughes934, line 5582: what's expected for multiple `as` imports an a trailing comma, e.g. `from lib import (a as b, c as d,)?

```python
from lib import a as b
from lib import c as d
```

or

```python
from lib import (
    a as b
    c as d
)
```

DanielNoord: Current code will do:
```python
from lib import a as b
from lib import c as d
```

staticdev:
@DanielNoord this Django integration test still failimg, did you have time to check if it is related or not to these black changes?

DanielNoord:
Like I told you in the message: by changing the black style django is no longer conformant. The tests will fail until we release and they reformat their repository.

staticdev:
@DanielNoord ok, then we can temporarily disable it and put a comment to reenable when they upgrade to 6.0.0+

staticdev (changes requested):
@DanielNoord ok, then we can temporarily disable it and put a comment to re-enable when they upgrade to 6.0.0+

DanielNoord:
@staticdev There we go.

I double checked and the offending line is:
https://github.com/django/django/blob/ec8cbea91ed6362d43298caa43f3d868452f7efa/django/db/backends/postgresql/compiler.py#L6

`django` does indeed use the `black` profile:
https://github.com/django/django/blob/ec8cbea91ed6362d43298caa43f3d868452f7efa/pyproject.toml#L61
_Which surprised me considering there is a `django` profile as well..._

I think this is fine to merge now :)

staticdev (approved):
@DanielNoord LGTM, thanks.

nessita:
Thank you for considering Django in this change. As you can see in the linked [PR #19109](https://github.com/django/django/pull/19109) this is forcing a reformat which does not seem correct?
```diff
- from django.db.models.sql.compiler import SQLUpdateCompiler
+ from django.db.models.sql.compiler import (
+     SQLUpdateCompiler,
+ )
```

DanielNoord:
That was raised as an issue in #2352.

Current status is: we're still determining whether this is a bug and are trying to see if we can provide a fix.

PR #2341, opened 2025-01-18
Bump `poetry` to `2.0.1`

DanielNoord:
Bump poetry and fix the deprecation warning about `dev-dependencies`.


staticdev (changes requested):
@DanielNoord I would say the best solution is replcing poetry by uv. It is much faster.. I didnit yesterday with https://github.com/staticdev/human-readable/pull/712

What do you think?

DanielNoord:
I don't have a lot of trouble with `poetry` being slow in this repository.

This seems like an easy fix, replacing with a different tool seems to be more work and there are still lots of PRs to go through

staticdev (approved):
Ok, we can merge it, actually it is a small effort to move to UV also in the future.

Issue #2337, opened 2025-01-17
Replace Poetry with UV


staticdev:
UV is faster and adheres to latest PEPs.

PR #2349, opened 2025-01-26
UV replacement of Poetry

staticdev:
Reverts PyCQA/isort#2347
Closes https://github.com/PyCQA/isort/issues/2337


pyproject.toml
```diff
@@ -132,8 +138,13 @@ logo = "art/logo.png"
 palette = {scheme = "isort"}
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[dependency-groups]
+dev = [
```
DanielNoord, line 145: What is the difference between this and the `dev` optional dependencies?

.github/workflows/integration.yml
```diff
@@ -24,18 +24,11 @@ jobs:
...
-        run: |
-          pip install --constraint=.github/workflows/constraints.txt pip
-          pip --version
-
-      - name: Install Poetry
-        run: |
-          pipx install --pip-args=--constraint=.github/workflows/poetry-constraints.txt poetry
-          poetry --version
+      - name: Install UV
+        uses: astral-sh/setup-uv@v5
```
DanielNoord, line 28: Should we set the version of `uv` to use like we did for `poetry`?

The documentation has [pointers](https://github.com/astral-sh/setup-uv) on how to do this

.github/workflows/lint.yml
```diff
@@ -24,18 +24,11 @@ jobs:
...
-        run: |
-          pip install --constraint=.github/workflows/constraints.txt pip
-          pip --version
-
-      - name: Install Poetry
-        run: |
-          pip install --constraint=.github/workflows/poetry-constraints.txt poetry
-          poetry --version
+      - name: Install UV
+        uses: astral-sh/setup-uv@v5
```
DanielNoord, line 28: See above

.github/workflows/release-dev.yml
```diff
@@ -0,0 +1,30 @@
+---
+name: Release Dev
```
DanielNoord, line 2: What is the intent of this workflow?

.github/workflows/test.yml
```diff
@@ -29,18 +29,11 @@ jobs:
...
-      - name: Upgrade pip
-        run: |
-          pip install --constraint=.github/workflows/constraints.txt pip
-          pip --version
-
-      - name: Install Poetry
-        run: |
-          pip install --constraint=.github/workflows/poetry-constraints.txt poetry
-          poetry --version
+      - name: Install UV
```
DanielNoord, line 32: See above

isort/setuptools_commands.py
```diff
@@ -1,7 +1,7 @@
 import glob
 import os
 import sys
-from typing import Any, Dict, Iterator, List
+from typing import Any, Iterator
```
DanielNoord, line 4: These changes seem unrelated? Why are we removing `user_options` ?

.github/workflows/release-dev.yml
barrelful: This one always creates a build, it can further be extended to send to TestPyPI like https://github.com/staticdev/human-readable/blob/main/.github/workflows/release-dev.yml.

staticdev: This one always will build the package for release for all merged to see if it builds. Also we can further extend to send to Test PyPI like in https://github.com/staticdev/human-readable/blob/main/.github/workflows/release-dev.yml#L32

.github/workflows/integration.yml
staticdev: Good point, it is a good idea. Thanks!

isort/setuptools_commands.py
staticdev: A bit, but for some reason mypy started complaining about it. It may be it happened because mypy got upgraded on the way.

.github/workflows/integration.yml
staticdev: Done.

pyproject.toml
staticdev: I did some mess in the conversion, but fixed.
[dependency-groups] are the correct part for dev dependencies, the optional dependencies are the extras.

isort/setuptools_commands.py
DanielNoord: Should we at least make a small mention of it in the changelog?

I'm not too up to speed with what this does, but I can assume users can be surprised this is removed without a notice.

DanielNoord (commented):
Only one comment about the removal. Rest LGTM!

isort/setuptools_commands.py
staticdev: @DanielNoord I actually returned with user_options to be extra safe and put a type ignore with a comment so mypy does not complain.

isort/setuptools_commands.py
```diff
@@ -16,7 +16,8 @@ class ISortCommand(setuptools.Command):
     """
 
     description = "Run isort on modules registered in setuptools"
-    user_options: List[Any] = []
+    # Potentially unused variable - check if can be safely removed
+    user_options: list[Any] = [] # type: ignore
```
DanielNoord, line 20:
```suggestion
    user_options: list[Any] = [] # type: ignore[error-code]
```

Nit, but I always add the error code that I'm ignoring so that I don't accidentally ignore a new error code after a `mypy` bump. Could you add the specific code?

staticdev: Added.

DanielNoord (approved):
Awesome! Nice job on doing this migration :rocket:

staticdev:
> Awesome! Nice job on doing this migration 🚀

Thanks a lot, nice job on review also!

.github/workflows/lint.yml
```diff
@@ -24,18 +24,13 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
```
DanielNoord, line 25: By the way, I think these caches are now outdated.

Maybe something for a follow up though!

PR #2353, opened 2025-01-29
Apply some ruff rules

cclauss:
Some optimizations for performance or readability.
% `ruff check --select=C4,E713,FLY,PERF401,PIE810,PLC0414,RUF005,RUF010,RUF021,SIM118,SIM201,UP004,UP033,W`
```
15	RUF005 	[*] collection-literal-concatenation
10	RUF010 	[*] explicit-f-string-type-conversion
 7	RUF021 	[*] parenthesize-chained-operators
 4	E713   	[*] not-in-test
 3	PIE810 	[*] multiple-starts-ends-with
 3	PERF401	[ ] manual-list-comprehension
 2	SIM118 	[*] in-dict-keys
 1	C413   	[*] unnecessary-call-around-sorted
 1	C417   	[*] unnecessary-map
 1	SIM201 	[*] negate-equal-op
 1	FLY002 	[*] static-join-to-f-string
 1	PLC0414	[*] useless-import-alias
 1	UP004  	[*] useless-object-inheritance
 1	UP033  	[*] lru-cache-with-maxsize-none
[*] fixable with `ruff check --fix`
```
https://docs.astral.sh/ruff/rules
---
I do not understand DeepSource's complaints that two functions have increased cyclomatic complexity when both ruff and flake8 agree that the cyclomatic complexity of these functions is unchanged.

% `ruff check --select=C901 ` # Results are the same on `main` and this branch.
```
isort/identify.py:43:5: C901 `imports` is too complex (33 > 10)
isort/settings.py:827:5: C901 `_get_config_data` is too complex (27 > 10)
```

PR #2359, opened 2025-02-10
Lint Python code with ruff

cclauss:
As suggested at https://github.com/PyCQA/isort/pull/2353#pullrequestreview-2583241874 let's add the [`ruff` linter](https://docs.astral.sh/ruff).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,7 @@ dev = [
     "pytest-benchmark>=3.4.1",
     "pytest-mock>=1.10",
     "requirementslib>=1.5",
+    "ruff>=0.9.6",
     "safety>=2.2.0",
     "stdlibs>=2024.10.21.16",
     "toml>=0.10.2",
@@ -168,6 +169,37 @@ module = "tests.*"
 allow_untyped_defs = true
 allow_incomplete_defs = true
 allow_untyped_calls = true
+
+[tool.ruff]
+line-length = 100
+lint.select = [
+    "ASYNC",
+    "B",
+    "C4",
+    "C90",
+    "E",
+    "F",
+    "FLY",
+    "PIE",
+    "S",
+    "W",
+]
+lint.ignore = [
+    "B017",
+    "B028",
+    "B904",
+    "E203",
+    "E501",
+]
+lint.exclude = [ "_vendored" ]
+lint.mccabe.max-complexity = 91  # Default is 10
+
+[tool.ruff.lint.per-file-ignores]
+"isort/hooks.py" = [ "S603" ]
+"isort/settings.py" = [ "S603", "S607" ]
+"tests/*" = [ "S" ]
+"tests/unit/example_crlf_file.py" = [ "F401" ]
+"tests/unit/test_wrap_modes.py" = [ "PIE804" ]
 
 [tool.isort]
 profile = "hug"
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -153,6 +154,37 @@ allow_untyped_defs = true
...
+    "S",
+    "W",
+]
+lint.ignore = [
+    "B017",
+    "B028",
+    "B904",
+    "E203",
+    "E501",
+]
```
DanielNoord, line 177: Some of these seem pretty useful. Why did you decide to ignore them instead of fixing them?

pyproject.toml
```diff
@@ -153,6 +154,37 @@ allow_untyped_defs = true
...
+lint.exclude = [ "_vendored" ]
+lint.mccabe.max-complexity = 91  # Default is 10
+
+[tool.ruff.lint.per-file-ignores]
+"isort/hooks.py" = [ "S603" ]
+"isort/settings.py" = [ "S603", "S607" ]
+"tests/*" = [ "S" ]
+"tests/unit/example_crlf_file.py" = [ "F401" ]
+"tests/unit/test_wrap_modes.py" = [ "PIE804" ]
```
DanielNoord, line 186: Similar question as above, why are we ignoring these here?

DanielNoord (commented):
Thanks! One question about the ignores

pyproject.toml
cclauss: Together those rules would require 65 fixes on the `main` branch even with `line-length=100`.  Modifying a dozen-plus files slows the review process.

My sense in the face of resistance to the tool was to land the tool in a pull request that did [not create too much change](https://github.com/PyCQA/isort/pull/2353#issuecomment-2637794904) and was therefore easy to review and merge before implementing more of the 800+ rules available.

pyproject.toml
cclauss: Similar position.  Let's land the tool first before we tighten up the config.

cclauss:
The fact that `DeepSource: Python` disagrees with both flake8 and ruff what code complexity does not help the review process.

pyproject.toml
staticdev: That's a lot.

staticdev:
@cclauss @DanielNoord to be honest, I am not very familiar with DeepSource, it was there since forever. Not sure it is a modern solution anymore. Does any of you have opinions on it already?

pyproject.toml
cclauss: The majority of rules are not for this codebase.

cclauss:
I'm not a fan of DeepSource for Python (I find it opaque) but I hear it is quite useful for other languages.

DanielNoord:
I would also be okay with removing it!

DanielNoord (approved):
I would probably open an issue for the follow up on the ignores, but other than that LGTM!

cclauss:
Failing GitHub Actions tests fixed in:
* #2363 or
* #2364

staticdev:
@cclauss after the other merges, this one needs a manual rebase.

DanielNoord (approved):
Still LGTM!

Thanks @cclauss

cclauss:
What up with the Windows tests?

DanielNoord:
A broken test added in https://github.com/PyCQA/isort/pull/2331

You're not having a lot of luck with this PR, sorry!

I'll leave it to @staticdev to decide how they want to resolve this.

Issue #2356, opened 2025-02-03
Remove .flake8 config


staticdev:
It should be added to pyproject.toml.

DanielNoord:
That's not possible right? `flake8` doesn't support `pyproject.toml`?

staticdev:
Ha, you are right.. maybe the solution is really replacing it with ruff-lint as mentioned in https://github.com/PyCQA/isort/pull/2353

Andrioden:
With Flake8-pyproject you can: https://github.com/john-hen/Flake8-pyproject

PR #2360, opened 2025-02-11
Move flake8 settings into pyproject.toml

cclauss:
Move `flake8` linter settings from the file `.flake8` into `pyproject.toml` by using [`Flake8-pyproject`](https://pypi.org/project/Flake8-pyproject).

% `uv run --with=Flake8-pyproject flake8 isort/ tests/`

Fixes: #2356 
* #2356

Related to:
* #2359


DanielNoord:
I think @staticdev pointed in https://github.com/PyCQA/isort/issues/2356#issuecomment-2631928376 towards just replacing `flake8` so that we would also no longer need this.

cclauss:
Some code imports flake8 code…
* https://github.com/search?q=repo%3APyCQA%2Fisort%20flake8.options

DanielNoord:
That's just a string in a test case. The test itself might be dependent on `flake8` being in the environment but we can 1) fix that and 2) even if we didn't we wouldn't need to config for `flake8` anymore.

cclauss:
* #2356 is not about removing flake8 so this pull request is merely about moving flake8 config to with PEP 621.

Several flake8 plugins require stable interactions between flake8 and isort.
https://github.com/orgs/wemake-services/repositories?q=flake8 has three flake8 plugins that rely on `isort.profile = wecare` and there are other flake8 plugins that depend on sorting imports like:
* https://pypi.org/project/flake8-isort
* https://pypi.org/project/flake8-imports

I am unsure what I am breaking by removing flake8 so none of [my pull requests](https://github.com/PyCQA/isort/pulls/cclauss) have attempted that but if someone else wants to that would be great.

pyproject.toml
```diff
@@ -141,6 +142,22 @@ dev = [
     "vulture>=1.0",
 ]
 
+[tool.flake8]
+max-complexity = 91
```
staticdev, line 146: @cclauss just curious why you added this max-complexity?
This one was not in original file.

staticdev:
> I think @staticdev pointed in [#2356 (comment)](https://github.com/PyCQA/isort/issues/2356#issuecomment-2631928376) towards just replacing `flake8` so that we would also no longer need this.

That would be indeed another way, but since @cclauss already put the effort. I think it is valid and better to maintain the settings in a single file.

pyproject.toml
cclauss: This function is 9 times more complex than the suggested code complexity setting.  This is also what DeepSource is complaining about but they have a different calculation than flake8 and ruff.

It is useful to set upper limits on complexity so that if someone makes this code even more complex (i.e. less understandable and more difficult to maintain) then the changing of this setting should set off a discussion in the code review about taking time to refactor or even add more tests.

staticdev: Yes, not the best software engineering example of logic. But I don't see value adding a super high complexity to it.

I would prefer to set a normal recommended value and ignore this method with a TODO to reduce the complexity and remove the ignore.

cclauss:
```suggestion
```
Let’s remove it here and I will do as you say in a separate code complexity PR using ruff after that lands.

staticdev (approved):
Thanks @cclauss LGTM.

cclauss:
Failing tests fixed in:
* #2363 or
* #2364

PR #2363, opened 2025-02-15
GitHub Actions: Use uv cache, not pip cache

cclauss:
Fix the failing GitHub Actions tests by using uv cache, instead of pip cache.
* https://docs.astral.sh/uv/guides/integration/github/#caching

A subset of:
* #2364 which also upgrades to uv v0.6.0 which contains breaking changes.


DanielNoord:
Let's just merge #2364 instead :)

Thanks for giving us an option here!

PR #2371, opened 2025-02-20
Ruff rules B017, B028, and E203

cclauss:
As requested at https://github.com/PyCQA/isort/pull/2359#discussion_r1952337350

Stop ignoring ruff rules
* B017 [`assert-raises-exception`](https://docs.astral.sh/ruff/rules/assert-raises-exception)
* B028 [`no-explicit-stacklevel`](https://docs.astral.sh/ruff/rules/no-explicit-stacklevel)
* E203 [`whitespace-before-punctuation`](https://docs.astral.sh/ruff/rules/whitespace-before-punctuation)

`ruff rule E203 `# is disabled by default in `ruff` (but not `flake8`) because it conflicts with `black` and `ruff format`.

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -199,7 +199,8 @@ def sort_stream(
                 raise ExistingSyntaxErrors(content_source)
             if config.verbose:
                 warn(
-                    f"{content_source} Python AST errors found but ignored due to Cython extension"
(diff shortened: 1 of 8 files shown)


DanielNoord (approved):
Thanks @cclauss

staticdev (approved):
Excellebt @cclauss! Thanks

PR #2378, opened 2025-03-04
Add tox for cross-platform, parallel test suite execution

### LINKED ISSUE #2351: Infrastructure Proposal: use tox or nox for test executions
This is a feature request / proposal for changes I would like to make, if the project is open to them.

## What

Add `noxfile.py`, `tox.ini`, or other config (e.g., `tox.toml`) to the project, and update documentation to instruct contributors to use `nox -s test` or `tox run` to execute tests.
Update CI to also run tests via the designated workflow tool.

## Why

As a new contributor looking to get involved in the project, it's not clear exactly how to run the tests. I can dig information out of CI configs and look at the test scripts, but it's non-obvious what the contributor setup should be.
(I did find that `uv sync --all-extras; ./scripts/test.sh` works)

Furthermore, if we want to test the code under multiple conditions (different interpreters, different extras installed), a matrix-style runner is needed.

## Additional Background

Because I'm a long-time `tox` user, I would be most inclined to tackle this by setting up a `tox.ini`.
Tox and nox are almost entirely interchangeable, and nox has significant usage as well (e.g., `pip`'s own tests use it).

Both tools also support use of `uv` as a speedup, e.g. via the `tox-uv` plugin.

My overall goal in proposing this is twofold:
1. Make local development of `isort` easier and more pleasant
2. Allow these environment managers to be used to script and automate non-testsuite tasks -- e.g. a new `tox run -e benchmark` suite to profile performance

---

Okay, that's my proposal. Would this be interesting to the project? I'd start small with a config which only handles the testing.

kurtmckee:
This change introduces tox and tox-uv for testing.

These changes allow for cross-platform testing by running `tox`, parallel execution across all supported Python versions by running `tox run-parallel` (or simply `tox p`), and cross-interpreter coverage reporting and HTML report output.

Closes #2351


staticdev:
Nice @kurtmckee, tox-uv is new to me. Let's give it a try.. having tox will make things more organized anyway.

pyproject.toml
```diff
@@ -141,6 +142,8 @@ dev = [
     "ruff>=0.9.6",
     "stdlibs>=2024.10.21.16",
     "toml>=0.10.2",
+    "tox>=4.24.1",
+    "tox-uv",
```
staticdev, line 146: Why didn't you also fix version of tox-uv?

kurtmckee: I'm not seeing an error about unresolvable interdependencies anymore, so I've rebased on `main` and fixed `tox-uv` to the latest version.

staticdev (approved):
Nice @kurtmckee, LGTM.

PR #2379, opened 2025-03-04
Resolve bandit warnings

kurtmckee:
The `nosec` flags benefit from human-readable explanation, but bandit was seeing the trailing text and throwing warnings:

```
WARNING Test in comment: Needed is not a test name or id, ignoring
WARNING Test in comment: for is not a test name or id, ignoring
WARNING Test in comment: gitignore is not a test name or id, ignoring
WARNING Test in comment: support is not a test name or id, ignoring
```

This is addressed by using another `#` character to hide the text.

### DIFF
--- a/isort/hooks.py
+++ b/isort/hooks.py
@@ -5,7 +5,7 @@ usage:
 """
 
 import os
-import subprocess  # nosec - Needed for hook
+import subprocess  # nosec # Needed for hook
 from pathlib import Path
 from typing import List, Optional
 
@@ -18,7 +18,7 @@ def get_output(command: List[str]) -> str:
     :param str command: the command to run
     :returns: the stdout output of the command
     """
-    result = subprocess.run(command, stdout=subprocess.PIPE, check=True)  # nosec - trusted input
+    result = subprocess.run(command, stdout=subprocess.PIPE, check=True)  # nosec # trusted input
     return result.stdout.decode()
 
 
--- a/isort/hooks.py
+++ b/isort/hooks.py
@@ -5,7 +5,7 @@ usage:
 """
 
 import os
-import subprocess  # nosec # Needed for hook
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

isort/hooks.py
```diff
@@ -5,7 +5,7 @@
 """
 
 import os
-import subprocess  # nosec - Needed for hook
+import subprocess  # nosec # Needed for hook
```
staticdev, line 8: This solves the warning but creates an awkward comment inside comment.. I would suggest instead to put one comment above.

```suggestion
# Needed for hook
import subprocess  # nosec
```

kurtmckee: That made the test suite fail so I'm going to revert that change.

staticdev: @kurtmckee then let's remove those # Need for hook and also # Needed for gitignore support, they are not that useful.

staticdev (changes requested):

isort/hooks.py
```diff
@@ -18,7 +18,7 @@ def get_output(command: List[str]) -> str:
     :param str command: the command to run
     :returns: the stdout output of the command
     """
-    result = subprocess.run(command, stdout=subprocess.PIPE, check=True)  # nosec - trusted input
+    result = subprocess.run(command, stdout=subprocess.PIPE, check=True)  # nosec # trusted input
```
staticdev, line 21: This one I believe the linter will not complain if you put `# Trusted input` in the line before.

staticdev (approved):
LGTM @kurtmckee, thanks.

kurtmckee:
Friendly ping. :grin:

staticdev:
@kurtmckee forgot to click merge haha

PR #2381, opened 2025-03-05
Fix bugs in `normalize_line()`, and with a 2.7x speedup

kurtmckee:
The following script was used for performance testing:

```
import timeit
import isort.parse

print(
    timeit.timeit(
        'isort.parse.normalize_line("from..cimport a")',
        number=5_000_000,
        globals={"isort": isort},
    )
)
```

Fastest time with code in `main`: 15.98 seconds
Fastest time with code in this PR: 5.82 seconds

### DIFF
--- a/isort/parse.py
+++ b/isort/parse.py
@@ -38,8 +38,8 @@ def _infer_line_separator(contents: str) -> str:
     return "\n"
 
 
-_from_absolute_normalizer_pattern = re.compile(r"from[\t ]+(.+?)[\t ]+(c?import)[\t ]+")
-_from_relative_normalizer_pattern = re.compile(r"from[\t ]*(\.+[^\t ]*?)[\t ]*(c?import)[\t ]+")
+_FROM_ABSOLUTE_NORMALIZER_PATTERN = re.compile(r"from[\t ]+(.+?)[\t ]+(c?import)[\t ]+")
+_FROM_RELATIVE_NORMALIZER_PATTERN = re.compile(r"from[\t ]*(\.+[^\t ]*?)[\t ]*(c?import)[\t ]+")
 
 
 def normalize_line(raw_line: str) -> Tuple[str, str]:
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):

kurtmckee:
Thanks for the review, @DanielNoord! Is there anything preventing this PR (and the others you approved) from merging?

DanielNoord:
Mostly a review from @staticdev I think

isort/parse.py
```diff
@@ -38,17 +38,18 @@ def _infer_line_separator(contents: str) -> str:
     return "\n"
 
 
+_from_absolute_normalizer_pattern = re.compile(r"from[\t ]+(.+?)[\t ]+(c?import)[\t ]+")
+_from_relative_normalizer_pattern = re.compile(r"from[\t ]*(\.+[^\t ]*?)[\t ]*(c?import)[\t ]+")
```
staticdev, line 42: I would just move this before _infer_line_separator and call them with uppercase and starting without _ to be standard global variable. Eg. `FROM_ABSOLUTE_NORMALIZER_PATTERN = ...`.

staticdev (changes requested):

isort/parse.py
kurtmckee: * I put these here because they're used by `normalize_line()` -- `_infer_line_separator` is not.
* I chose the names to reflect that these are for use in the module, not for importing and using outside of the module.

If these reasons aren't convincing, I can move these away from their usage site and rename them. :+1:

staticdev: @kurtmckee since those are not being imported in other files, it is ok to keep `_` in the beginning of the name, but I would at least keep them uppercase since we follow mostly Google Style Guide -> https://google.github.io/styleguide/pyguide.html#316-naming

It is a bit personal preference for organization, I always put all global variables before all methods and classes, but it does not hurt also to keep there.

kurtmckee: Nice, I appreciate the info re: Google Style Guide usage. Thanks!

Issue #2386, opened 2025-03-13
Feature Request: Display Project-Related Links in PyPI Sidebar


guillermodotn:
## Description

It would be useful if `PyPI` could display project-related links in the sidebar for easy access and reference. This would provide a fast and standardized way for users to find important resources such as source code and documentation directly from the project page.

## Ex. Project Links

- **Homepage**: Visit the official project homepage.
- **Documentation**: Find the full documentation here.
- **Repository**: Access the source code on GitHub.
- **Changelog**: View the project’s release notes and version history.

PR #2387, opened 2025-03-13
Add Project URLs to PyPI Side Panel

guillermodotn:
Ensures that useful project links appear in the sidebar on `PyPI`. These links help users easily find documentation, source code, releases, and other important resources.

Resolves Issue #2386.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,12 +31,6 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
     "Topic :: Utilities",
 ]
-urls = { 
-    Homepage = "https://pycqa.github.io/isort/index.html",
-    Documentation = "https://pycqa.github.io/isort/index.html",
-    Repository = "https://github.com/PyCQA/isort",
-    Changelog = "https://github.com/PyCQA/isort/releases"
-}
 include = [
     { path = "tests", format = "sdist" },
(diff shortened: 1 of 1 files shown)


DanielNoord:
Shouldn't we use trusted publishing so that most of these are taken from GitHub instead?

guillermodotn:
With PyPI **trusted publishing**, only **repository metrics** (GitHub Statistics) will be pulled directly from GitHub. In this case, the `Repository` URL becomes unnecessary since it would be redundant.  

Below is an example from the PyPI [ruff](https://pypi.org/project/ruff) project:

image  

The **Project Links** section will **not** appear unless you explicitly specify the URLs. The only difference when using **trusted publishing** is that the links will be classified into two groups: **verified** and **unverified**, based on the criteria outlined here:  
🔗 [PyPI Project Metadata Documentation](https://docs.pypi.org/project_metadata/)

DanielNoord (approved):
Seems good to me then!

guillermodotn:
I have modified the `pyproject.toml` file to follow the [PEP 621](https://peps.python.org/pep-0621/) guidelines by using the `[project.urls]` sub-table.

DanielNoord (approved):
@staticdev Mind if I merge this?

staticdev (approved):

staticdev:
No prob @DanielNoord but i sved you 2 clicks =)

PR #2405, opened 2025-09-01
Drop use of non-standard pkg_resources API

### LINKED ISSUE #2404: isort not compatible with Python 3.14
In Python 3.14 `pkg_resource` is not available, resulting in:

```
ERROR: Unrecoverable exception thrown when parsing ./test-notifies.py! This should NEVER happen.
If encountered, please open an issue: https://github.com/PyCQA/isort/issues/new
  File "/usr/local/bin/isort", line 8, in <module>
    sys.exit(main())
             ~~~~^^
...
  File "/usr/local/lib/python3.14/site-packages/isort/settings.py", line 718, in sorting_function
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

The correct API to find the entry points is described in the [python packaging guide](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata) as "from importlib.metadata import entry_points".

dvarrazzo:
Use the canonical `importlib.metadata.entry_points` instead.

Fix #2404

### DIFF
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -45,12 +45,12 @@ from .utils import Trie
 from .wrap_modes import WrapModes
 from .wrap_modes import from_string as wrap_mode_from_string
 
-if sys.version_info < (3, 10):  # pragma: no cover
-    from importlib_metadata import entry_points
-else:
-    from importlib.metadata import entry_points
-
 if TYPE_CHECKING:
+    if sys.version_info < (3, 10):  # pragma: no cover
+        EntryPoints = Any
+    else:
+        from importlib.metadata import EntryPoints
+
     tomllib: Any
 else:
     if sys.version_info >= (3, 11):
@@ -935,6 +935,19 @@ def _as_bool(value: str) -> bool:
         return _STR_BOOLEAN_MAPPING[value.lower()]
     except KeyError:
         raise ValueError(f"invalid truth value {value}")
+
+
+def entry_points(group: str) -> "EntryPoints":
+    """Call entry_point after lazy loading it.
+
+    TODO: The reason for lazy loading here are unknown.
+    """
+    if sys.version_info < (3, 10):  # pragma: no cover
+        from importlib_metadata import entry_points
+    else:
+        from importlib.metadata import entry_points
+
+    return entry_points(group=group)
 
 
 DEFAULT_CONFIG = Config()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,6 +180,10 @@ allow_untyped_defs = true
 allow_incomplete_defs = true
 allow_untyped_calls = true
 
+[[tool.mypy.overrides]]
+module = "importlib_metadata.*"
+ignore_missing_imports = true
+
 [tool.ruff]
 line-length = 100
 lint.select = [


pyproject.toml
```diff
@@ -38,7 +38,9 @@ include = [
 ]
 requires-python = ">=3.9.0"
 
-dependencies = []
+dependencies = [
+    "importlib_metadata; python_version < '3.10'",
```
DanielNoord, line 42: Should we add a lower bound on the version that is needed?

dvarrazzo: - The package [`importlib.medatata`](https://docs.python.org/3.10/library/importlib.metadata.html) is in the stdlib since Python 3.10.
- The package [`importlib_metadata`](https://importlib-metadata.readthedocs.io/) is a backport for older Python version.

The latter cover all the versions that your package supports, so I don't think you need a lower bound.

isort/settings.py
```diff
@@ -45,6 +45,11 @@
 from .wrap_modes import WrapModes
 from .wrap_modes import from_string as wrap_mode_from_string
 
+if sys.version_info < (3, 10):
```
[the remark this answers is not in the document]
dvarrazzo: Well, when the code was importing from pkg_resources, which is a non-standard module teleported in Python virtualenv thanks to obscure forces, I can somewhat understand why the import was lazy: if it suddenly started failing (like now, in Python 3.14) I assume the thought was _at least it won't fail for everyone on import but just for people who use this less used feature_.

Now that the code is importing either a stdlib function or a function from a package marked as dependency, I don't see a good reason for failing.

AFAICS the package was first used in 2d76984f1345763f31aff8c1fc940000700cd46e, then some code following a similar pattern was added in b40e584d0742a52fc061f4dd55a4e8469d19c34f, then again in 4cb72fa93fdcd78df00687de7cd91428ab0fdc93. I assume the code was copied over and modified, including the lazy-loading pattern. Maybe @timothycrosley might explain why those import are lazy?

Anyways, the last commit in this MR restores the lazy import behaviour. I don't think it is needed, especially if you have unit tests testing that the package works, but if you want it this way, it's there.

staticdev (changes requested):
I would actually remove support for python below 3.10., as 3.9 is pretty much eol: https://endoflife.date/python
I think we can keep the lazy loading like it is right now.

DanielNoord:
Seems like a lot more work to do proper clean up of 3.9? I'd say if this work we could at least just merge and release this and unblock `3.14` compatibility.

Afterwards we can worry about cleaning up for `3.9`?

rafaelcapucho:
I installed it using `uv`, as in `uv tool install isort`,

and when using with `.isort.cfg`, if I add `profile = "black"` it fails:
```
[settings]
profile = "black"
multi_line_output = 3
lines_between_types = 1
force_alphabetical_sort_within_sections = 1
Lexicographical = 1
group_by_package = 1
```

```
$ isort hdash/project/tasks.py

  File "/Users/kepler/.local/bin/isort", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/kepler/.local/share/uv/tools/isort/lib/python3.13/site-packages/isort/main.py", line 1124, in main
  File "/Users/kepler/.local/share/uv/tools/isort/lib/python3.13/site-packages/isort/settings.py", line 359, in __init__
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

DanielNoord:
@staticdev shall we just merge and release as is?

dvarrazzo:
I have rebased this branch on master and fixed the lint job.

staticdev (approved):
That is fine, LGTM. Thanks @dvarrazzo and @DanielNoord

Issue #2291, opened 2024-09-13
Long imports and noqa


matan1008:
Hola,
I have an import from the following format:
```
from aaaaaaa import bbbbbbbbbbbbbbbbb, ccccccccccccccccccc, dddddddddddddddd  # noqa: E402
from aaaaaaa import eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, fffffffffffffffffffff  # noqa: E402
```

When running `isort . -m HANGING_INDENT -l 120 --check-only --combine-as --df` it prints:
```
-from aaaaaaa import bbbbbbbbbbbbbbbbb, ccccccccccccccccccc, dddddddddddddddd  # noqa: E402
-from aaaaaaa import eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, fffffffffffffffffffff  # noqa: E402
+from aaaaaaa import bbbbbbbbbbbbbbbbb, ccccccccccccccccccc, dddddddddddddddd, eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, \
+    fffffffffffffffffffff  # noqa: E402
```

Running `isort . -m HANGING_INDENT -l 120 --check-only --combine-as --df` again, it prints:
```
-from aaaaaaa import bbbbbbbbbbbbbbbbb, ccccccccccccccccccc, dddddddddddddddd, eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, \
-    fffffffffffffffffffff  # noqa: E402
+from aaaaaaa import fffffffffffffffffffff  # noqa: E402
+from aaaaaaa import bbbbbbbbbbbbbbbbb, ccccccccccccccccccc, dddddddddddddddd, eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
```

Besides of the strange behavior  that yields two different formattings, the second one missing the noqa.

kurtmckee:
I'm not able to reproduce locally using isort 5.13.2.

I recommend closing this issue. Thanks!

kurtmckee:
@DanielNoord This issue is not reproducible. I recommend closing this issue.

matan1008:
Can I help reproducing it? It happens to me on two different ubuntu computers

![Image](https://github.com/user-attachments/assets/f573c9fc-d2a5-4f0f-ba79-a31d7fe3f5e0)

kurtmckee:
I see in your diff output that the file you're testing is larger than just the two lines that you.

[...]

PR #2407, opened 2025-09-11
Fix output of hanging indent for long lines with noqa

matan1008:


matan1008:
Fixes #2291

DanielNoord (commented):
Also seems like we need to reformat some files

DanielNoord (approved):
Thanks!

@staticdev Can we merge this?

PR #2416, opened 2025-10-02
Show absolute paths in skipped file messages

### LINKED ISSUE #2412: --verbose does not show full path to skipped file
```
$ isort -v . | fgrep -v -- -type

Fixing .../src/mypkg/schemas.py
Fixing .../src/mypkg/common.py
.git was skipped as it's listed in 'skip' setting, matches a glob in 'skip_glob' setting, or is in a .gitignore file with --skip-gitignore enabled.
.venv was skipped as it's listed in 'skip' setting, matches a glob in 'skip_glob' setting, or is in a .gitignore file with --skip-gitignore enabled.
Skipped 2 files
```

I would like to see the full path to the skipped file as it is the case with fixed files now.

pranlawate:
When files are skipped, isort now displays their absolute paths in verbose mode, consistent with the "Fixing" messages. This improves debugging by making it clear which files are being skipped.

Changes:
- isort/main.py:1178 - Convert file_name to absolute path before appending
- isort/files.py:24 - Use full_path instead of just dirname
- isort/files.py:35 - Use absolute filepath instead of just filename

Fixes #2412

### DIFF
--- a/isort/files.py
+++ b/isort/files.py
@@ -21,7 +21,7 @@ def find(
                     full_path = base_path / dirname
                     resolved_path = full_path.resolve()
                     if config.is_skipped(full_path):
-                        skipped.append(dirname)
+                        skipped.append(str(full_path))
                         dirnames.remove(dirname)
                     else:
                         if resolved_path in visited_dirs:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
Do you have a particular reason for using `os.path` and `Path`?

pranlawate:
Thanks for reviewing this PR. 
I am still fairly new with Python devlopement so I honestly don't know which is better to use.
 
I tried to use what's present in the nearby lines of where I changed the code.  In files.py line 34 uses os.path so I used the same on 35th and in main.py 1177 line uses Path so I used that on 1178. 

I generally like using Path myself so I can make both lines use that or if you would prefer os.path, then I can use that as well.
I think I can raise another PR for using either os.path or Path throughout the code by studying which is more suitable for isort to have more uniformity. Will prove a good improvement for me as well as the code if you would like that.

PR #2425, opened 2025-10-06
Test on `3.14` and fix any bugs

DanielNoord:
This makes sure we actually test on 3.14, which uncovers our support is quite lacking.
We need to remove `requirementslib` as it relies on a version of `pydantic` that does not support 3.14 yet.

We also need to remove the `pylama` extension.
`pylama` is unmaintained, see https://github.com/klen/pylama

As can be seen in this PR https://github.com/klen/pylama/pull/254 it looks like this won't be fixed soon.
Let's just remove


DanielNoord:
@staticdev Since I'm removing a feature here I think I'd like you to review this. It turns out we don't actually support 3.14 yet 😅

staticdev (approved):

staticdev:
That is perfect @DanielNoord. Thanks a lot, and actually it is an import missong pieces. Those libs are not critical amyway.

Issue #1959, opened 2022-08-02
Publish an official isort image on Docker Hub


ryboe:
Please publish an official image for `isort` on [Docker Hub](https://hub.docker.com/), [AWS Public ECR](https://gallery.ecr.aws/), [GCR](https://cloud.google.com/container-registry/), the GitHub Container Registry, or some other equivalent registry. This will make it easier to use `isort` in CI pipelines. I'm currently using a third-party [`kiwicom/isort`](https://hub.docker.com/r/kiwicom/isort) image, but it's out of date and inherently not as trustworthy as an official image.

For reference, [here's the official image for the `black` code formatter](https://hub.docker.com/r/pyfound/black).

Here's an example of how I use `kiwicom/isort` in a GitLab CI pipeline:

```yaml
Isort:
  image:
    name: kiwicom/isort:5.7.0
  script:
    - isort --version
    - isort --check-only app
```

daveisfera:
Anything we can do to get this done?

Bengt:
@daveisfera I would expect the ideal deployment process to be a GitHub action, that is integrated into the workflow of this project and listens for the tag to be increased for building and uploading the docker image to dockerhub.

I just ran into a long-fixed bug because my ci job was using an old image. The missing official image probably causes lots of unspoken pain for others, too.

DanielNoord:
A docker container for a single executable seems very overkill. Just use a package manager to do this?

Bengt:
Hi, @DanielNoord! Thanks for raising your concern. Yes, there is storage and also runtime overhead to packaging a (rather small) utility like isort as a Docker image. These are both small, though: A minimal alpine image including Python and isort should be about 15 - 20 MB in size, depending on the instruction set architecture. The execution overhead should be on the order of a few percent. Both should not be relevant for the intended use cases of conveniently running isort either locally or in the cloud, e.g. when running GitLab pipelines. For these use cases, there is an organization that packages often-used tools like isort, namely pipeline components: https://pipeline-components.dev/components/ Unfortunately, isort ist missing from that project, so I usually pick https://hub.docker.com/r/xcgd/isort, which is unfortunately outdated. Hence, the best way would be if the upstream isort project would provide a Docker image themselves.

DanielNoord:
Okay! Seems reasonable, but also something we're not very likely going to do. Maintainers are a bit sparse and I don't think we're likely to get into the weeds of setting up a docker hub account etc. etc. I'll keep this open though!

[...]

PR #2428, opened 2025-10-07
Feature/fix issue 1959 - Publish an official isort Docker image

Bengt:
This adds steps to the existing release workflow which package the isort Python distribution as a Docker image and push it to the GitHub container registry.

This fixes issue #1959.


DanielNoord:
Going to close this for now. Leaving it open might give the impression that we are still working on it, which we aren't.

Issue #2438, opened 2025-10-27
python -Wdefault -m pytest fails isort unittests


eli-schwartz:
https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/python-utils-r1.eclass#n1433

Gentoo distro packaging normalizes warning filters when running packaging tests. Trying to upgrade isort to 3.14, I get a weird failure. Minimized:

```
$ git clone https://github.com/PyCQA/isort
$ cd isort

$ python3.13 -m venv venv13
$ python3.14 -m venv venv14

$ ./venv13/bin/pip install . pytest
$ ./venv14/bin/pip install . pytest



$ ./venv13/bin/python -Wdefault -m pytest tests/unit/test_isort.py -k test_command_line
============================= test session starts ==============================
platform linux -- Python 3.13.5, pytest-8.4.2, pluggy-1.6.0
rootdir: /var/tmp/portage/dev-python/isort-7.0.0/work/isort
configfile: pyproject.toml
collected 227 items / 225 deselected / 2 selected

tests/unit/test_isort.py ..                                              [100%]

=============================== warnings summary ===============================
tests/unit/test_isort.py::test_command_line[True]
  /usr/lib/python3.13/multiprocessing/pool.py:268: ResourceWarning: unclosed running multiprocessing pool <multiprocessing.pool.Pool state=RUN pool_size=2>
    _warn(f"unclosed running multiprocessing pool {self!r}",
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================= 2 passed, 225 deselected, 1 warning in 0.10s =================



$ ./venv14/bin/python -Wdefault -m pytest tests/unit/test_isort.py -k test_command_line
============================= test session starts ==============================
platform linux -- Python 3.14.0rc1, pytest-8.4.2, pluggy-1.6.0
rootdir: /var/tmp/portage/dev-python/isort-7.0.0/work/isort
configfile: pyproject.toml
collected 227 items / 225 deselected / 2 selected

tests/unit/test_isort.py .F                                              [100%]

=================================== FAILURES ===================================
___________________________ test_command_line[True] ____________________________

tmpdir = local('/tmp/pytest-of-root/pytest-28/test_command_line_True_0')
[...]

PR #2443, opened 2025-11-05
Turn some warnings into errors in test suite

DanielNoord:
References https://github.com/PyCQA/isort/issues/2438

Inspired by that issue it would be good to at least get early notifications of `DeprecationWarnings`, `ResourceWarnings` and others.

Needs a little bit of work though.


DanielNoord:
Unfortunately too much work while we still have `pipreqs`.

PR #2446, opened 2025-11-07
feat!: remove old finders flag and legacy finder logic

### LINKED ISSUE #2445: Remove `--old-finders`
Because of `--old-finders` we need to support a lot of legacy code and dependencies. It also prevents us from following modern practices such as https://github.com/PyCQA/isort/pull/2443

@staticdev what do you think about removing the flag? It has been five years since it was introduced as a fallback option after the breaking changes in `isort` `5.0`. At this point I think it would be fine to do a breaking release where we remove it and clean up some of that old code :)

joao-faria-dev:
## Description
Hey everyone, removing `--old-finders` and, i believe, all related stuff!
It's a breaking change, but i believe i covered pretty much everything. Please let me know if i forgot anything :) 

## Related issue
Closes #2445

### DIFF
--- a/isort/parse.py
+++ b/isort/parse.py
@@ -141,6 +141,9 @@ class ParsedContent(NamedTuple):
     trailing_commas: set[str]
 
 
+# Ignore DeepSource cyclomatic complexity check for this function. It is one
+# the main entrypoints so sort of expected to be complex.
+# skipcq: PY-R1000
 def file_contents(contents: str, config: Config = DEFAULT_CONFIG) -> ParsedContent:
     """Parses a python file taking out and categorizing imports."""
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Changelog
 NOTE: isort follows the [semver](https://semver.org/) versioning standard.
 Find out more about isort's release policy [here](https://pycqa.github.io/isort/docs/major_releases/release_policy).
 
+### Unreleased
+
+  - Removed `--old-finders` flag and `old_finders` configuration option. The legacy finder logic that relied on environment introspection has been removed (#2445) @joao-faria-dev
```
DanielNoord: Can you also mention the `magic-placement` flag?

PR #2448, opened 2025-11-07
chore: replace black with ruff in clean.sh

### LINKED ISSUE #2447: Apply code formatting across codebase
## Description
Hello team! I executed the `./scripts/clean.sh` and it revealed that many files need lint updates. I'll be opening a PR for that due the amount of files affected

joao-faria-dev:
## Description
After running the `./scripts/done.sh`it reveled a lot of code formatting that should be fixed (+1321, -486). Then analyzing the workflow i noticed that we are actually using ruff. So instead of add the lint changes and i'm just going to update the clean script to use ruff 

## Related issue
- Close #2447

### DIFF
--- a/scripts/clean.sh
+++ b/scripts/clean.sh
@@ -3,5 +3,4 @@ set -euxo pipefail
 
 uv run isort --profile hug isort/ tests/ scripts/
 uv run isort --profile hug example_*/
-uv run ruff format isort/ tests/ scripts/
-uv run ruff format example_*/
+uv run ruff format


DanielNoord (commented):
It is kind of strange to me that the Lint job doesn't fail without this change. Is it possible that we also need to update that job so we also check everything in `scripts` for formatting?

joao-faria-dev:
Hey @DanielNoord, from my understanding the CI did not fail because the imports in scripts were already sorted and formatting was already covered by `ruff format --check` running on CI, but the isort step was not covering scripts. I made a small change to add that :)

DanielNoord (approved):

PR #2464, opened 2026-03-03
Add order_by_qualified_sort option

stephenfin:
Add support for [hacking](https://github.com/openstack/hacking)'s preferred sort order. Instead of

```python
from foo import baz
from foo.bar import wow
```

order by fully qualified paths:

```python
from foo.bar import wow
from foo import baz
```

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -803,6 +803,16 @@ Order imports by type, which is determined by case, in addition to alphabeticall
 - --ot
 - --order-by-type
 
+## Order By Qualified Name
+
+Order imports by their full module path. That is, `from foo.bar import wow` would be ordered before `from foo import baz` (since `foo.bar.wow` < `foo.baz`).
+
+**Type:** Bool  
+**Default:** `False`  
+**Config default:** `false`  
+**Python & Config File Name:** order_by_qualified_name  
+**CLI Flags:** **Not Supported**
+
 ## Atomic
 
 Ensures the output doesn't save if the resulting file contains syntax errors.
--- a/isort/output.py
+++ b/isort/output.py
@@ -63,10 +63,17 @@ def sorted_imports(
 
         from_modules = parsed.imports[section]["from"]
         if not config.only_sections:
+            _from_modules = from_modules
             from_modules = sorting.sort(
                 config,
                 from_modules,
-                key=lambda key: sorting.module_key(key, config, section_name=section),
+                key=lambda key: sorting.module_key(
+                    f"{key}.{min(_from_modules[key])}"
+                    if config.order_by_qualified_name and _from_modules[key]
+                    else key,
+                    config,
+                    section_name=section,
+                ),
                 reverse=config.reverse_sort,
             )
 
--- a/isort/settings.py
+++ b/isort/settings.py
@@ -173,6 +173,7 @@ class _Config:
     balanced_wrapping: bool = False
     use_parentheses: bool = False
     order_by_type: bool = True
+    order_by_qualified_name: bool = False
     atomic: bool = False
     lines_before_imports: int = -1
     lines_after_imports: int = -1
--- a/tests/unit/test_isort.py
+++ b/tests/unit/test_isort.py
@@ -1676,6 +1676,29 @@ def test_order_by_type() -> None:
     )
 
 
+def test_order_by_qualified_name() -> None:
+    # Without option: sorted by module name ("foo" < "foo.bar")
+    test_input = "from foo.bar import wow\nfrom foo import baz\n"
+    assert isort.code(test_input, order_by_qualified_name=False) == (
+        "from foo import baz\nfrom foo.bar import wow\n"
+    )
+    # With option: sorted by fully qualified path ("foo.bar.wow" < "foo.baz")
+    assert isort.code(test_input, order_by_qualified_name=True) == (
+        "from foo.bar import wow\nfrom foo import baz\n"
+    )
+    # Multiple imports from the same module use the first alphabetically
+    test_input = "from foo import zebra\nfrom foo.bar import wow\n"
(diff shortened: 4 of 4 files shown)


stephenfin:
> As maintainer, but mostly a Python community member I'm wondering: do we really need this?
> 
> We already have about 1237 competing standards for sorting imports. Do we really need another one to be added? It complicates tools like `isort` further with, I would argue, little benefit for end users.

I get where you're coming from, but this is a pretty widespread pattern that been enforced by [hacking](https://pypi.org/project/hacking) for [a very long time](https://github.com/openstack/hacking/commit/4e90f08824bc947b8fc968b19ff1a3729ee6a222). I also believe hacking was originally derived from [the Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#313-imports-formatting) which is even older. Considering just hacking alone, this formatting style covers all OpenStack projects plus countless more non-OpenStack projects using hacking (based on [this *very* crude sample set](https://github.com/search?q=%5Bhacking%5D+path%3Atox.ini+NOT+org%3Aopenstack&type=code)), so you've already got a lot of potential users right there. I'd really like us to be able to use a standard tool like `isort` than have to write our own, and given the base of potential users here coupled with the size of the functional changes needed (~10 lines, excluding tests and docs) I think we have a pretty compelling argument for doing so?

DanielNoord:
We already support the google style guide with a profile that is included by default:
https://github.com/PyCQA/isort/blob/a333737ed43df02b18e6c95477ea1b285b3de15a/isort/profiles.py#L25

`hacking` derives from Google, but has a crucial difference. To quote from Google:

> Within each grouping, imports should be sorted lexicographically, ignoring case, according to each module’s full package path (the path in from path import ...). Code may optionally place a blank line between import sections.

Their example is also very different from that of hacking:

```python
from book.genres import scifi
from myproject.backend import huxley
from myproject.backend.hgwells import time_machine
from myproject.backend.state_machine import main_loop
from otherproject.ai import body
from otherproject.ai import mind
from otherproject.ai import soul
```

vs.

```python
import nova.api.ec2
from nova.api import manager
from nova.api import openstack
from nova.auth import users
from nova.endpoint import cloud
import nova.flags
from nova.i18n import _, _LC
from nova import test
```

Notice that `from myproject.backend import huxley` comes before `from myproject.backend.hgwells import time_machine`.

It is clearly a very different style than the Google one, which is already supported.

Cost is not only in lines of code being committed. It is also the confusion (new) users can get from the plethora of configuration options, style guides and (in my opinion) other distractions they need to deal with when writing their Python code.

If the intent is to "Agree on a common style guide so reviews don't get bogged down on style nit picks." I'd urge you to standardize on one of the many style guides out there (or go back to the one you originally derived from) instead of introducing your own variant.

stephenfin:
The sample you've given is only a part of the sample and only includes "from" imports. This is the entire sample, which paints a very different picture:

```python
import collections
import queue
import sys

from absl import app
from absl import flags
import bs4
import cryptography
import tensorflow as tf

from book.genres import scifi
from myproject.backend import huxley
from myproject.backend.hgwells import time_machine
from myproject.backend.state_machine import main_loop
from otherproject.ai import body
from otherproject.ai import mind
from otherproject.ai import soul

# Older style code may have these imports down here instead:
#from myproject.backend.hgwells import time_machine
#from myproject.backend.state_machine import main_loop
```

There is also a sample at the very top of the section I originally linked to:

```python
from collections.abc import Mapping, Sequence
import os
import sys
from typing import Any, NewType
```

In both cases, you see "from" imports and "straight" imports mixed. This is the pattern that hacking has duplicated. I don't believe there's any difference?

Assuming I've got this right, then I suspect the Google profile provided by isort is not actually correct. But I would also suspect that changing this behavior now would likely have a large fallout, and a wiser option is to add a new option.

> If the intent is to "Agree on a common style guide so reviews don't get bogged down on style nit picks." I'd urge you to standardize on one of the many style guides out there (or go back to the one you originally derived from) instead of introducing your own variant.

We have standardized: it's called hacking and as I've already said, it's been around for a very long time (longer than isort has been on PyPI, [from the looks of things](https://pypi.org/project/isort/#history)), has been used within a very large Python-orientated community, and also has adoption outside of that community.

DanielNoord:
Turns out we already have an `open_stack` profile as well. That's my bad, I should have known about it. Perhaps you can help update it, because it might be out of date as well? It looks quite different from the Google profile.

Still, the main issue you are running into comes from `hackage` insisting on using the "full name" whereas all other styles guides only look at the `path` in `from path import ...`. This is why the new option is needed. Mixing "from" and "straight" imports is not the issue in this case.

I'm still reluctant to add the option as, in my opinion, it is confusing (due to dealing with terms that can have multiple interpretations such as "full" and "qualified") and is only needed for a single style guide.
I understand that is a frustrating response but I'm trying to balance the interest of one feature request against overall tool complexity. Im aware I might not be making the right balance here so I'll definitely leave this open for now to let others weigh in as well.

DanielNoord:
@stephenfin Did some more digging, and apparently there was an open pull request to "fix" this in `hacking`:
https://github.com/openstack/hacking/pull/7

That seems to change `hacking` to respect the code style as written in the Google style guide. Perhaps we can revive that PR?

DanielNoord:
Closing this for now.

I still believe the Google style guide is recommending a different pattern than the one you are looking for:
```python
from myproject.backend import huxley
from myproject.backend.hgwells import time_machine
```

To quote:
> Within each grouping, imports should be sorted lexicographically, ignoring case, according to each module’s full package path (the path in from path import ...).

It is of course fine for `hacking` to have its own style, but I don't think most users of `isort` would benefit from this option. Instead, they would likely be further confused by another configuration option.

Updates to the `open_stack` profile to bring it more in line with how it should be are of course more than welcome!

Issue #2462, opened 2026-02-25
Support explicit lazy imports in Python 3.15 (PEP 810)


hugovk:
[PEP 810 "Explicit lazy imports"](https://peps.python.org/pep-0810/) has been accepted and is coming to Python 3.15 (2026-03-10's 3.15.0a7 will be the first release).

This adds a new `lazy` keyword for imports:

```python
lazy import json
lazy from json import dumps
```

And also a `__lazy_modules__` for backwards compatibility with pre-3.15:

```python
__lazy_modules__ = ["json"]
import json
```

It would be useful to also sort `__lazy_modules__`. Two options:
* Alphabetical
* Follow the isort settings, for example: stdlib, third-party, local

The [Steering Council's acceptance](https://discuss.python.org/t/pep-810-explicit-lazy-imports/104131/466) said:

> We agree that the PEP should take no position on any style recommendations for sorting lazy imports. While we generally like the idea of grouping lazy imports together, let’s leave that up to the linters and auto-formatters to decide the details.

See also https://github.com/astral-sh/ruff/issues/21305.

hugovk:
Quick note: lazy imports is now available for testing in this week's 3.15 alpha 7, also available via GitHub Actions, and today's uv release.

https://discuss.python.org/t/python-3-15-0-alpha-7/106475

DanielNoord:
Support for sorting of import statements (so no `__lazy_modules__`) has been implemented in https://github.com/PyCQA/isort/pull/2503.
If somebody has feedback that is very welcome. I followed the same style as `ruff` does.

DanielNoord:
Support was released in `9.0.0a2` and announced on [DPO](https://discuss.python.org/t/isort-and-ruff-now-support-sorting-pep-810-lazy-imports/107075). We can adapt based on feedback :)

brettcannon:
In case anyone else is interested, based on the tests in #2503, it looks like the choice was to sort lazy imports to the end of the section they belong to, and alphabetical among lazy imports.

PR #2466, opened 2026-03-07
Add initial support for Python 3.15

DanielNoord:
Before we can consider https://github.com/PyCQA/isort/issues/2462 we first need to support 3.15 to begin with.

This:
- Bumps `stdlibs` to the latest version to get `py315` support
- Adds the `3.15` marker where needed in `tox` and CI files
- Ignores `tests/integration/test_hypothesmith.py` on 3.15 as `libcst` doesn't support it (yet)
- Unignores the `django` integration test
- Adds version specifiers to some imports that do not support `3.15` yet
- Adds the `3.15` marker to metadata
- Ignores an issue in `pygments`
- Updates the lock file

Quite a lot of changes, but all necessary to make 3.15 pass on CI in its fullest capacity.

### DIFF
--- a/tests/integration/test_projects_using_isort.py
+++ b/tests/integration/test_projects_using_isort.py
@@ -13,6 +13,8 @@ from collections.abc import Generator, Sequence
 from pathlib import Path
 from subprocess import check_call
 
+import pytest
+
 from isort.main import main
 
 


DanielNoord:
@staticdev Do you think you can give this as review?

staticdev:
Sure @DanielNoord i am on vacation but will be back in 9 days.

DanielNoord:
@staticdev I think adding support for lazy imports might be quite a lot of work. See the linked issue, as we'll also need to coordinate a bit with ruff and others.

Merging this will at least unblock this as we'll be able to start testing and parsing the lazy imports.

I think it would be good for the community if isort supports lazy imports from day 1.

PR #2517, opened 2026-04-15
Remove unused and broken dependencies

DanielNoord:
`scripts/check_acknowledgments.py` imports `hug`, which is no longer a dependency as it imports `cgi` and that has been removed from the `stdlib`.

The other dependencies are also unused I think so we might as well remove them? No reason to keep them around.


staticdev (approved):

staticdev:
Good finding.. I also think we can get rid.

PR #2518, opened 2026-04-17
Fix #2242: report lines_before_imports-only changes

### LINKED ISSUE #2242: isort.core.process returns False if only changes to lines before imports were made, with no changes to raw import section

`isort.core.process`:

```python
from a import b, x

foo = 'bar'
```
```python
from a import x, b

foo = 'bar'
```
___

```python
import isort

ln = '\n'

input_1 = ln + input_0

```

lawrence3699:
Fixes #2242

When `lines_before_imports` is the only requested change, `core.process()` can still return `False` because the buffered blank lines ahead of the import section never participate in the existing `raw_import_section` comparison. That makes `check_code()` return `True` and prevents file-based paths from applying the formatted output.

This change treats a mismatch between the buffered blank lines and the effective `lines_before_imports` count as a change, while keeping the existing `black`/`.pyi` special case. The regression test covers both adding and removing leading blank lines.

Validation:
- `uv run pytest tests/unit/test_regressions.py -q`
- `uv run pytest 'tests/unit/test_isort.py::test_custom_lines_before_import_section[True]' 'tests/unit/test_isort.py::test_custom_lines_before_import_section[False]' -q`
- `uv run pytest tests/unit -q -k 'not test_importable'`
- `uv run ruff check isort tests/unit`

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -371,6 +371,10 @@ def process(
                     continue
                 if not import_section:
                     output_stream.write("".join(lines_before))
+                made_changes = made_changes or bool(
+                    import_section
+                    and len(lines_before) != _configured_lines_before_imports(config, extension)
+                )
                 lines_before = []
 
             raw_import_section: str = import_section
@@ -515,3 +519,9 @@ def _has_changed(before: str, after: str, line_separator: str, ignore_whitespace
             != remove_whitespace(after, line_separator=line_separator).strip()
         )
     return before.strip() != after.strip()
+
+
+def _configured_lines_before_imports(config: Config, extension: str) -> int:
+    if config.profile == "black" and extension == "pyi":
+        return 1
+    return config.lines_before_imports
--- a/tests/unit/test_regressions.py
+++ b/tests/unit/test_regressions.py
@@ -1904,6 +1904,31 @@ def test_check_code_should_not_false_positive_with_float_to_top_and_add_imports(
     )
 
 
+@pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


isort/core.py
```diff
@@ -515,3 +519,9 @@ def _has_changed(before: str, after: str, line_separator: str, ignore_whitespace
             != remove_whitespace(after, line_separator=line_separator).strip()
         )
     return before.strip() != after.strip()
+
+
+def _configured_lines_before_imports(config: Config, extension: str) -> int:
```
DanielNoord, line 524: What is the existing special case you refer to?

tests/unit/test_regressions.py
```diff
@@ -1904,6 +1904,31 @@ def test_check_code_should_not_false_positive_with_float_to_top_and_add_imports(
...
 
 
+@pytest.mark.parametrize(
+    ("test_input", "lines_before_imports", "expected_output"),
+    [
+        ("from a import b, x\n\nfoo = 'bar'\n", 1, "\nfrom a import b, x\n\nfoo = 'bar'\n"),
+        ("\nfrom a import b, x\n\nfoo = 'bar'\n", 0, "from a import b, x\n\nfoo = 'bar'\n"),
+    ],
+)
+def test_lines_before_imports_changes_are_reported_issue_2242(
```
DanielNoord, line 1914: Let's use `isort.code` instead. We don't need to do the whole stream seeking

DanielNoord (commented):
Thanks for the PR!

PR #2531, opened 2026-05-07
Remove references to defunct `git_ignore` config

### LINKED ISSUE #1920: git_ignore documentation could be improved
I was checking [here](https://pycqa.github.io/isort/docs/configuration/options) if there's a way for isort to ignore the files from my `.gitignore` without requiring `git` installed (using it in a Jenkins pipeline with the `--check` flag). I stumbled upon `git_ignore` which looked promising and I had two issues with it: 

1. It didn't mention that it's only working when `skip_gitignore` is `True` which seems to be the case when you check the code
2. It mentions that it should be set to `True` but the actual value it expects is a `dict` so it's very confusing how you should be using it

I think the documentation can be improved to make the previous two points clearer.

And a little bit of an offtopic: You think a feature to ignore the contents of `.gitignore` using basic regex instead of relying of `git` to be present is a good idea? I know for CI that would be great cause you usually use slim images that don't have a lot of system dependencies installed. Should we create an issue for this as well?

sparrowt:
https://pycqa.github.io/isort/docs/configuration/options.html#git-ignore is nonsense and doesn't work - it raises `UnsupportedSettings` because e16806990ed862ac2bdc3125060d1a35741c1e83 removed it without removing the documentation that refers to it.

The remaining config option that still works is `skip_gitignore` https://pycqa.github.io/isort/docs/configuration/options.html#skip-gitignore.

Remove the defunct one to avoid confusion.

Fixes #1920

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -1629,16 +1629,6 @@ Forces star imports above others to avoid overriding directly imported variables
 
 - --star-first
 
-## Git Ignore
-
-If `True` isort will honor ignores within locally defined .git_ignore files.
-
-**Type:** Dict  
-**Default:** `{}`  
-**Config default:** `{}`  
-**Python & Config File Name:** git_ignore  
-**CLI Flags:** **Not Supported**
-
 ## Format Error
 
 Override the format used to print errors.
--- a/scripts/build_config_option_docs.py
+++ b/scripts/build_config_option_docs.py
@@ -128,7 +128,6 @@ description_mapping = {
     "auto_identify_namespace_packages": "Automatically determine local namespace packages, generally by lack of any src files before a src containing directory.",
     "namespace_packages": "Manually specify one or more namespace packages.",
     "follow_links": "If `True` isort will follow symbolic links when doing recursive sorting.",
-    "git_ignore": "If `True` isort will honor ignores within locally defined .git_ignore files.",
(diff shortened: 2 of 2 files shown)


Foryah:
It's true that `git_ignore` was removed but I think it's more accurate to say it was replaced with `git_ls_files` as we  can see in the commit. This PR solves half the problem, to solve #1920 we still need to document how `git_ls_files` works 😁

sparrowt:
I think the real question is whether `git_ls_files` should really be in `_Config` at all - AFAICS it's more of an implementation detail, [caching the result of calling `git ls-files`](https://github.com/PyCQA/isort/blob/ab9cd3601d92d59416a57f457dcabb27ff622dd3/isort/settings.py#L537-L559) (which is likely to be pretty big for most repos) and something which I can't imagine anyone wanting to pass in as a config.

So I think this PR probably does all of the documentation changes that we want related to this - perhaps a separate ticket is warranted for whether `git_ls_files` should be moved out of the config class seeing as that doesn't quite seem appropriate, but I'm new to this codebase so may have missed something & didn't want to go trying to move functional code around, to keep this "avoid confusion" PR simple.

Foryah:
Makes a lot of sense 😁 I never read the code other than the commit you passed and there I just saw that it was a replacement so I assumed is similar funtionality, different flag. But yeah, if the functionality is different, then all good 🥳

DanielNoord (approved):
Thanks!

Issue #2533, opened 2026-05-08
Outdated docs site on pycqa.github.io should send people to isort.readthedocs.io


sparrowt:
I spent a while rather confused why some docs were outdated - I ended up on https://pycqa.github.io/isort/ from a search engine and it looks legit and even says `8.0.1` in the top-right corner as if it's up to date, but... then I got to the changelog https://pycqa.github.io/isort/CHANGELOG.html and it doesn't have anything since 5.11.2 (in 2022) which is rather confusing.

Eventually I realised there's another up-to-date copy on readthedocs so there are 2 conflicting sites both still live:
- https://pycqa.github.io/isort/ ⚠️
- https://isort.readthedocs.io/en/latest/ ✅

I think it would help avoid a lot of confusion to do one of:
1. 💬 Add a banner to the top of all pages on the old site e.g.
`Note: the docs have moved, see https://isort.readthedocs.io/en/latest/ for the latest`
3. ➡️ or redirect from the old to the new (aiming to retain paths where possible)

For the redirect option:
- it seems that the following pattern would cover the majority of pages:
  - `https://pycqa.github.io/isort/docs/<PATH>` => `https://isort.readthedocs.io/en/latest/<PATH>`
  - e.g. https://pycqa.github.io/isort/docs/configuration/options.html => https://isort.readthedocs.io/en/latest/configuration/options.html
- there are a few things not under `docs` such as the home page & changelog (there may be more, I've not checked exhaustively):
  - https://pycqa.github.io/isort/ => https://isort.readthedocs.io/en/latest/
  - https://pycqa.github.io/isort/CHANGELOG.html => https://isort.readthedocs.io/en/latest/changelog.html

P.S. having searched the issues before raising this, I guess I'm one for the list of "reports of confused users" as mentioned here 😉 https://github.com/PyCQA/isort/issues/2417#issuecomment-4179850980
P.P.S. thanks for all your work on getting the docs publishing again among other things in this repo.

DanielNoord:
To be honest I don't really have a good idea about how to do this. If anybody does, please create a PR and I'll have a look 😄

sparrowt:
I guess the first question is, what's the setup behind https://pycqa.github.io/isort/ is it just pointing at Github Pages for this repo?

[...]

PR #2534, opened 2026-05-08
Old docs page should point to the new one.

sparrowt:
Possible improvement re https://github.com/PyCQA/isort/issues/2533, assuming that https://pycqa.github.io/isort/ is served up from the `gh-pages` branch (which this PR targets) as I suspect.

### DIFF
--- a/CHANGELOG.html
+++ b/CHANGELOG.html
@@ -2776,6 +2776,7 @@
                 
                 
                 <h1 id="changelog">Changelog</h1>
+                <h3>⚠️ Docs have moved, see <a href="https://isort.readthedocs.io/en/latest/">https://isort.readthedocs.io/en/latest/</a> for the latest</h3>
 <p>NOTE: isort follows the <a href="https://semver.org/">semver</a> versioning standard.
 Find out more about isort's release policy <a href="https://pycqa.github.io/isort/docs/major_releases/release_policy">here</a>.</p>
 <h3 id="5112-december-12-2022">5.11.2 December 12 2022</h3>
--- a/CHANGELOG.html
+++ b/CHANGELOG.html
@@ -2776,7 +2776,7 @@
                 
                 
                 <h1 id="changelog">Changelog</h1>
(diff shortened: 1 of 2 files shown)


DanielNoord (approved):
Can at least try this, thanks! :)

sparrowt:
It shows up on https://pycqa.github.io/isort/CHANGELOG.html 🎉

sparrowt:
And now https://pycqa.github.io/isort/index.html too 🎉 (I must've been getting a cached copy when I first checked)

sparrowt:
Now that we know this works, should we just replace all of the html pages with something like this?
```html
<!doctype html>
<html>
  <head>
    <meta http-equiv="refresh" content="5; URL=https://isort.readthedocs.io/en/latest">
    <style>
      p {
        font-family: sans-serif;
        font-size: large;
      }
    </style>
  </head>

  <body>
    <p>⚠️ isort docs have moved - redirecting to <a href="https://isort.readthedocs.io/en/latest/">https://isort.readthedocs.io/en/latest/</a> ...</p>
  </body>
</html>
```
which just shows this for 5 seconds before redirecting:
image

@DanielNoord I can make a PR for this against the `gh-pages` branch, but only if you're happy with it?

I could do a best effort job of redirecting to the corresponding new page where possible, or could just send them all to the home page as it's not rocket science to find what you want from the menu.

sparrowt:
Hm, though based on https://developers.google.com/search/docs/crawling-indexing/301-redirects#metarefresh we shouldn't have a delay and instead should just set it to 0 seconds so that Google treats it as a permanent redirect, which I think would be more useful from an SEO point of view (currently the old gh-pages docs are the first Google search result for "isort" sadly).

DanielNoord:
@sparrowt If you're up for that that would be appreciated!

Redirect to homepage is probably best :)

sparrowt:
https://github.com/PyCQA/isort/pull/2536 updates all pages to redirect.

It actually turned out to be pretty trivial to automate linking to the corresponding page on the new site, so I have put that up for your consideration - but if you strongly prefer the 'homepage in all cases' then I can simplify it to that (sorry, I know I asked and then just went and did the less-simple thing anyway... 🫣)

Given how much of the old site shows up in search engine results I think it could be worthwhile to have the specific redirects in place.

PR #2538, opened 2026-05-09
Fix --resolve-all-configs not using local setting for directly passed files

HYJ737:
## Summary

  Fixes an issue where files passed directly on the command line could ignore their local isort configuration when "--resolve-all-configs" was enabled.
  When "--resolve-all-configs" was used with its default "--config-root" of ".", and the target file lived outside that config root, the config trie returned the root fallback config as "default". That fallback empty config then replaced the file-specific config that had already been resolved from the file path. As a result, local settings such as extend_skip = ['a.py'] were not applied, and verbose output showed:
  - default used for file /tmp/isort-2335/a.py
  This PR preserves the local file configuration in that case and allows skip settings to take effect when sorting directly passed files with --resolve-all-configs.

## Changes

  - Added root_path tracking to the config Trie used by find_all_configs().
  - Updated config trie lookup so files outside the configured search root return no trie match instead of incorrectly matching the trie root
    fallback.
  - Updated check_file() and sort_file() to replace the active config only when the trie returns an actual config match.
  - Updated CLI sorting with --resolve-all-configs to respect skip settings for directly passed files.
  - Added a regression test covering a directly passed file whose containing directory defines:
 
## Verification

  Ran the focused regression test:
  uv run pytest tests/unit/test_main.py::test_resolve_all_configs_honors_skip_for_direct_file -q 
  Result:
  1 passed

  Ran related CLI tests:
  uv run pytest tests/unit/test_main.py -q
  Result:
  22 passed

  Ran the project unit test script:
  ./scripts/test.sh
  Result:
  571 passed, 1 skipped

  Ran the project clean/format script:
  ./scripts/clean.sh
  Result:
  110 files left unchanged

### DIFF
--- a/isort/api.py
+++ b/isort/api.py
@@ -327,10 +327,11 @@ def check_file(
         config_trie = config_kwargs.pop("config_trie", None)
         if config_trie:
             config_info = config_trie.search(filename)
-            if config.verbose:
+            if config_info[0] and config.verbose:
                 print(f"{config_info[0]} used for file {filename}")
 
-            file_config = Config(**config_info[1])
+            if config_info[0]:
+                file_config = Config(**config_info[1])
 
     with io.File.read(filename) as source_file:
         return check_stream(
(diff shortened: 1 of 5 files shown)


DanielNoord:
Do we have an issue for this?

PR #2554, opened 2026-06-01
Fix non-idempotent output with split_on_trailing_comma and a non-default wrap mode

sarathfrancis90:
## Summary

isort's output is not idempotent — `isort(isort(x)) != isort(x)` — when `split_on_trailing_comma` and `include_trailing_comma` are both enabled together with a non-default `multi_line_output` mode (e.g. `GRID`, `VERTICAL`, modes `0`/`1`/`4`/`5`, or the Black profile with an explicit wrap-mode override). The first pass wraps the import using the requested mode and appends a trailing comma; only the *second* pass collapses it to `VERTICAL_HANGING_INDENT`. Running isort twice should always reach a fixpoint, so this is a stability bug.

## Reproduction

```python
import isort

src = "from a import b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w\n"
cfg = dict(multi_line_output=0, split_on_trailing_comma=True,
           include_trailing_comma=True, line_length=40)

first  = isort.code(src, **cfg)
second = isort.code(first, **cfg)
assert first == second   # fails on current main
```

The same instability reproduces for `multi_line_output` modes `0`, `1`, `4` and `5`.

## Root cause

In `isort/output.py`, the "explode onto individual lines" branch was only triggered when the **input** already carried a trailing comma (`module in parsed.trailing_commas`):

```python
if (
    import_statement
    and config.split_on_trailing_comma
    and module in parsed.trailing_commas
    and not processed_as_imports_this_iteration
):
```

But when `include_trailing_comma` is set, isort itself *adds* a trailing comma to every import it wraps across multiple lines. So on the first pass the import is wrapped with the requested mode (no explode, because the source had no trailing comma yet) and gains a trailing comma; on the second pass that freshly-added comma now matches `parsed.trailing_commas` and the import is exploded — producing different output.

## Fix

Treat an import as if it already carried a trailing comma when `include_trailing_comma` is enabled and the statement is going to be wrapped across multiple lines, so the explode happens on the first pass too and the output reaches a fixpoint regardless of the requested `multi_line_output` mode:

```python
explode_on_trailing_comma = config.split_on_trailing_comma and (
    module in parsed.trailing_commas
    or (config.include_trailing_comma and do_multiline_reformat)
)
```

This is also consistent with the documented `split_on_trailing_comma` contract: an import that ends up with a trailing comma should be split onto separate lines.

## Behavior impact

- No behavior change when `split_on_trailing_comma` is `False`.
- No behavior change for imports that fit on a single line (they are never wrapped, so `do_multiline_reformat` stays `False`).
- No imports are added or dropped.
- The output simply stabilizes on the first pass instead of the second.

## Tests

- Added a regression test in `tests/unit/test_regressions.py` asserting idempotency for the affected configurations. It fails on `main` and passes with this change.
- Full unit suite: `572 passed, 1 skipped` (the single unrelated `test_importable` failure reproduces identically on pristine `main` in this environment and is not touched by this change).
- `ruff check`, `ruff format --check`, and `mypy` are clean for the changed files.

## Changelog

Added an entry under `### Unreleased` in `CHANGELOG.md`.

### DIFF
--- a/tests/unit/test_regressions.py
+++ b/tests/unit/test_regressions.py
@@ -2021,3 +2021,45 @@ def test_sort_reexports_with_stdin_raises_error_issue_2393():
     with pytest.raises(SystemExit) as exc_info:
         main(argv=["--sort-reexports", "-"], stdin=fake_stdin)
     assert exc_info.value.code != 0
+
+
+def test_split_on_trailing_comma_idempotent_with_non_default_wrap_mode():
+    """Ensure isort output is idempotent when ``split_on_trailing_comma`` and
+    ``include_trailing_comma`` are combined with a non-default ``multi_line_output`` mode.
+
+    With both options enabled, isort always appends a trailing comma when it wraps an
+    import across multiple lines.  ``split_on_trailing_comma`` then explodes any import
+    that ends with a trailing comma onto individual lines on the *next* run.  Previously
+    the explode was only applied when the *input* already carried a trailing comma, so the
+    first pass wrapped using the requested ``multi_line_output`` mode (e.g. VERTICAL or
+    GRID) and only the second pass collapsed the result to VERTICAL_HANGING_INDENT, making
+    the output unstable.
+    """
+    to_sort = "from a.b.c import (d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s)\n"
+
+    # Every multi-line wrap mode must reach a fixpoint after a single pass.
+    for multi_line_output in (0, 1, 2, 3, 4, 5):
+        first_pass = isort.code(
+            to_sort,
+            include_trailing_comma=True,
+            split_on_trailing_comma=True,
+            multi_line_output=multi_line_output,
+            line_length=40,
+        )
+        second_pass = isort.code(
+            first_pass,
+            include_trailing_comma=True,
+            split_on_trailing_comma=True,
+            multi_line_output=multi_line_output,
+            line_length=40,
+        )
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Find out more about isort's release policy [here](docs/major_releases/release_po
 
 ### Unreleased
 
+  - Fixed unstable output when `split_on_trailing_comma` and `include_trailing_comma` are combined with a non-default `multi_line_output` mode (#2554) @sarathfrancis90
```
DanielNoord: Could you remove this?

tests/unit/test_regressions.py
```diff
@@ -2012,3 +2012,45 @@ def test_comment_on_opening_line_of_aliased_import_does_not_move():
         isort.code(short_line, profile="black")
         == "from mod import attr as alias  # type: ignore[attr-defined]  # My comment\n"
     )
+
+
+def test_split_on_trailing_comma_idempotent_with_non_default_wrap_mode():
```
DanielNoord, line 2026: Can you rebase your PR?

sarathfrancis90:
Done — dropped the CHANGELOG entry and rebased onto main. The regression test now sits next to the new `--sort-reexports` one (that was the rebase conflict). Thanks for taking a look!

DanielNoord (approved):

DanielNoord:
Thanks!

PR #2559, opened 2026-06-15
Fix over-long wildcard imports being rewritten into invalid Python

### LINKED ISSUE #2267: Bug: Word wrapping works incorrectly with force_single_line=True
Word wrapping works incorrectly with `force_single_line=True` for `from ... import *`.

## Steps to Reproduce

```
; .isort.cfg

[settings]
py_version = 39
profile = black
force_single_line = true
```

```python
# example_file.py

from very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.long.line import *
```

After running isort the file is broken:
```python
# example_file.py

from very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.(
    line import *.long.very.very,
)
```

## Context (Environment)
- python 3.9
- isort 5.13.2

## Possible fixes

I didn't find a way to disable word wrapping.

I've tried setting `wrap_length`  and `line_length` to 999999 to avoid wrapping, but it starts conflicting with `black` (`isort` removes wrapping, then `black` puts it back).

Labib-Bin-Salam:
## What

Fixes #2267. An over-long wildcard import combined with `force_single_line` (and reproducible under the `black` profile) is currently rewritten into **invalid Python**:

```python
# input
from very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.long.line import *

# isort output (broken!)
from very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.very.(
    line import *.long.very.very,
)
```

The output no longer parses.

## Root cause

A wildcard import cannot be wrapped at all: parenthesising a star (`from x.y.z import (*)`) is a `SyntaxError`, and the only over-long part — the dotted module path — cannot be split either.

In `wrap.line()` the splitter loop tries `("import ", "cimport ", ".", "as ")`. For a star import the `import ` splitter's `\bimport \b` pattern fails to match, because there is no word boundary between the trailing space and `*` (both are non-word characters). isort then falls through to the `.` splitter and splits the **module path**, mangling the statement.

## Fix

Detect star imports in `wrap.line()` and return the statement untouched, which matches Black's behaviour (Black also leaves long `from x import *` lines as-is). The change is guarded to the wrapping branch, so `multi_line_output=NOQA` behaviour is unaffected.

## Verification

- New regression test `test_long_star_import_should_not_be_corrupted_issue_2267` covers `force_single_line`, the `black` profile, `combine_star`, a trailing-comment case, validity (`compile`) and idempotency.
- `ruff check`, `ruff format --check`, `flake8`, `mypy -p isort -p tests` and `isort --profile hug --check` all pass on the changed files.
- Full `tests/unit/` suite shows no new failures (the two pre-existing Windows/CRLF + import-ordering failures are present on `main` as well, unrelated to this change).

### DIFF
--- a/tests/unit/test_regressions.py
+++ b/tests/unit/test_regressions.py
@@ -2117,3 +2117,43 @@ def test_noqa_wrap_mode_does_not_accumulate_spaces_with_as_import():
         first_pass, multi_line_output=7, force_single_line=True, line_length=40
     )
     assert second_pass == first_pass
+
+
+def test_long_star_import_should_not_be_corrupted_issue_2267():
+    """Ensure an over-long wildcard import is left intact instead of being mangled.
+
+    A wildcard import such as ``from x.y.z import *`` cannot be wrapped - parenthesising a
+    star (``from x.y.z import (*)``) is a SyntaxError and the dotted module path cannot be
+    split. When the line exceeded the configured length and ``force_single_line`` was set,
+    isort used to fall through to its ``.`` splitter and rewrite the statement into invalid
+    Python, e.g.::
+
+        from very.very.(
+            line import *.long.very.very,
+        )
+
+    as reported in issue #2267: https://github.com/PyCQA/isort/issues/2267
+
+    The statement must be returned unchanged (matching Black) and must remain valid,
+    parseable Python across repeated runs.
+    """
+    test_input = (
+        "from very.very.very.very.very.very.very.very.very.very."
+        "very.very.very.very.very.very.very.very.long.line import *\n"
+    )
+
+    for kwargs in (
+        {"profile": "black", "force_single_line": True},
+        {"force_single_line": True},
+        {"profile": "black"},
+        {"profile": "black", "force_single_line": True, "combine_star": True},
+    ):
+        first_pass = isort.code(test_input, **kwargs)
+        assert first_pass == test_input, f"corrupted with {kwargs}"
+        # Must be valid Python and idempotent.
+        compile(first_pass, "<test>", "exec")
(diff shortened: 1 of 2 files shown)


tests/unit/test_regressions.py
```diff
@@ -2094,3 +2094,43 @@ def test_noqa_wrap_mode_idempotent_with_existing_comment():
...
+        "from very.very.very.very.very.very.very.very.very.very."
+        "very.very.very.very.very.very.very.very.long.line import *\n"
+    )
+
+    for kwargs in (
+        {"profile": "black", "force_single_line": True},
+        {"force_single_line": True},
+        {"profile": "black"},
+        {"profile": "black", "force_single_line": True, "combine_star": True},
+    ):
```
DanielNoord, line 2150: Just doing this with `"force_single_line": True` is enough. The rest is a bit overkill

tests/unit/test_regressions.py
```diff
@@ -2094,3 +2094,43 @@ def test_noqa_wrap_mode_idempotent_with_existing_comment():
...
+        first_pass = isort.code(test_input, **kwargs)
+        assert first_pass == test_input, f"corrupted with {kwargs}"
+        # Must be valid Python and idempotent.
+        compile(first_pass, "<test>", "exec")
```
DanielNoord, line 2154: Let's not do `compile`.

DanielNoord (commented):
Thanks! Good fix, just some nits about the test

DanielNoord:
@Labib-Bin-Salam Please see previous review.

DanielNoord:
Closing as this was an AI PR where the author failed to communicate with us.

PR #2560, opened 2026-06-17
Fix trailing comment dropped from aliased imports with combine_as_imports + force_single_line

### LINKED ISSUE #2094: Trailing comment removed from `as` imports with `combine_as_imports = true` and `force_single_line = true`
With `combine_as_imports = true` and `force_single_line = true`, trailing comments are removed from lines with `as` imports.

#### Versions

```shell
Python v3.11.0
isort v5.12.0
```

#### Config

```toml
# pyproject.toml
combine_as_imports = true
force_single_line = true
```

#### Input

```python
from other_module import other_function  # type: ignore [import]  # pylint: disable=no-name-in-module
from some_module import the_function as some_function  # type: ignore
from some_other_module import another_function as yet_another_function  # type: ignore [import]  # pylint: disable=no-name-in-module
```

#### Expected output

Comments preserved

#### Actual output

```python
from other_module import \
    other_function  # type: ignore [import]  # pylint: disable=no-name-in-module
from some_module import the_function as some_function
from some_other_module import another_function as yet_another_function
```

Labib-Bin-Salam:
## Summary

Fixes #2094.

With both `combine_as_imports = true` and `force_single_line = true`, the trailing comment on an `as` import is **silently dropped** (a data-loss bug):

```python
# combine_as_imports = true
# force_single_line = true

# input
from some_module import the_function as some_function  # type: ignore

# output before this PR  -> the comment is gone
from some_module import the_function as some_function
```

A trailing comment on a *non*-aliased single import was already preserved correctly; only aliased (`as`) imports lost theirs.

## Root cause

In `isort/parse.py`, when `combine_as_imports` is set, a comment on an aliased import is routed into the `{module}.__combined_as__` bucket. That bucket is only ever emitted by the *combined* code path in `output.py` (where aliases share one line). Under `force_single_line`, isort never takes that path, so the stored comment has nowhere to be emitted and is discarded.

## Fix

Because `force_single_line` puts every alias on its own line, the comment is now stored **per-import** (keyed by the full `name as alias`), mirroring how a non-aliased single import already keeps its trailing comment. The `__combined_as__` bucket is still used for every other case, and modules listed in `single_line_exclusions` are explicitly unaffected.

Result:

```python
from some_module import the_function as some_function  # type: ignore
```

The output is valid Python and idempotent (re-running isort is a no-op), and it now matches the output produced with `combine_as_imports = false`.

## Tests

Added `test_combine_as_imports_keeps_comment_with_force_single_line_issue_2094` to `tests/unit/test_regressions.py`, covering per-import comment association, idempotency, the exact example from the issue, and that `single_line_exclusions` behaviour is preserved. The test fails on `main` and passes with this change.

Full `tests/unit/` suite passes (no new failures); `ruff check`, `ruff format --check`, `flake8`, `mypy -p isort -p tests`, and `isort --profile hug --check` are all clean on the changed files.

### DIFF
--- a/isort/parse.py
+++ b/isort/parse.py
@@ -294,9 +294,31 @@ def file_contents(contents: str, config: Config = DEFAULT_CONFIG) -> ParsedConte
 
                     if comments and attach_comments_to is None:
                         if nested_module and config.combine_as_imports:
-                            attach_comments_to = categorized_comments["from"].setdefault(
-                                f"{top_level_module}.__combined_as__", []
-                            )
+                            if (
+                                config.force_single_line
+                                and top_level_module not in config.single_line_exclusions
+                            ):
+                                # ``force_single_line`` emits each aliased import on its own
+                                # line, so the trailing comment has to stay attached to this
+                                # specific import.  The ``__combined_as__`` bucket is only
+                                # consumed when the aliases are combined onto a single line, so
+                                # routing here would silently drop the comment (see issue #2094).
+                                # Store it per-import instead, mirroring how non-aliased single
+                                # imports keep their comment below.
+                                nested_for_module = categorized_comments["nested"].setdefault(
+                                    top_level_module, {}
+                                )
+                                existing_comment = nested_for_module.get(full_name, "")
+                                nested_for_module[full_name] = (
+                                    f"{existing_comment}"
+                                    f"{'; ' if existing_comment else ''}"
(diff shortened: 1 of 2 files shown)


DanielNoord:
Please engage with humans. See https://github.com/PyCQA/isort/pull/2559

PR #2566, opened 2026-06-25
Docs: clarify that files is a positional CLI arg and cannot be set in…

### LINKED ISSUE #2231: files configuration option missing from config file(s)
Is there a reason for [not supporting the files option in a config file](https://pycqa.github.io/isort/docs/configuration/options.html#files) (for example pyproject.toml)?

For me it would be useful to specify the files/directories in a pyproject.toml file and when I call isort, the files specified in the files option would get sorted without specifying them on the command line.

abcdemilyfghua:
Fixes #2231
The files option had a blank CLI flags entry and no explanation of why it isn't supported in config files. This adds a clarifying note explaining that files is a positional CLI argument (not a named option), and suggests workarounds like isort . or a Makefile target.

### DIFF
--- a/docs/configuration/options.md
+++ b/docs/configuration/options.md
@@ -1836,7 +1836,10 @@ One or more Python source files that need their imports sorted.
 **Python & Config File Name:** **Not Supported**  
 **CLI Flags:**
 
-- 
+- (positional argument)
+
+!!! note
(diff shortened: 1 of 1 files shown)


DanielNoord:
Fixing the docs to documenting is not working is not the same as making it work :smile:

The issues is clearly asking for this to be fixed, rather than documented as not supported.

PR #2574, opened 2026-07-03
Honor `# isort: skip` when a __future__ import is present (#2092)

### LINKED ISSUE #2092: `isort: skip` not working when futures imports are present
## Bug summary
`isort: skip` seems to not work when there is a futures import present.

## Reproducible example

Running `isort test.py` on this example works well. No changes are made.

```python
# test.py
from foo import bar  # isort: skip
from bar import baz

baz + bar
```

When modifying the file to include a futures import:

```python

from foo import bar  # isort: skip
from bar import baz

baz + bar
```

isort now moves the marked import to the bottom.

apoorvdarshan:
Fixes #2092.

## Problem

A per-line `# isort: skip` directive was silently ignored when a `__future__` import was present in the same block:

```python

from foo import bar  # isort: skip
from bar import baz
```

became

```python

from bar import baz

from foo import bar  # isort: skip
```

The skipped import is relocated below the block, violating the directive.

## Root cause

`__future__` imports are always floated to the very top of the file. Internally, isort collapses each contiguous run of imports into a single sorted block that is spliced back in at the position of the **first** import in the run (`parse.py` `import_index`). A `# isort: skip` line is kept out of the sort and re-emitted as a plain "out line", so it can only ever land *before* or *after* the whole block — never in the middle.

When the skipped import is the first line of its section (the common case), the block is inserted *after* it and it stays put. But as soon as any real import precedes it — most naturally a floated `__future__` import — the block is inserted *ahead* of the skipped line, pushing it below.

## Fix

Treat a per-line `isort: skip` import as an implicit section boundary **when earlier imports have already been collected into the current section**: flush those imports and emit the skipped statement in place, so nothing can be sorted above it. This mirrors isort's existing `# isort: split` behavior and keeps the skipped line exactly where the user put it.

The `contains_imports` guard is important — it means a skipped import that is already alone in its section (e.g. after a non-import statement) keeps flowing through the existing path unchanged, so trailing-whitespace normalization and other established behaviors are untouched.

## Verification

- Added `test_isort_skip_is_honored_with_future_import_issue_2092` to `tests/unit/test_regressions.py`. It fails on `main` (the skipped line is relocated) and passes with this change. It also covers an interleaved skip (`import` / skip / `import`) and asserts idempotency.
- The reported case is now a no-op, and the fix is stable across re-runs (`isort.check_code(...)` passes).
- Full unit suite green: `pytest tests/unit/` — `576 passed` (the only non-passing tests are pre-existing and environment-specific: test-ordering pollution in `test_importable`, which passes in isolation, and tests that shell out to the `isort` CLI / require the example plugins).
- `ruff check` and `ruff format --check` clean on the changed files (the `process` mccabe complexity stays within the configured limit — the new predicate is folded into the existing branch), and `mypy isort/core.py` reports no issues.
- Added a `CHANGELOG.md` entry.

---

*Disclosure: this change was prepared with the assistance of an AI tool (Claude Code). I reproduced the issue, reviewed and verified the fix, the test, and the full test run, and take responsibility for the contribution and will respond to review feedback personally.*

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -30,6 +30,12 @@ CODE_SORT_COMMENTS = (
     "# isort: assignments",
 )
 LITERAL_TYPE_MAPPING = {"(": "tuple", "[": "list", "{": "set"}
+SKIP_IMPORT_COMMENTS = ("isort:skip", "isort: skip")
+
+
+def _has_skip_comment(import_statement: str) -> bool:
+    """Return whether an import statement carries a per-line ``isort: skip`` directive."""
+    return any(comment in import_statement for comment in SKIP_IMPORT_COMMENTS)
 
 
 # Ignore DeepSource cyclomatic complexity check for this function.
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
Please remove the changelog line :)

apoorvdarshan:
Done — removed the changelog line. Thanks for the review!

DanielNoord (approved):

Issue #1815, opened 2021-09-24
Tuple sorting doesn't consider profile


kasium:
The tuple, list, etc sorting options don't consider the profile option.

Code:
```python
# isort: tuple
data = ("therearesuperlong", "therearesuperlong", "therearesuperlong", "therearesuperlong", "therearesuperlong")
```

`isort dummy.py --profile black`
```python
# isort: tuple
data = ('therearesuperlong', 'therearesuperlong', 'therearesuperlong', 'therearesuperlong',
 'therearesuperlong')
```

`black dummy.py`
```python
# isort: tuple
data = (
    "therearesuperlong",
    "therearesuperlong",
    "therearesuperlong",
    "therearesuperlong",
    "therearesuperlong",
)
```

I would expect that these sorting options consider the profile as normal. Also, the sorting options are not documented but noted  here #1358

jherman3:
I'm also experiencing this with `# isort: list`. It'd be great to have this so we can sort `__all__`.

tueda:
Actually, the tuple sorting should follow the `multi_line_output` option (or possibly a similar new option for the tuple sorting) rather than the profile.

As a first step, it would be OK if the output is compatible with black (as described in #1358), but in general [`literal.assignment`](https://github.com/PyCQA/isort/blob/be0fbd07817abcd5d2e62db9233b6633715d3405/isort/literal.py#L38) needs to implement calling a variant of [`wrap.import_statement`](https://github.com/PyCQA/isort/blob/99faae897f9da95cfd74ef9914c70313ed9abac3/isort/wrap.py#L10). (Maybe the author is aware of it, but he is too busy to implement it...)

timothycrosley:
This is true for all literal / type sorts, and is why they aren't documented / advertised features, even though If they were in a good spot I would do so because I think they are useful! 

```As a first step, it would be OK if the output is compatible with black (as described in #1358), but in general literal.assignment needs to implement calling a variant of wrap.import_statement. (Maybe the author is aware of it, but he is too busy to implement it...)

```
That's exactly right, I've just been busy, will accept PRs for this or will get to it at some point

DanielNoord:
Should be fixed by https://github.com/PyCQA/isort/pull/2576

kasium:
Thanks 👍

PR #2573, opened 2026-07-02
Wrap sorted literals with black profile settings

HarperZ9:
## Summary

- make literal sorting use vertical-hanging, parenthesized wrapping when the flat sorted assignment exceeds the configured line length
- keep existing `pprint` output for shorter literals and non-vertical modes
- add regression coverage for `# isort: tuple` under the Black profile

Closes #1815.

## Verification

- `uv run pytest tests/unit/test_literal.py -q`
- `uv run ruff check isort/literal.py tests/unit/test_literal.py`
- `uv run black --check --fast isort/literal.py tests/unit/test_literal.py`
- `uv run mypy isort/literal.py`

PR #2576, opened 2026-07-06
fix: make sorted __all__ and literals black-compatible (#2280)

### LINKED ISSUE #2280: `--sort-reexports` does not respect `--profile=black`
### Problem

Attempting to sort a multiline `__all__` attribute that is longer than the line limit with `isort . --sort-reexports --profile=black` wraps the lines in a non-black compatible way:

#### Before

```
__all__ = [
    "ArgumentReference",
    "ConfigurationSource",
    "HeptoFlow",
    "HeptoFlowContainer",
    "InMemoryCache",
    "Input",
    "InputReference",
    "InputValue",
    "Service",
    "parse_input_expression",
]
```

#### After

```
__all__ = ['ArgumentReference', 'ConfigurationSource', 'HeptoFlow', 'HeptoFlowContainer',
 'InMemoryCache', 'Input', 'InputReference', 'InputValue', 'Service',
 'parse_input_expression']
```

### Env

```
Python 3.12
isort 5.13.2
black 24.4.2
```

PS: Very willing to fix this myself and to PR it if someone can point me to the right approach.

lord-haffi:
Resolves: #2280, #1815

## Problem
`--sort-reexports` (and the `# isort: list/dict/tuple/…` comments) format the sorted result with stdlib `pprint`. Under `--profile=black` the result is single-quoted, packed in `pprint`'s wrap style and lacks a trailing comma — output black immediately reformats, so isort + black never reach a fixpoint.

## Fix
Format the sorted literal to be black-compatible: double quotes, a single line when it fits `line_length`, otherwise one item per line (vertical-hanging-indent) with a trailing comma when `include_trailing_comma` is set, keeping the literal's own bracket type. It's a fixed black/PEP 8 layout — it deliberately does not reproduce every `multi_line_output` mode for literals (VHI is what black uses).

isort's output for a long `__all__` under `--profile=black`, before → after this fix:

```diff
-__all__ = ['Address', 'AliasAddress', 'BankAccountConnection', 'Certificate',
- 'ConcessionFee']
+__all__ = [
+    "Address",
+    "AliasAddress",
+    "BankAccountConnection",
+    "Certificate",
+    "ConcessionFee",
+]
```

Why drop pprint

pprint exposes only width; quote style, wrap mode and trailing comma are hardcoded to non-black output, so it could never satisfy `--profile=black`. It was the sole cause of the bug, so it's removed entirely — dict sorting now uses the same formatter and gets the same fix.

**Also fixes a `--check` crash on multi-line `__all__`**

Emitting multi-line `__all__` exposed a latent bug in the `--sort-reexports` rollback: it wrote the `__all__` opening line then did `output_stream.seek(output_stream.tell() - reexport_rollback)` to overwrite it with the sorted literal. In check mode the output is a no-op stream whose `tell()` is always `0`, so the seek went to a negative position and raised `ValueError: Negative seek position`. The seek target is now clamped to `0`; real seekable output is unaffected.

Note: sorted string literals are now double-quoted by default (isort has no quote setting); existing literal-sort test expectations are updated to match.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -153,7 +153,7 @@ def process(
                 if is_reexport:
                     # Clamp to 0: in check mode the output is a no-op stream whose tell()
                     # stays 0, so an unclamped rollback would seek to a negative position
-                    # and raise ValueError. See issue #2280.
+                    # and raise ValueError. See PR #2576.
                     output_stream.seek(max(0, output_stream.tell() - reexport_rollback))
                     reexport_rollback = 0
                 sorted_code = textwrap.indent(
@@ -280,7 +280,7 @@ def process(
                         )
                         if is_reexport:
                             # Clamp to 0 so check mode's no-op output stream (tell() == 0)
-                            # cannot produce a negative seek position. See issue #2280.
+                            # cannot produce a negative seek position. See PR #2576.
                             output_stream.seek(max(0, output_stream.tell() - reexport_rollback))
                             reexport_rollback = 0
                         output_stream.write(sorted_code)
--- a/isort/core.py
+++ b/isort/core.py
@@ -151,7 +151,10 @@ def process(
 
             if code_sorting and code_sorting_section:
                 if is_reexport:
-                    output_stream.seek(output_stream.tell() - reexport_rollback)
+                    # Clamp to 0: in check mode the output is a no-op stream whose tell()
+                    # stays 0, so an unclamped rollback would seek to a negative position
+                    # and raise ValueError. See issue #2280.
+                    output_stream.seek(max(0, output_stream.tell() - reexport_rollback))
                     reexport_rollback = 0
                 sorted_code = textwrap.indent(
                     isort.literal.assignment(
@@ -276,7 +279,9 @@ def process(
                             ignore_whitespace=config.ignore_whitespace,
                         )
                         if is_reexport:
-                            output_stream.seek(output_stream.tell() - reexport_rollback)
+                            # Clamp to 0 so check mode's no-op output stream (tell() == 0)
+                            # cannot produce a negative seek position. See issue #2280.
+                            output_stream.seek(max(0, output_stream.tell() - reexport_rollback))
                             reexport_rollback = 0
                         output_stream.write(sorted_code)
                         if is_reexport:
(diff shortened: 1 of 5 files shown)


lord-haffi:
@Helveg Please take a look. For me it would work. I deliberately did not try to merge it with the formatting logic for the imports. In my head it made sense to keep that seperated from this logic here even though it is very similar at some parts. What are your thoughts on this?

lord-haffi:
I just realized, it probably addresses the same problem as https://github.com/PyCQA/isort/pull/2573.

lord-haffi:
Claude says:

### Overlap with #2573

Heads-up that #2573 (thanks @HarperZ9) targets the same area and also closes #1815, so these two should be reconciled. This PR also resolves **#1815** and is a superset of #2573 — I ran #2573 locally against the relevant cases to compare fairly:

| Case | #2573 | this PR (#2576) |
|---|---|---|
| `# isort: tuple` under `--profile=black` (#1815) | wraps to VHI but keeps **single quotes** (via `pprint.pformat`), so `black --check` still reformats it | double-quoted, `black --check` clean |
| Long `__all__` under `--profile=black` | single-quoted VHI → not black-stable | black-stable (verified on a real 196-entry `__all__`) |
| `isort --check` on a multi-line `__all__` preceded by imports | still raises `ValueError: Negative seek position` | fixed |
| `# isort: dict` | unchanged (still `pprint`) | same config-aware formatting |

Root differences:

- #2573 adds VHI wrapping only for the over-length + `use_parentheses` + `VERTICAL_HANGING_INDENT` path and otherwise **keeps `pprint`**, which emits single quotes — so the output isn't black-stable (black renormalizes the quotes).
- This PR removes `pprint` from literal sorting entirely and applies black's quote rule + single-line-or-VHI to every literal type, and additionally clamps the `sort_reexports` rollback seek in `core.py` (`isort --check` on a multi-line `__all__` currently crashes — see the check-mode note above).

Helveg:
Hi, I'm not a maintainer of this repo so I can't merge it, but FWIW I approve these changes :D

DanielNoord (approved):
Thanks

PR #2583, opened 2026-07-15
Avoid leading newline with float-to-top

### LINKED ISSUE #1935: `lines_before_import` adds newlines above module's docstring when used with `float_to_top`
Using MacOS, python `3.7.11`, isort `5.10.1`, and the following config in `pyproject.toml`:

```toml
[tool.isort]
lines_before_imports = 1
float_to_top = true
```

Before running `isort tmp.py`:

```python
"""
Tmp module.
"""
import os
os.getcwd()
```

After:

```python
  
"""
Tmp module.
"""

import os

os.getcwd()
```

(note the added leading empty line)

sakshichitnis27:
Fixes #1935

## Summary
- Avoid applying `lines_before_imports` during the preliminary `float_to_top` pass.
- Let the normal sorting pass apply the configured spacing in the correct location.
- Add a regression test ensuring module docstrings remain at the beginning of the file.

## Testing
- Reproduced the original issue using `lines_before_imports = 1` and `float_to_top = true`.
- Verified that the module docstring remains on the first line after the fix.
- `pytest -q` — 621 passed, 2 skipped
- `pytest -q tests/unit/test_regressions.py -k "float_to_top or lines_before_imports"`
- `ruff check isort/core.py tests/unit/test_regressions.py`
- `mypy isort/core.py tests/unit/test_regressions.py`
- `flake8 isort/core.py tests/unit/test_regressions.py`
- `isort --check-only isort/core.py tests/unit/test_regressions.py`

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -121,7 +121,10 @@ def process(
                         before = before[:-1]
                     extra_space = extra_space.replace("\n", "", 1)
                     sorted_output = output.sorted_imports(
-                        parsed, config, extension, import_type="import"
+                        parsed,
+                        Config(config=config, lines_before_imports=-1),
+                        extension,
+                        import_type="import",
                     )
                     made_changes = made_changes or _has_changed(
                         before=before,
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
Thanks for the fix!

It works, but I don't think this is how we'd like to solve this. I'd expect the fix to actually be in `sorted_imports` instead of overwriting a configuration value on the fly. That could have unintended side effects. Do you think you can fix this without doing this?

DanielNoord:
Closing for now, we can always reopen later :)

Issue #1396, opened 2020-08-21
Commented Import sorting


jameswinegar:
We've been using isort significantly across projects recently, and we've seen significant improvement in the capabilities. 

We have one sticking issue across projects with commented imports. It would be nicer if these commented imports were section aware. Below is our most common example, but it also applies to a strictly commented import as well. 

Setup.cfg
```
[isort]
line_length = 88
known_first_party = apps
known_django = django
sections = FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
```
Before isort
```
from django.shortcuts import render
from apps.profiler.models import Project
from django.contrib.auth.decorators import login_required
from django.views.generic import (
    # ListView, 
    # DetailView,
    TemplateView,
    # CreateView,
    # View
)
```
After isort (Actual)
```
from django.contrib.auth.decorators import login_required
from django.shortcuts import render

from apps.profiler.models import Project

from django.views.generic import TemplateView  # ListView,; DetailView,; CreateView,; View
```
After isort (Expected)
```
from django.contrib.auth.decorators import login_required
from django.shortcuts import render
from django.views.generic import TemplateView  # ListView,; DetailView,; CreateView,; View

from apps.profiler.models import Project
```

ndevenish:
Just encountered this issue (semicolon breaking sort), tracked it's cause down and then was pleased to find it's already been fixed!

PR #2588, opened 2026-07-16
fix: keep comment-only lines inside from-import groups

### LINKED ISSUE #1852: Comments in group imports are moved to a single line causing line too long violation (black, flake8)
Hi,
this issue seems to be related to #1009 #1843.

When using `isort` as a pre-commit hook with `black` profile. My commented imports are being moved to a single line.

Commented imports are intentional because I have some work in progress but I'd still like to commit (without `--no-verify`)

Example:
Before:
```python
from dj_rest_auth.views import (
    LoginView,
    LogoutView,
    # PasswordChangeView,
    # PasswordResetConfirmView,
    # PasswordResetView,
    # UserDetailsView,
)
```

After:
```
from dj_rest_auth.views import (  # PasswordChangeView,; PasswordResetConfirmView,; PasswordResetView,; UserDetailsView,
    LoginView,
    LogoutView,
)
```

My pre-commit hook configuration:
```yaml
# isort
  - repo: https://github.com/pycqa/isort
    rev: 5.10.1
    hooks:
      - id: isort

        name: isort (python)
        args: &isort_args
          - --line-length=100
          - --profile=black
          - --combine-as
          - --lines-after-imports=2
          - --src-path=app
      - id: isort
        name: isort (cython)
        types: [ cython ]
        args: *isort_args
      - id: isort
        name: isort (pyi)
        types: [ pyi ]
        args: *isort_args
```

I tried to use black flags `# fmt: off ... # fmt: on` but that had no effect.

alexchen-sys:
## Summary
Comment-only lines inside a parenthesised `from … import (` group were collapsed onto the opening import line as one semicolon-joined comment. With the black profile that rewrote intentional WIP comments and produced lines that fail line-length checkers.

This change tracks those body comments separately. When parentheses are used, they are re-emitted as their own indented comment lines inside the group. Non-parenthesised wrap modes keep the historical collapse behaviour so existing cases such as issue 1396 stay stable.

## Test plan
- [x] `pytest tests/unit/test_regressions.py::test_comment_only_lines_in_from_import_group_issue_1852`
- [x] `pytest tests/unit/test_regressions.py::test_incorrect_grouping_when_comments_issue_1396`
- [x] `pytest tests/unit/test_regressions.py::test_comment_on_opening_line_of_aliased_import_does_not_move`
- [x] `pytest tests/unit/test_isort.py tests/unit/test_regressions.py tests/unit/test_comments.py` (331 passed, 1 skipped)

Fixes #1852


alexchen-sys:
DeepSource was failing because the existing `skipcq: PY-R1000` comment landed on the small helper I added instead of `_with_from_imports`.

Moved the skip back onto `_with_from_imports` (same pattern as before). No product change; regression for #1852 still passes.

alexchen-sys:
Fixed flake8 E303 by removing one extra blank line before the #1852 regression test. No product change.

alexchen-sys:
Lint Attempt #2 failed on `ruff format --check` for `tests/unit/test_regressions.py` (not flake8 E303). Applied `ruff format` to the #1852 regression test only; product behavior unchanged. Focused pytest still passes.

alexchen-sys:
Residual loop closeout (multi-stage A/C/D/B1/B2 + mechanical gate PASS):

- Fixed force_single_line silent drop of from_body comments (now attaches to first single-line import, main-compatible).
- Fixed combine_star orphan indented body comments (fold onto star statement).
- Added regressions for both paths; primary black #1852 + #1396 still green.
- Head includes 123e7d3 after prior style residuals.

No further product thrash planned; remaining secondary nits (as-only / all-comment-only) disclosed as ALLOW_WITH_NITS scope.

alexchen-sys:
Quick update without the process noise from my last note:

I fixed two edge cases on the current head: force_single_line was dropping body comments, and combine_star left orphan indented comments. Both have regression tests, and the original #1852 black-profile case still passes.

No further product thrash planned from me. Happy to adjust if review wants a different default.

DanielNoord (commented):
Please ensure the full patch has coverage :)

For the tests I'd prefer the style of the first test so making explicit what the final output looks like instead of using `"..." in out`. However, if the output is the same as the input please use something like `assert isort.code(to_sort, profile="black", line_length=100) == to_sort`

alexchen-sys:
Thanks — I rewrote the #1852 tests to use explicit expected output (including the force_single_line case) and added coverage for the remaining from_body / inject branches.

DanielNoord (commented):
Can you fix CI to pass? I'll have a look at this afterwards :)

DanielNoord (commented):
CI still isn't fixed.

alexchen-sys:
Rebased onto main (kept the new alias-comment regression), fixed the lint/format gap, and closed the from_body drops on as-only / force_single emit paths. Local unit suite + mypy/flake8/ruff/hug check are green on this tip.

DanielNoord (commented):
We're missing coverage again.

alexchen-sys:
Thanks — added coverage for the remaining from_body inject branches (`ignore_comments`, nested-merge fold, and force_single plain-before-as). No product change on this tip.

alexchen-sys:
Added two regression asserts covering the last uncovered branch in the existing-comment merge (empty segment from a trailing separator, plus the dedup case). Tests only, no product changes — codecov/patch should be at 100% on the diff now.

DanielNoord:
This will need a rebase unfortunately. It is quite a big PR and is currently not on the top of my priority list for the next release, sorry!

tests/unit/test_regressions.py
```diff
@@ -2443,3 +2447,237 @@ def test_isort_does_not_drop_aliased_import_when_plain_name_has_a_comment():
...
+    section = "THIRDPARTY"
+    module = "foo"
+    captured: list[list[str]] = []
+    original_with_comments = with_comments
+
+    def with_comments_probe(comments, *args, **kwargs):
+        captured.append(list(comments or []))
+        return original_with_comments(comments, *args, **kwargs)
+
+    monkeypatch.setattr("isort.output.with_comments", with_comments_probe)
```
DanielNoord, line 2684: This won't work for the `mypyc` compiled tests. Can we refactr this? If not we'll need to skip it.

DanielNoord (commented):
Sorry, still haven't gotten to this. I did do a rebase on `main` for you which uncovered an issue, hopefully you can fix it!

alexchen-sys:
Yes — reproduced the failure locally too; I had been testing without mypyc before, so the compiled legs never showed it on my side.

Fixed by skipping the monkeypatch-based test under mypyc, using the same _IS_COMPILED pattern as test_main.py:58. I ran locally on my pc, 13/13 pass; behavior is covered by the sibling e2e test_combine_star_folds_body_comments_issue_1852.

isort/output.py
```diff
@@ -297,6 +298,69 @@ def _build_import_group(
...
 
 
+def _inject_from_body_comments(
+    import_statement: str,
+    body_comments: list[str],
+    line_separator: str,
+    indent: str,
+    *,
+    comment_prefix: str,
+    ignore_comments: bool = False,
```
Manny7717, line 308: Nit: both call sites invoke this with ignore_comments=False (each is already inside a `not config.ignore_comments` guard), so this parameter is currently dead — fine to keep if you expect a future caller, otherwise removable.

Manny7717 (approved):
## Review — Approved

Verified locally on head: **109/109 tests pass** (test_regressions.py + test_output.py), and I ran an idempotency check on a comment-only-body sample in both modes:

- default (single-line fold): `from foo import Bar, Baz, Qux  # PasswordChangeView,; keep this` — stable under a second sort ✓
- black profile (parenthesized): comment lines kept as their own indented lines inside the parens — stable under a second sort ✓

The parse-side split of comment-only continuation lines into a separate `from_body` bucket is the right design: it keeps attribute comments attached to their import while giving the output stage enough information to re-emit body comments as real lines. The `pending_comments` handling for `force_single_line` and the star-import fold both look correct.

Two non-blocking nits (inline): the `ignore_comments` parameter of `_inject_from_body_comments` is always passed `False` by its call sites (both are already guarded by `not config.ignore_comments`), and the `comment_prefix` thread-through could default to `config.comment_prefix` — both trivially removable if you want to slim the signature.

Issue #2352, opened 2025-01-29
Unexpected change to imports when updating to isort 6.0.0


sarahboyce:
On Django, our `isort` config is:
```
[tool.isort]
profile = "black"
default_section = "THIRDPARTY"
known_first_party = "django"
```

With `isort==5.13.2`, the following imports are sorted as:
```python
from django.db.models.sql.compiler import (
    SQLAggregateCompiler,
    SQLCompiler,
    SQLDeleteCompiler,
)
from django.db.models.sql.compiler import SQLInsertCompiler as BaseSQLInsertCompiler
from django.db.models.sql.compiler import SQLUpdateCompiler
```

But on `isort==6.0.0`, they are now sorted as:
```python
from django.db.models.sql.compiler import (
    SQLAggregateCompiler,
    SQLCompiler,
    SQLDeleteCompiler,
)
from django.db.models.sql.compiler import SQLInsertCompiler as BaseSQLInsertCompiler
from django.db.models.sql.compiler import (
    SQLUpdateCompiler,
)
```

We don't believe this is a desirable change and so are pinned at 5.13.2

(Please let me know if you need more details. Our pr for reference: https://github.com/django/django/pull/19109)

LepkoQQ:
We also noticed this issue, but with an extra step. This doesn't always happen the first time depending on the input.

For example if you start with code like this:
 
```python
from another_file_aaaaaaaaaaaaaaaaaa import \
    SOME_THING_CONSTANT as SOME_THING_CONSTANT_RENAMED
from another_file_aaaaaaaaaaaaaaaaaa import some_func
```

then you run `isort . --profile black` and the output is:

```python
from another_file_aaaaaaaaaaaaaaaaaa import (
    SOME_THING_CONSTANT as SOME_THING_CONSTANT_RENAMED,
)
from another_file_aaaaaaaaaaaaaaaaaa import some_func
```

then you run `isort . --profile black` AGAIN and then the output is:

```python
from another_file_aaaaaaaaaaaaaaaaaa import (
    SOME_THING_CONSTANT as SOME_THING_CONSTANT_RENAMED,
)
from another_file_aaaaaaaaaaaaaaaaaa import (
    some_func,
)
```

aholmes:
@LepkoQQ this seems related to this breaking change (not documented as such!) in default behavior. https://github.com/PyCQA/isort/pull/2236

[...]

PR #2592, opened 2026-07-19
fix: ignore wrapping paren in force_sort_within_sections keys

### LINKED ISSUE #2455: `--fss` may lead to incorrect import separation for mixed multi-line and alias imports
When using `--force-sort-within-sections` (`--fss`), imports from the same module can be incorrectly separated.

Reproduction (`repro.py`):

```py
from module import (
    AAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCC,
)
from module import DDDDDDDDDDDDDDDDDDDDDDDDDD as d
from module import (
    EEEEEEEEEEEEEEEEEEEEEEEEEE,
    FFFFFFFFFFFFFFFFFFFFFFFFFF,
    GGGGGGGGGGGGGGGGGGGGGGGGGG,
)
```

Sorting this file with `isort repro.py --fss` results in imports from the same module being split in a seemingly meaningless way:
```py
from module import (
    AAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCC,
)
from module import (
    EEEEEEEEEEEEEEEEEEEEEEEEEE,
    FFFFFFFFFFFFFFFFFFFFFFFFFF,
    GGGGGGGGGGGGGGGGGGGGGGGGGG,
)
from module import DDDDDDDDDDDDDDDDDDDDDDDDDD as d
```

This happens because `_with_from_imports` sees those as three separate groups, and then during within-section sorting `sorting.section_key` determines that `import (...` should precede `import D...`.

It's hard to say that isort is doing something wrong, but splitting imports from the same module is undesirable. My preferred output in this situation would be:
```py
from module import (
    AAAAAAAAAAAAAAAAAAAAAAAAAA,
    BBBBBBBBBBBBBBBBBBBBBBBBBB,
    CCCCCCCCCCCCCCCCCCCCCCCCCC,
    EEEEEEEEEEEEEEEEEEEEEEEEEE,
    FFFFFFFFFFFFFFFFFFFFFFFFFF,
    GGGGGGGGGGGGGGGGGGGGGGGGGG,
)
from module import DDDDDDDDDDDDDDDDDDDDDDDDDD as d
```

I think there are a few ways to approach this:                                                                         
                                                                                                                         
- Add a new configuration option: Introduce a flag (e.g., `--force-separate-as-imports`) that operates within the `_with_from_imports` function to explicitly place [as_imports](https://github.com/PyCQA/isort/blob/024aa37e44c9f29964a4fca355a4680fd4511979/isort/output.py#L285) at the end of the list.                                     
- Normalize `(` in section_key: Modify the key generation so that `module import (AAAA` is treated as `module import AAAA` for sorting purposes. This would not produce my "preferred output," but it would at least leave repro.py unchanged, avoiding multiple separated `from module import (` blocks.
- No new flag, no new sort logic, just merge outputs from [new_section_output](https://github.com/PyCQA/isort/blob/024aa37e44c9f29964a4fca355a4680fd4511979/isort/output.py#L111): Iterate over the new section output, parse them again, and merge imports where possible. I personally think this would be the least preferred option because it doesn't integrate well with the rest of the system and would have a high maintenance cost.

I would appreciate any input from the maintainers on whether this is worth addressing and, if so, which approach is preferred. I am happy to contribute a PR if the team agrees on a direction.

l46983284-cpu:
## Summary
Under `--force-sort_within_sections`, multi-line `from` imports were sorted as `import (` / newline-indented names, which collates before `import Alias` and can pull a later plain group ahead of an alias from the same module (issue #2455).

This normalizes the wrapping parenthesis and following whitespace in `section_key` so those lines sort by the first imported name, matching single-line imports.

## Non-goals
Does **not** merge separate plain `from module import (...)` statements across an intermediate alias. That larger contract change conflicts with #2352 trailing plain-after-alias stability. Issue #2455 listed this `section_key` normalization as a valid approach.

## Test plan
- [x] `tests/unit/test_regressions.py::test_from_import_alias_does_not_split_plain_names_issue_2455` (default GRID + black multi-line under fss)
- [x] existing #2352 trailing-comma mixed-as tests still pass
- [x] mypy, isort --check, flake8, ruff check, ruff format --check

Closes #2455

### DIFF
--- a/isort/__init__.py
+++ b/isort/__init__.py
@@ -21,9 +21,8 @@ __all__ = (
 
 from . import settings
 from ._version import __version__
-from .api import ImportKey
-from .api import check_code_string as check_code
 from .api import (
+    ImportKey,
     check_file,
     check_stream,
     find_imports_in_code,
@@ -33,6 +32,7 @@ from .api import (
     place_module,
     place_module_with_reason,
 )
+from .api import check_code_string as check_code
 from .api import sort_code_string as code
 from .api import sort_file as file
 from .api import sort_stream as stream
--- a/isort/output.py
+++ b/isort/output.py
@@ -335,6 +335,32 @@ def _with_from_imports(
             for from_import, sub_module in zip(from_imports, sub_modules, strict=False)
             if sub_module in parsed.as_map["from"]
         }
+        # For #2455, defer a single unannotated alias so all direct names for the
+        # module share one statement before that alias is rendered. Complex alias sets,
+        # comment-bearing imports, and compatibility modes retain the established path.
+        defer_as_imports = (
+            not config.combine_as_imports
+            and not config.force_single_line
+            and not config.no_inline_sort
+            and not config.remove_redundant_aliases
+            and not config.combine_star
+            and not ("*" in from_imports and config.combine_star)
+            and any(
+                parsed.imports[section][import_key][module][from_import]
+                for from_import in from_imports
+            )
+            and not parsed.categorized_comments["from"].get(module)
+            and not any(
+                parsed.categorized_comments["straight"].get(f"{module}.{from_import}")
+                or parsed.categorized_comments["nested"].get(module, {}).get(from_import)
+                or any(
+                    parsed.categorized_comments["nested"].get(module, {}).get(as_import)
+                    for as_import in aliases
+                )
+                for from_import, aliases in as_imports.items()
+            )
+        )
+        deferred_as_imports: list[str] = []
         if config.combine_as_imports and not ("*" in from_imports and config.combine_star):
             if not config.no_inline_sort:
                 for as_import in as_imports:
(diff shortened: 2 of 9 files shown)


alexchen-sys:
DeepSource flagged PY-R1000 on `section_key` after the #2455 paren-normalization branch. I moved that strip into a tiny `_strip_wrapping_import_paren` helper so the behavior is unchanged and the complexity bump is gone.

Local: ruff clean on the touched files, `tests/unit/test_regressions.py` 92 passed.

DanielNoord (commented):
This does not seem to implement the style as agreed upon in the final comments of that issue. Any reason why you deviated from that?

alexchen-sys:
Thanks — I switched this to the agreed style from the final comments on #2455: plain names from the same module are merged into one import, with aliases after. Also fixed --fss so a lexically earlier alias no longer sorts above plains.

DanielNoord (commented):
Instead of keeping track of these cases (in this quite complex code), wouldn't it be easier to merge import statements at the end? That seems like it would be simpler to understand and maintain

DanielNoord (commented):
This seems like it changes our style signficantly. I think we should first discuss other alternatives in the issue. Unfortunately the style as discussed there woulkd be a breaking change for the project.

alexchen-sys:
No pressure at all — feel free to close this PR or reuse any part of the code if it's useful. It was a fun exercise to work through. If you have thoughts on what direction would make sense for the issue, I'd be happy to collaborate on a follow-up.

DanielNoord:
I'll close for now, thanks for having a stab at this!

PR #2597, opened 2026-07-24
fix: don't crash on --config-root without --resolve-all-configs

### LINKED ISSUE #1990: `--config-root` flag doesn't work
```
$ isort --check --config-root .. .
  File "/home/eitan/.pyenv/versions/venv/bin/isort", line 8, in <module>
    sys.exit(main())
  File "/home/eitan/.pyenv/versions/3.9.7/envs/venv/lib/python3.9/site-packages/isort/main.py", line 1108, in main
  File "/home/eitan/.pyenv/versions/3.9.7/envs/venv/lib/python3.9/site-packages/isort/settings.py", line 524, in __init__
    raise UnsupportedSettings(unsupported_config_errors)
isort.exceptions.UnsupportedSettings: isort was provided settings that it doesn't support:

        - config_root = ..  (source: 'runtime')

For a complete and up-to-date listing of supported settings see: https://pycqa.github.io/isort/docs/configuration/options.

### PR DESCRIPTION
**Pull request by Gooh456**
yeah this one's just been broken since forever. `--config-root` on its own throws a raw `UnsupportedSettings` traceback instead of doing... anything.

Root cause is in `main.py`: `config_root` only gets popped out of `config_dict` inside the `if resolve_all_configs:` branch. So if you pass `--config-root` without also passing `--resolve-all-configs`, it's still sitting in the dict when it gets splatted into `Config(**config_dict)`, and `Config` has no idea what `config_root` is, so it errors out.

The help text for the flag literally says it's "used with the --resolve-all-configs flag" - so without that flag it should just be a no-op, not a crash. Fix is to pop it unconditionally, same as `resolve_all_configs` itself already is a few lines up.

Verified against main: reverted my change locally and reproduced the exact traceback from #1990 (`isort.exceptions.UnsupportedSettings: ... - config_root = ..`). With the fix it runs fine.

Added a regression test in `tests/unit/test_main.py` that calls `--config-root` without `--resolve-all-configs` and just checks it doesn't blow up.

Fixes #1990

### DIFF
--- a/isort/main.py
+++ b/isort/main.py
@@ -1038,13 +1038,14 @@ def main(argv: Sequence[str] | None = None, stdin: TextIOWrapper | None = None) 
     ext_format = config_dict.pop("ext_format", None)
     allow_root = config_dict.pop("allow_root", None)
     resolve_all_configs = config_dict.pop("resolve_all_configs", False)
+    config_root = config_dict.pop("config_root", ".")
     wrong_sorted_files = False
     all_attempt_broken = False
     no_valid_encodings = False
 
     config_trie: Trie | None = None
     if resolve_all_configs:
-        config_trie = find_all_configs(config_dict.pop("config_root", "."))
+        config_trie = find_all_configs(config_root)
 
     if "src_paths" in config_dict:
         # Keep CLI-provided values as-is so wildcard patterns can be expanded later
--- a/isort/main.py
+++ b/isort/main.py
@@ -1038,7 +1038,10 @@ def main(argv: Sequence[str] | None = None, stdin: TextIOWrapper | None = None) 
     ext_format = config_dict.pop("ext_format", None)
     allow_root = config_dict.pop("allow_root", None)
     resolve_all_configs = config_dict.pop("resolve_all_configs", False)
+    config_root_given = "config_root" in config_dict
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by DanielNoord — COMMENTED** [review]
We should give a warning (by raising an error) that this config isn't useful without `resolve_all_configs`. See other examples in the codebase.

**Comment by Gooh456** [comment]
good call. added that, exits with a clear message now instead of the confusing UnsupportedSettings crash. same sys.exit("Error: ...") pattern used a few lines up for the no-paths case. updated the old test to expect the error and added one for config-root + resolve-all-configs together still working.

File: isort/main.py [hunk 1]
```diff
@@ -1038,13 +1038,17 @@ def main(argv: Sequence[str] | None = None, stdin: TextIOWrapper | None = None)
     ext_format = config_dict.pop("ext_format", None)
     allow_root = config_dict.pop("allow_root", None)
     resolve_all_configs = config_dict.pop("resolve_all_configs", False)
+    config_root_given = "config_root" in config_dict
+    config_root = config_dict.pop("config_root", ".")
```
DanielNoord: This can remain on line 1051 now

Gooh456:
yeah good catch, moved it back inline. no reason to hold it in a separate var anymore.

DanielNoord (approved):

PR #2599, opened 2026-07-26
Fix preservation of form-feed blank lines

### LINKED ISSUE #2562: ISort incorrectly removes some blank lines
When processing lines, ISort treats some whitespace-only lines incorrectly.

[isort-blank-lines.py.txt](https://github.com/user-attachments/files/29186598/isort-blank-lines.py.txt)
The attached file contains this content:

    import os
    import sys

    
    print("Hello, world!")

When ISort process that input file, it incorrectly removes one of the blank lines:

    $ python3 -m isort --check-only --diff ./isort-blank-lines.py 
    ERROR: /home/bignose/Projects/python/isort-blank-lines.py Imports are incorrectly sorted.
    --- /home/bignose/Projects/python/isort-blank-lines.py:before   2019-01-26 21:14:47.421581
    +++ /home/bignose/Projects/python/isort-blank-lines.py:after    2019-01-26 21:15:18.837737
    @@ -1,6 +1,4 @@
     import os
     import sys
     
    --
     print("Hello, world!")

The input file is valid, and should not be altered by ISort, because the second blank line (containing only U+000C FORM FEED) is a valid blank line in Python code.

According to Python [syntax rules for indentation](https://docs.python.org/3/reference/lexical_analysis.html#indentation), (and recently acknowledged in https://github.com/PyCQA/isort/pull/2436) U+000C do not constitute indentation space:

> A formfeed character may be present at the start of the line; it will be ignored for the indentation calculations […].

PEP 8 [recommendations for blank lines](https://peps.python.org/pep-0008/#blank-lines) specify that those two blank lines should remain:

> Surround top-level function and class definitions with two blank lines.

By removing one of the two blank lines, ISort renders the formerly PEP-8 conformant file, now in violation of that recommendation.

So, a line consisting of U+000C is a valid empty blank line; as part of a “two blank lines” as recommended by PEP 8, it should not be stripped, and should not be altered by ISort.

utkarshalpha:
## Summary

- split source text only on actual newline sequences so form feeds remain part of their source line
- preserve form feeds while normalizing the configured number of blank lines after imports
- add the maintainer-proposed regression cases for both default and explicit `lines_after_imports`

Closes #2562.

## Tests

- `python -m pytest tests/unit/test_parse.py tests/unit/test_regressions.py -q` (113 passed)
- `python -m pytest tests/unit --ignore=tests/unit/test_importable.py -q` (607 passed, 5 skipped; the excluded smoke test is an unrelated baseline failure on current main)
- `ruff check isort/parse.py isort/output.py tests/unit/test_regressions.py --ignore PLC0207`
- `python -m compileall -q isort`
- `git diff --check`

### DIFF
--- a/isort/output.py
+++ b/isort/output.py
@@ -132,10 +132,11 @@ def sorted_imports(
 
     if output:
         imports_tail = output_at + len(output)
+        trailing_blank_lines: list[str] = []
         while [
             character.strip() for character in formatted_output[imports_tail : imports_tail + 1]
         ] == [""]:
-            formatted_output.pop(imports_tail)
+            trailing_blank_lines.append(formatted_output.pop(imports_tail))
 
         if config.lines_before_imports != -1:
             lines_before_imports = config.lines_before_imports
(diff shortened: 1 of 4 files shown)


DanielNoord (commented):
It seems we are missing some coverage. Can you add that? :)

utkarshalpha:
Added the missing coverage in f766bf60: tests for the previously unexercised empty-contents branch in `parse.file_contents`, plus parametrized `\n`/`\r\n`/`\r` cases pinning down that splitting happens only on real newlines (so form feeds stay attached to their line). Patch coverage should be 100% now.

DanielNoord (approved):
Thanks!

PR #2603, opened 2026-07-28
Fix Windows stdin/stdout newline preservation

shleder:
## Problem

When isort processes source through stdin on Windows, the default `TextIOWrapper` newline translation can convert LF input to CRLF output or introduce duplicated carriage returns. This affects normal sorting, skipped-file passthrough, check mode, and diff output.

## Root cause

The CLI streamed through `sys.stdin` and `sys.stdout` without disabling platform newline translation. On Windows, the standard text streams could therefore translate newlines independently of isort's own line-ending handling.

## Implementation

- Reconfigure `sys.stdout` with `newline=""` before processing stdin.
- Reconfigure `sys.stdin` with `newline=""` only when the CLI uses the real standard input.
- Never reconfigure an explicitly injected stdin stream.
- Gracefully tolerate streams that do not support `reconfigure()`.
- Narrow the standard streams to `TextIOWrapper` with `cast()` for strict mypy compatibility.
- Add an Unreleased changelog entry.

## Regression coverage

The new regression test verifies exact LF output, exact CRLF output, non-ASCII content preservation, LF and CRLF `isort: skip_file` passthrough, sorted and unsorted `--check-only` behavior, LF diff output without introduced carriage returns, CRLF diff output without duplicated carriage returns, injected stdin streams remaining untouched, and streams without `reconfigure()` completing without a crash.

## Validation

The patch was applied to a pristine checkout of base commit `47c4f9819c3e92d296efe43c2f460e6742a1a75f` and compared with an independent pristine checkout of the same commit.

- The focused regression passes on the patched checkout. The test does not exist on the base checkout.
- The patched checkout adds no new full-suite failures. The same two failures occur on the pristine base and patched checkouts: `tests/unit/test_importable.py::test_importable` and `tests/unit/test_regressions.py::test_windows_diff_too_large_misrepresentative_issue_1348`.
- Full strict mypy produces the same pre-existing Windows diagnostic on base and patched: `tests/unit/test_settings.py:86: error: Module has no attribute "mkfifo"  [attr-defined]`.
- Ruff check passes on base and patched.
- Ruff format check passes on base and patched.
- Configured flake8 passes on base and patched.
- `git diff --check` passes for the patch.
- The full patched unit run reports 606 passed, 5 skipped, and the same 2 baseline failures. The full base unit run reports 605 passed, 5 skipped, and those same 2 failures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Find out more about isort's release policy [here](docs/major_releases/release_po
 
 ### Unreleased
 
+  - Fixed Windows stdin/stdout newline preservation (#2453)
+
 ### 8.0.0 February 19 2026 
 
   - Removed `--old-finders` and `--magic-placement` flags and `old_finders` configuration option. The legacy finder logic that relied on environment introspection has been removed (#2445) @joao-faria-dev
--- a/isort/main.py
+++ b/isort/main.py
@@ -8,9 +8,9 @@ import sys
 from collections.abc import Iterator, Sequence
 from contextlib import AbstractContextManager, nullcontext
 from gettext import gettext as _
-from io import TextIOWrapper
+from io import TextIOWrapper, UnsupportedOperation
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 from warnings import warn
 
 from . import __version__, api, files, sections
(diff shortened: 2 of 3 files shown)


DanielNoord (commented):
I don't think we should be reconfguring `sys.stdin`. `main` might be called programatically by other Python programs that do not want a reconfigured `sys.stdin`.

PR #2612, opened 2026-08-04
Group plain imports

### LINKED ISSUE #2455: `--fss` may lead to incorrect import separation for mixed multi-line and alias imports

DanielNoord:
Closes https://github.com/PyCQA/isort/issues/2455

I'm not very happy with the code, that might need another pass.
This does seem to pass the test cases as discussed in the linked issue.

### DIFF
--- a/isort/__init__.py
+++ b/isort/__init__.py
@@ -21,9 +21,8 @@ __all__ = (
 
 from . import settings
 from ._version import __version__
-from .api import ImportKey
-from .api import check_code_string as check_code
 from .api import (
+    ImportKey,
     check_file,
     check_stream,
     find_imports_in_code,
@@ -33,6 +32,7 @@ from .api import (
     place_module,
     place_module_with_reason,
 )
+from .api import check_code_string as check_code
 from .api import sort_code_string as code
 from .api import sort_file as file
 from .api import sort_stream as stream
--- a/isort/output.py
+++ b/isort/output.py
@@ -297,6 +297,72 @@ def _build_import_group(
     return group_output
 
 
+def _build_as_import(
+    parsed: parse.ParsedContent,
+    module: str,
+    from_comments: list[str],
+    as_import: str,
+    import_start: str,
+    config: Config,
+) -> tuple[str, list[str]]:
+    # `from_comments` at this point contains any comments that appeared on
+    # the *opening* "from X import" line. These are distinct from
+    # `specific_comment`, which is an inline comment on the attribute line
+    # itself. We snapshot `from_comments` here so that we can later distinguish
+    # the two types: opening-line comments must stay on the "import (" line
+    # when parentheses are used, while attribute-line comments stay on the
+    # import attribute line.
+    specific_comment = parsed.categorized_comments["nested"].get(module, {}).pop(as_import, None)
+    opening_line_comments = [comment for comment in from_comments if comment != specific_comment]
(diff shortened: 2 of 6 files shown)


DanielNoord:
Superseded by #2614

PR #2621, opened 2026-08-08
docs: explain comment handling during sorting

### LINKED ISSUE #1843: Add some documentation about how isort moves comments.
It appears that:
1. comments before an import line (with no intervening blank line) will be sorted with that import
2. comments on their own line within an multiline import statement are moved to the opening line of that import (at least for multi_line_mode = 3)
3. trailing comments on an import item (in multi_line_mode = 3) are floated to a new single import of that item
4. If there are multiple import lines from the same module (for example due point 3 above), comments before any of them are floated to be above the first one.

I spent some time trying to figure out this behavior, or see if there was some way to configure it.

In particular, I was looking to see if there was a way to change some the behavior. A few of things I was looking for
a. the ability to have a comment on a multiline import be on the line following the `from module import (` line, rather than on it
b. the ability to have a comment on the line before an import item in a multiline import sort with that import.
c. the ability to have import items with trailing comments kept in the original multiline import, rather than floated to the top.
 
I realize that a comment before the first import item would have an ambiguous interpretation if both (a) and (b) were supported. These occurred to me at different times looking at specific places where I had comments, and trying to figure out a way to have comments near the appropriate thing in each case. If I had known about (1) and (3), I probably would have spent less time trying to find a way to do (a) or (b) (though (c) would still be useful, I think).

Himanshuagrawal4:
## Summary

- document how leading, same-module, multiline, and inline import comments are rendered
- link to action comments and the configuration options that affect comment handling
- add the new page to the documentation index

Closes #1843

## Validation

- compared all four documented transformations with the current `isort.code` output
- built the complete Sphinx documentation environment with warnings treated as errors using the network-independent dummy builder

### DIFF
--- a/docs/configuration/comments.md
+++ b/docs/configuration/comments.md
@@ -1,0 +1,103 @@
+# Comments on imports
+
+isort preserves comments on imports, but their position can change when the
+associated import is sorted or reformatted. This is because comments are
+associated with import statements and imported names while parsing, then
+rendered with the sorted imports.
+
+## Comments before imports
+
+A comment immediately before an import, with no blank line between them, moves
+with that import:
+
+```python
+import zlib
+# Used to inspect the current process state.
+import os
+```
+
+becomes:
+
+```python
+# Used to inspect the current process state.
+import os
+import zlib
+```
+
+When imports from the same module are combined, comments attached to those
+imports are emitted before the combined statement:
+
+```python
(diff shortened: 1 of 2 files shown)


DanielNoord:
Not sure how helpful this is. It can always change and this is not tied to any tests. Closing for now.

PR #2626, opened 2026-08-14
Keep add_imports below a prefixed module docstring

### LINKED ISSUE #1893: add_import adds imports above module docstring when docstring is raw string
* I tried to use the add_import functionality like so:
```add_imports = "from __future__ import annotations"```

However, it kept destroying my module level docstrings by adding an import before the """ string.
In the project I was working on, we use raw multiline strings`r"""` which may be messing with the heuristics to detect module level docstrings here: https://github.com/PyCQA/isort/blob/c6a41965247a858a0afd848fbebfca18b8983917/isort/parse.py#L218

As a temporary workaround, adding the append_only flag has seemed to fix it.

In short, isort will format my code like the following:
```python
r"""module docstring
"""
```

Here is my isort config:
```
add_imports = "from __future__ import annotations"
# append_only = true # Enabling this is the workaround
profile = 'black'
treat_comments_as_code = "# %%"
```

Eljees:
Closes #1893.

`add_imports` inserts the import *above* the module docstring when that docstring carries a string prefix:

```pycon
>>> import isort
>>> src = 'r\"\"\"module docstring\n\"\"\"\nfrom __future__ import annotations\n'
>>> print(isort.code(src, add_imports=["from __future__ import annotations"]))

r"""module docstring
"""
```

The import is both moved above the docstring and duplicated, because the file already contained it. Reported by @Skylion007 in 2022 and hit again by @sodul in 2024.

### Cause

The guard in `isort/core.py` that stops the insertion tests the raw line:

```python
and not line.lstrip().startswith(COMMENT_INDICATORS)
```

It only recognises a docstring that *begins* with a quote. `r"""...` begins with a letter, so the guard misses it and the import goes in above the docstring.

A single-line prefixed docstring escapes this by accident: the neighbouring `endswith(DOCSTRING_INDICATORS)` guard catches it. Only multi-line prefixed docstrings are affected. Every legal prefix and both quote flavours break; an unprefixed docstring never does.

(The report points at `isort/parse.py:218`; the actual place is the guard above.)

### Fix

A small helper that strips a legal string prefix, but only when a quote follows it immediately -- so ordinary identifiers like `rating = 1` are left alone.

### Tests

One regression test in `tests/unit/test_regressions.py`, three blocks: the reporter's exact input; a control that the import is *still added*, below the docstring; and a sweep over 12 prefixes x 2 quote flavours. It fails on `main` with the duplicated import shown above.

`pytest tests/unit` gives 615 passed with this change against 614 on a clean tree -- the same 3 failures in both runs (`test_importable` and two `test_ticketed_features` cases needing the example plugin packages), so no regressions. `flake8`, `ruff check`, `ruff format --check`, `isort --profile hug --check` and `mypy` are clean on both changed files.

### DIFF
--- a/isort/core.py
+++ b/isort/core.py
@@ -47,6 +47,15 @@ def _strip_string_prefix(line: str) -> str:
     return line
 
 
+def _is_comment_or_string_start(line: str) -> bool:
+    """Return whether `line` opens a comment, a docstring or any other string literal.
+
+    A string prefix (`r`, `b`, `f`, `u` or a legal combination of them) is skipped
+    first, so a prefixed docstring is recognised exactly as an unprefixed one is.
+    """
+    return _strip_string_prefix(line.lstrip()).startswith(COMMENT_INDICATORS)
+
+
 def _has_skip_comment(import_statement: str) -> bool:
     """Return whether an import statement carries a per-line ``isort: skip`` directive."""
     return any(comment in import_statement for comment in SKIP_IMPORT_COMMENTS)
@@ -439,7 +448,7 @@ def process(
                 and not in_top_comment
                 and not was_in_quote
                 and not import_section
-                and not _strip_string_prefix(line.lstrip()).startswith(COMMENT_INDICATORS)
+                and not _is_comment_or_string_start(line)
                 and not (line.rstrip().endswith(DOCSTRING_INDICATORS) and "=" not in line)
             ):
                 add_line_separator = line_separator or "\n"
--- a/isort/core.py
+++ b/isort/core.py
@@ -20,6 +20,8 @@ IMPORT_START_IDENTIFIERS = (
 )
 DOCSTRING_INDICATORS = ('"""', "'''")
 COMMENT_INDICATORS = (*DOCSTRING_INDICATORS, "'", '"', "#")
+# Every legal Python string prefix, lower-cased; `u` cannot be combined with another.
+STRING_PREFIXES = frozenset(("r", "u", "f", "b", "fr", "rf", "br", "rb"))
 CODE_SORT_COMMENTS = (
     "# isort: list",
     "# isort: dict",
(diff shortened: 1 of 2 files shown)


tests/unit/test_regressions.py
```diff
@@ -2443,3 +2443,29 @@ def test_isort_does_not_drop_aliased_import_when_plain_name_has_a_comment():
...
+    # The import still has to be added, below the docstring -- where an unprefixed
+    # docstring already puts it.
+    with_code = 'r"""module docstring\n"""\n\nx = 1\n'
+    assert (
+        isort.code(with_code, add_imports=["import a"])
+        == 'r"""module docstring\n"""\n\nimport a\n\nx = 1\n'
+    )
+
+    # Every legal prefix, in both quote flavours and both cases.
+    for prefix in ("r", "R", "b", "B", "f", "F", "u", "U", "rb", "bR", "Fr", "rf"):
```
DanielNoord: Can we reuse `STRING_PREFIXES` here?

isort/core.py
```diff
@@ -425,7 +439,7 @@ def process(
                 and not in_top_comment
                 and not was_in_quote
                 and not import_section
-                and not line.lstrip().startswith(COMMENT_INDICATORS)
+                and not _strip_string_prefix(line.lstrip()).startswith(COMMENT_INDICATORS)
```
DanielNoord: Probably better to create one shared "Is start of a comment" utility instead to make clear what we are testing for here.

Eljees:
Both done, thanks — and the branch is rebased onto current main.

**The predicate has a name.** `_is_comment_or_string_start(line)` replaces the inline `_strip_string_prefix(line.lstrip()).startswith(COMMENT_INDICATORS)` at the call site in `process`, so the condition reads as the question it is asking. I did not fold the other `startswith("#")` sites into it: those ask "is this a comment", while this one asks "is this a comment *or* a string of any kind", and the second is what the docstring case needs.

**The test reuses `STRING_PREFIXES`.** It now iterates `sorted(STRING_PREFIXES)` and derives the case variants from each prefix rather than repeating a hand-written list, so the two cannot drift:

```python
for prefix in sorted(STRING_PREFIXES):
    for cased in sorted({prefix, prefix.upper(), prefix.capitalize()}):
        for quote in ('"""', "'''"):
```

### Measured

The reported input, run against the two revisions of `core.py`:

| `isort/core.py` | `isort.code(r-prefixed docstring + import, add_imports=[same import])` |
| --- | --- |
| current `main` | the import is emitted a second time above the docstring |
| this branch | unchanged, as it should be |

`pytest tests/unit`: **621 passed, 1 skipped, 2 failed** — and those two, `test_isort_supports_shared_profiles_issue_970` and `test_sort_configurable_sort_issue_1732`, fail identically on a clean checkout of `main` in this environment (they want an `example` profile and a `natural_plus` sort order that come from plugins that are not installed). `main` alone gives 620 passed, so the delta is exactly the one test this PR adds.

One thing the rebase fixed rather than the review: before it, `tests/unit/test_importable.py::test_importable` failed on the branch and passed on `main` — the branch simply predated the `pytest.raises(SystemExit)` change to that file. Worth knowing it was staleness and not this change.

`ruff check` and `ruff format --check` are clean on both files.

DanielNoord (approved):

PR #2634, opened 2026-08-18
Clarify src path help text

amarisgreat:
## Summary

Clarify the `--src-path` help text to make the behavior of configured
source paths easier to understand.

## Changes

- Clarify that modules found within configured source paths are automatically
  categorized as `first_party`.
- No classification logic or behavior was changed.

## Testing

- `pytest tests/unit/test_place.py`
- 4 passed


DanielNoord:
I don't see what is wrong with the old code to be honest.

PR #2605, opened 2026-08-02
Preserve trailing-comma layout when sorting reexports

PSR94:
## Summary
Preserve explicitly multiline `__all__` collections with trailing commas when `--sort-reexports` runs under trailing-comma-aware formatting such as the Black profile.

## Why this helps
This keeps sorted reexports stable with Black's magic trailing comma behavior, so short `__all__` declarations that are intentionally one item per line are not collapsed after sorting.

## Changes made
- Detect when a source literal is multiline and ends with a trailing comma.
- Preserve that multiline layout when the active config includes trailing commas.
- Add regression coverage for short tuple-style and list-style `__all__` exports.

## Testing
- `uv run pytest tests/unit/test_regressions.py -k '2578 or 2280'`
- `uv run pytest tests/unit/test_isort.py -k 'reexport_multiline'`
- `uv run ruff format --check isort/literal.py tests/unit/test_regressions.py`
- `uv run ruff check isort/literal.py tests/unit/test_regressions.py`
- `uv run mypy isort tests`
- `./scripts/done.sh`

Closes #2578

PR #2644, opened 2026-08-28
fix: keep one-per-line layout of short multiline __all__ under trailing-comma profiles

Manny7717:
## What

`--sort-reexports` rewrote a short multi-line `__all__` (one element per line, trailing comma) into a single line whenever the sorted result happened to fit within `line_length` — even under `--profile black`, dropping the trailing comma and the one-per-line style that black itself preserves. Reported in #2578.

```python
# before
__all__ = (
    "FirstClass",
    "SecondClass",
)
# after --sort-reexports --profile black
__all__ = ("FirstClass", "SecondClass")
```

## Change

`isort.literal` now preserves the vertical block form when the original literal was multi-line AND the active config uses trailing commas (`include_trailing_comma`, set by the black/comma-style profiles). Single-line sources and default-config behavior are unchanged — the pre-existing tests encoding the collapse behavior for the default config still pass.

## Tests

- `test_sort_reexports_preserves_short_multiline_all_issue_2578`: the issue's repro stays one-per-line and sorted, trailing comma intact. Fails on the previous implementation (verified).
- `test_sort_reexports_keeps_short_single_line_all_issue_2578`: single-line `__all__` is sorted in place and stays single-line.
- All existing reexport tests (incl. the black-stability fixpoint test) pass; full suite 623 passed, 4 pre-existing env-dependent failures unrelated to this change.


Manny7717:
Closing as a duplicate of #2605, which implements the same fix for the same issue (#2578) with the same mechanism — a preserve-trailing-comma flag threaded through the collection sorters, gated on `include_trailing_comma`. #2605 predates this PR (Aug 17 vs Aug 28) and is already under maintainer review, and I verified its head passes the full literal + reexport test suite locally. Deferring to it — thanks @PSR94 for the cleaner `_has_trailing_comma` formulation.

PR #2645, opened 2026-08-29
Keep long comments off the opening line and within line_length

### LINKED ISSUE #2124: Line Length Ignores Comments
I've been trying to deploy isort to my code repository using the black profile, but when it ran it would merge the comments inside our imports and add them to the same line as the import location which violates the line length significantly.
This is what It looks like initially.
```python
from os.path import (
join,
# this is a really really really really really really really really 
# really really really really really really long comment
getsize,
)
```
This is what happens after it runs.
```python
from os.path import (  # this is a really really really really really really really really; really really really really really really long comment
    getsize,
    join,
)
```

SemTiOne:
- Resolves #2124

vertical_hanging_indent now renders each comment fragment on its own `#` line instead of merging them onto the `from ... import (` line with `"; "`.

### DIFF
--- a/isort/wrap_modes.py
+++ b/isort/wrap_modes.py
@@ -174,8 +174,10 @@ def vertical_hanging_indent(**interface: Any) -> str:
     _imports = ("," + interface["line_separator"] + interface["indent"]).join(interface["imports"])
     _comma_maybe = "," if interface["include_trailing_comma"] else ""
     opening = f"{interface['statement']}({_line_with_comments}"
-    _comment_text = " ".join(interface["comments"]) if interface["comments"] else ""
-    _is_functional_comment = "noqa" in _comment_text or "type:" in _comment_text
+    _is_functional_comment = any(
+        comment.strip().lower().startswith(("noqa", "type: ignore"))
+        for comment in (interface["comments"] or [])
+    )
     if (
         _line_with_comments
         and len(opening) > interface["line_length"]
--- a/isort/wrap_modes.py
+++ b/isort/wrap_modes.py
@@ -173,8 +173,28 @@ def vertical_hanging_indent(**interface: Any) -> str:
     )
     _imports = ("," + interface["line_separator"] + interface["indent"]).join(interface["imports"])
     _comma_maybe = "," if interface["include_trailing_comma"] else ""
+    opening = f"{interface['statement']}({_line_with_comments}"
+    _comment_text = " ".join(interface["comments"]) if interface["comments"] else ""
+    _is_functional_comment = "noqa" in _comment_text or "type:" in _comment_text
+    if (
+        _line_with_comments
+        and len(opening) > interface["line_length"]
+        and not _is_functional_comment
+    ):
+        _comment_on_own_line = isort.comments.add_to_line(
+            interface["comments"],
+            interface["indent"],
+            removed=interface["remove_comments"],
+            comment_prefix=interface["comment_prefix"].lstrip(),
+        )
+        return (
+            f"{interface['statement']}({interface['line_separator']}"
+            f"{_comment_on_own_line}{interface['line_separator']}"
+            f"{interface['indent']}{_imports}{_comma_maybe}"
+            f"{interface['line_separator']})"
+        )
     return (
-        f"{interface['statement']}({_line_with_comments}{interface['line_separator']}"
+        f"{opening}{interface['line_separator']}"
         f"{interface['indent']}{_imports}{_comma_maybe}{interface['line_separator']})"
     )
 
--- a/isort/wrap_modes.py
+++ b/isort/wrap_modes.py
@@ -183,11 +183,14 @@ def vertical_hanging_indent(**interface: Any) -> str:
         and len(opening) > interface["line_length"]
         and not _is_functional_comment
     ):
-        _comment_on_own_line = isort.comments.add_to_line(
-            interface["comments"],
-            interface["indent"],
-            removed=interface["remove_comments"],
-            comment_prefix=interface["comment_prefix"].lstrip(),
+        _comment_on_own_line = interface["line_separator"].join(
+            isort.comments.add_to_line(
+                [single_comment],
+                interface["indent"],
+                removed=interface["remove_comments"],
+                comment_prefix=interface["comment_prefix"].lstrip(),
+            )
+            for single_comment in (interface["comments"] or [])
         )
         return (
             f"{interface['statement']}({interface['line_separator']}"
--- a/isort/wrap.py
+++ b/isort/wrap.py
@@ -19,7 +19,7 @@ def import_statement(
     """Returns a multi-line wrapped form of the provided from import statement."""
     if explode:
         formatter = vertical_hanging_indent
-        line_length = 1
+        line_length = config.wrap_length or config.line_length
         include_trailing_comma = True
     else:
         formatter = formatter_from_string((multi_line_output or config.multi_line_output).name)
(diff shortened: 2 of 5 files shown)


isort/wrap_modes.py
```diff
@@ -173,8 +173,28 @@ def vertical_hanging_indent(**interface: Any) -> str:
     )
     _imports = ("," + interface["line_separator"] + interface["indent"]).join(interface["imports"])
     _comma_maybe = "," if interface["include_trailing_comma"] else ""
+    opening = f"{interface['statement']}({_line_with_comments}"
+    _comment_text = " ".join(interface["comments"]) if interface["comments"] else ""
+    _is_functional_comment = "noqa" in _comment_text or "type:" in _comment_text
```
Manny7717: Nit (non-blocking): `"type:" in _comment_text` also matches comments that merely mention `type:` without being a `# type: ignore` directive, e.g. `# check type: int and more...` — verified locally that such a comment is kept on an over-long opening line. Safe direction (old behavior), but anchoring to a directive shape (`# noqa` / `# type: ignore` at comment start) would avoid surprises.

Manny7717 (approved):
Verified locally on head 67213aa (tests/unit: 627 passed / 2 failed / 1 skipped — the 2 failures are the known pre-existing FileNotFoundError env noise (issue_909/938), byte-identical on main; main baseline is 623 passed, so the +4 are exactly this PR's new tests, zero regressions).

**Fix behavior confirmed:**
- Issue #2124 shape (black profile, long comment inside a multi-line from-import): the comment now moves to its own line under the opening paren instead of being merged onto `from ... import (` — `isort.code` output verified, and the result is idempotent (re-running produces identical output).
- Functional comments are correctly exempt: `# noqa: F401` and `# type: ignore` stay on the opening line (verified end-to-end — moving them would break their meaning).
- Short comments still stay on the opening line (existing behavior preserved, covered by `test_vertical_hanging_indent_short_comment_stays_on_opening_line`).
- `wrap.import_statement(explode=True)` now passes the real `line_length` instead of `1`; checked the `vertical_hanging_indent` interface — `line_length` is only consumed by the new branch, so no other explode-mode behavior changes.
- isort self-lint (`--profile hug --check --diff isort/ tests/`) and ruff on the changed files: clean.

**Non-blocking nit:** the functional-comment detection uses a loose substring (`"noqa" in text or "type:" in text`). A comment that merely mentions these tokens (e.g. `# check type: int ...`) is treated as functional and stays on the over-long opening line. This errs in the safe direction (it's the old behavior), so not a blocker — but anchoring the check to a directive-shaped pattern (e.g. comment starting with `# noqa` / `# type: ignore`) would make it tighter. No change requested.

isort/wrap_modes.py
SemTiOne: Valid, thanks!

Manny7717 (approved):
Re-verified on 9683ad6 — my wrap_modes.py:178 nit is fully addressed.

**Verification (local):**

1. **Regression proven:** `test_vertical_hanging_indent_non_directive_type_comment_moves_off` FAILS on 67213aa (old substring `"type:" in _comment_text` kept a `check type: int` comment on the opening line) and PASSES on 9683ad6. Exactly the false-positive I flagged.
2. **Functional exemptions intact (executed probes):** long `noqa: F401 ...` comment stays on the opening line (`from os.path import (  # noqa: ...`); long `type: ignore ...` stays as well. Non-directive `check type: int` and plain long comments move off to their own line under the paren.
3. **No regressions:** full unit suite on head = 628 passed / 2 failed — the 2 failures (issue_909/938 FileNotFoundError) are the documented pre-existing env noise, byte-identical on base. Base = 627 passed / 3 failed (env noise + the new regression test). Zero new failures.

The `comment.strip().lower().startswith(("noqa", "type: ignore"))` per-comment check is a strictly better heuristic than the substring match, and the test pins the non-directive case so it can't regress.

DanielNoord (commented):
After seeing the code I am not sure that this is a bug. Not sorting the imports also isn't the correct behaviour. Perhaps there is just no solution for this?

SemTiOne:
Isn't issue #2124 about the merged comment violating line_length? The user had two separate comment lines that got collapsed onto `from ... import (` and blew past 88 chars.

With the fix:
```python
from os.path import (
    # this is a really really really really really really really really
    # really really really really really really long comment
    getsize,
    join,
)

--- line lengths ---
1:  21 | from os.path import (
2:  71 |     # this is a really really really really really really really really
3:  60 |     # really really really really really really long comment
4:  12 |     getsize,
5:   9 |     join,
6:   1 | )
--- max: 71 ---
--- idempotent: True ---
```

Manny7717 (approved):
Re-approving on the new head 94104cf ("Render each comment fragment on its own line"). This commit answers the maintainer's concern that a long merged comment could still violate line_length: each comment fragment now renders on its own # line under the paren via line_separator.join(add_to_line([fragment], ...)).

Verified locally on 94104cf:
- New test test_vertical_hanging_indent_multi_fragment_comments_each_on_own_line passes; every output line asserted <= 88 chars.
- End-to-end: two long comment fragments under VHI (multi_line_output=3, line_length=88) render on separate indented lines, imports sorted, idempotent (re-run byte-identical).
- Short comments unchanged (stay on opening line when they fit); functional comments (noqa/type: ignore) still exempt.
- Full unit suite: 629 passed / 2 failed (issue_909/938 FileNotFoundError env noise, identical on main) / 1 skipped - zero regressions, exactly +1 test.

Note on the maintainer thread: per-fragment rendering is the right resolution - the previous behavior (all fragments joined onto one line) could still blow past line_length, which is the actual #2124 complaint. Approved.

SemTiOne:
After looking at it again, I found some nits and edge cases. I'll fix them soon.

DanielNoord:
This seems to be a breaking change for one of the projects in the primer. Are we sure we are on the right track here? I don't expect this to require breaking changes, or, put differently, don't think this is important enough to fix to warrant breaking changes.

SemTiOne:
> This seems to be a breaking change for one of the projects in the primer. Are we sure we are on the right track here? I don't expect this to require breaking changes, or, put differently, don't think this is important enough to fix to warrant breaking changes.

Oop, my bad. I accidentally changed the existing primer behavior, and this fix is way too broad. Let me fix it.

SemTiOne:
`parse.py:306` stores these opening comments in the `__combined_as__` bucket, while `output.py:766` pops them back without preserving opening provenance. The current `opening` recording only covers the plain `from` bucket.

Could this be a follow-up issue? @DanielNoord

isort/parse.py
```diff
@@ -31,6 +31,7 @@
             "straight": dict[str, list[str]],
             "nested": dict[str, dict[str, str]],
             "above": CommentsAboveDict,
+            "opening": dict[str, list[str]],
```
DanielNoord: What is the difference between these and the `from` comments?

_As you have seen I am a bit worried about the scope and size of this fix for a relatively simple "bug". At some point a tool can't handle all the different code that is coming in. As long as it is still valid Python we do not to accept some edge cases can't be handled if the complexity to handle them becomes too much._

DanielNoord (changes requested):

isort/parse.py
SemTiOne: > What is the difference between these and the `from` comments?

`opening` is only provenance for the subset of `from` comments that originally appeared on the opening line.

> _As you have seen I am a bit worried about the scope and size of this fix for a relatively simple "bug". At some point a tool can't handle all the different code that is coming in. As long as it is still valid Python we do not to accept some edge cases can't be handled if the complexity to handle them becomes too much._

This only addresses #2124 in VHI, it doesn't try to preserve every possible comment edge case or change other wrap modes. If any part still feels too broad, point it out and I'll trim it.

SemTiOne:
Oh the “Cannot be rebased due to conflicts” status was confusing. I tried rebasing it locally, but that didn’t work. The fix was simply to sync the fork on my forked repo.

PR #2651, opened 2026-09-04
Fix multi_line_output=10 emitting unparsable code around trailing comments

### LINKED ISSUE #2650: multi_line_output=10 emits unparsable code when an import has a trailing comment
### Summary

With `multi_line_output=10` (`HANGING_INDENT_WITH_PARENTHESES`), isort can rewrite a valid file into one that no longer parses when an import carries a trailing comment. Both shapes below start as valid Python.

**Closing parenthesis captured by the comment**

```console
$ echo 'from a import b, c  # trailing
from a import d' | isort --multi-line 10 --line-length 40 --force-grid-wrap 2 -
from a import (b, c, d  # trailing)
```

`)` is inside the comment, so the statement is never closed.

**Comma lost at a wrap point**

```console
$ echo 'from mypkg.submodule import alpha, beta, gamma, delta, epsilon, zeta, eta  # noqa: F401' | isort --multi-line 10 -
from mypkg.submodule import (alpha, beta, delta, epsilon, eta  # noqa: F401
    gamma, zeta)
```

The comma after `eta` is gone, so the names read as `eta gamma`.

Both outputs raise `SyntaxError` under `ast.parse`, and isort itself then refuses them with `ExistingSyntaxErrors: Parenthesis is not closed`.

### How much of the option space this affects

Sweeping 7 sources x 12 wrap modes x 3 line lengths x 6 option combinations (1512 runs), **17 produce output that Python cannot parse, and all 17 are `multi_line_output=10`.** The other 11 modes are clean. It reproduces at the default line length and at 88, with `# noqa:` and `# type: ignore` — the comments most likely to sit on an import.

### Why only this mode

The sibling grouped modes hoist the comment onto the `(` line, so their last line never ends in a comment and appending `,` or `)` is always safe. `hanging_indent_with_parentheses` keeps the comment on the import line and then appends punctuation to it unconditionally:

- the final `return` appends `)` after the comment, and
- at a wrap point, `interface["statement"] + ","` puts the comma after the comment, where `comments.add_to_line` strips it off with the old comment before re-adding it — so the comma is dropped rather than commented out.

### Version

`main` at 131f4adc, isort 9.0.1, CPython 3.12.13.

I have a fix and a regression test ready and will open a PR referencing this issue.

dylanpulver:
Fixes #2650.

`multi_line_output=10` (`HANGING_INDENT_WITH_PARENTHESES`) can rewrite a valid file into one that no longer parses when an import carries a trailing comment:

```console
$ isort --multi-line 10 --line-length 40 --force-grid-wrap 2 -
from a import (b, c, d  # trailing)
```
`)` sits inside the comment.

```console
$ isort --multi-line 10 -
from mypkg.submodule import (alpha, beta, delta, epsilon, eta  # noqa: F401
    gamma, zeta)
```
the comma after `eta` is gone. Both raise `SyntaxError`.

### Cause

The sibling grouped modes hoist the comment onto the `(` line, so their last line never ends in a comment. This mode keeps it on the import line and then appends punctuation unconditionally: the final `return` puts `)` after the comment, and at a wrap point `statement + ","` puts the comma after the comment, where `comments.add_to_line` strips it off with the old comment before re-adding.

### Fix

`_add_syntax()` places `,`/`)` before a trailing comment on the last line, used at both sites.

### Verification

- 7 sources x 12 modes x 3 line lengths x 6 option sets (1512 runs): **17 unparsable outputs before, all mode 10; 0 after.**
- Suite: 625 passed 1 skipped before, 626 after (the new test); nothing else changed.
- `isort --profile hug --check-only`, `ruff format --check`, `ruff check`, `mypy` clean.
- Reverting `wrap_modes.py` fails the new test on `'(' was never closed`; a partial fix that only relocates `)` still fails it on the lost comma.


Manny7717 (approved):
Verified locally against head f09b1ce8 (base = origin/main @ 131f4adc). Bug is real and the fix is correct.

**Bug confirmed:** mode 10 (`HANGING_INDENT_WITH_PARENTHESES`) appended punctuation after a trailing comment. CLI repro on base: `from a import b, c, d  # trailing\nfrom a import e\n` with `--multi-line 10 --line-length 40 --force-grid-wrap 2` emits `from a import (b, c, d, e  # trailing)` — the `)` lands inside the comment (SyntaxError). Same input on head emits valid `from a import (b, c, d, e)  # trailing`.

**Regression coverage proven:** transplanted the new test into the base worktree — it fails there with the exact reported `SyntaxError: '(' was never closed`, and passes on head.

**No collateral:** full unit suite — head 624 passed / 2 failed / 1 skipped; base+new-test 623 passed / 3 failed. The 2 failures on head (`test_settings_path_skip_issue_909`, `test_skip_paths_issue_938`, FileNotFoundError) reproduce identically on base (pre-existing env noise, unrelated). `ruff check` and `ruff format --check` clean on both changed files.

**Fix review:** `_add_syntax()` correctly handles both append sites (the wrap-point comma and the final closing paren), preserves original spacing before `#` (`code[len(code.rstrip()):] or "  "`), and is a no-op passthrough when the last line carries no comment. `partition("#")` is safe here because import statements in this mode cannot contain `#` before the comment start. The in-test matrix (all `WrapModes` × 3 line lengths × 3 sources × trailing-comma on/off, every output `ast.parse`d) is a good guard against this class of bug across sibling modes.

DanielNoord (approved):

PR #2652, opened 2026-09-04
Keep overlong star imports stable with Black

### LINKED ISSUE #2649: isort disagrees with black on wrapping over-long star imports
## Description

Only over-long wildcard imports are affected. Under `profile = black` (`line_length = 88`), a `from x import *` under 88 chars is left alone; once it crosses 88, isort tries to wrap it. Parentheses are not an option, `from x import (*)` is a `SyntaxError`, and the dotted module path is the only over-long part, so isort falls back to a backslash continuation.

Black leaves such a line untouched (it also has no way to split it), so isort and black now fight over every long `import *` in a codebase.

Introduced by 65bfa80 ("Fix word-wrapping of 'from ... import *' into invalid Python (#2267)"), first shipped in 9.0.1. That commit fixed real corruption — thanks — but its message says star imports are "left on a single line", while the code in `isort/wrap.py` deliberately emits a backslash instead. PR #2559, which proposed returning the statement untouched to match black, was closed unmerged.

## Steps to reproduce

```ini
# .isort.cfg
[settings]
profile = black
```

```python
# 90 chars - isort rewrites this
from mypkg.vendor.hardware.sdk.bindings.generated.device_parameter_constants_v210 import *

# 88 chars - left alone, as expected
from mypkg.vendor.hardware.sdk.bindings.generated.device_parameter_constants_v2 import *
```

`isort .` gives:

```python
from mypkg.vendor.hardware.sdk.bindings.generated.device_parameter_constants_v210 import \
    *
```

`black .` then rejoins it to one line. Running `isort` and `black` in either order does not converge.

## Expected

Star imports left on a single line regardless of length under `profile = black`, matching black.

## Notes

No setting avoids it: tested every `multi_line_output` (0-11), `use_parentheses`, `wrap_length`, `balanced_wrapping`, `force_single_line`, `combine_star`, `split_on_trailing_comma`. Only raising `line_length` above the line's length works, which then desyncs isort from black on ordinary long imports. `multi_line_output = 7` (NOQA) keeps one line but appends `# NOQA`.

Reproduced on 9.0.1 and on `main` (Python 3.11).

Spidy092:
## Summary

- Keep over-long `from ... import *` and `from ... cimport *` statements intact when they cannot be safely wrapped.
- Preserve trailing comments and the existing `NOQA` wrapping mode behavior.
- Add regression coverage for direct wrapping and the Black profile.

Fixes #2649

## Testing

- `python3 -m pytest -q tests/unit/test_wrap.py` (8 passed)
- `ruff check isort/wrap.py tests/unit/test_wrap.py`
- `ruff format --check isort/wrap.py tests/unit/test_wrap.py`
- Direct reproduction: output remains unchanged, compiles, and is idempotent under `profile="black"`.
- The full `tests/unit` collection is blocked in the isolated harness because the test-only dependencies `black`, `hypothesis`, and `colorama` are unavailable.


Manny7717 (approved):
Verified at head 9ef16a16 — this resolves the isort↔black non-convergence from #2649 exactly as reported.

**Regression-proven:** the 3 rewritten star-import spec tests FAIL on base 131f4adc (backslash output still emitted: `import \\` + `    *`) and PASS at head (statement returned untouched). 8/8 in tests/unit/test_wrap.py at head.

**End-to-end convergence probe (real CLI, black 26.3.1, profile=black, over-long `from ... import *`):**
- Base 131f4adc: isort backslash-wraps → black rejoins to one line → isort wraps again → alternates forever (no fixed point — reproduces #2649 verbatim, incl. "Fixing" on every isort pass).
- Head: isort leaves the line → black leaves it → isort leaves it. Stable after one pass.

**No collateral:** full tests/unit at head 623 passed / 2 failed / 1 skipped — the 2 failures are the pre-existing issue_909/issue_938 FileNotFoundError env-noise tests (identical on clean main across all my prior runs). The `endswith("import *")` condition is stricter than the old `endswith("*")` (only star from/cimport lines match; short star imports still return via the length check before this branch; comments preserved via the earlier `#` split). The #2267 corruption fix (never emit `import (*)`) remains satisfied — the statement is now kept whole and valid instead of backslash-split.

Non-blocking nit: the `# skipcq: PY-R1000` directive comments a function whose complexity this PR actually reduces (branch removed, ~6 lines deleted); the two-line comment is also redundant. If the DeepSource gate still trips at the threshold, fine to keep, but consider dropping both if the removal alone clears it — external-service gates shouldn't pin unrelated comments in the tree.

tests/unit/test_wrap.py
```diff
@@ -50,33 +50,25 @@ def test_line__comment_with_brackets__expects_unchanged_comment(multi_line_outpu
     assert wrap.line(content=content, line_separator="\n", config=config) == expected
 
 
-def test_line_star_import_wrapped_with_backslash() -> None:
-    """Star imports cannot use parenthesis-based wrapping, so should use backslashes.
-
-    See issue #2267.
-    """
+def test_line_star_import_is_not_wrapped() -> None:
```
DanielNoord: This is not using the `black` profile. Why are we changing behaviour here?

DanielNoord:
Superseded by https://github.com/PyCQA/isort/pull/2665 which is a more targetted fix.

PR #2655, opened 2026-09-09
Preserve blank lines before comment-only sections

### LINKED ISSUE #2156: --lines-before-import inconsistent between writing to a stream compared to a file in some cases
Hi,

First of all I'm sorry if this issue is a bit of a mess, I dug into `isort` quite a bit to try to get to the bottom of the issue, and the process might not be as straightforward as it could.

The symptom of the issue is that there is a small difference in behavior between running `isort` when writing to a file directly, compared to writing to a stream.

In my testing, it's at least easily reproducible with the `--lines-before-import` option: it removes some newlines when streaming, in some cases.

I did not test this on 5.12 as I'm stuck with Python 3.7 for now, but I didn't see anything mentioning this issue in the changelog. So for this test, I was on 5.11.5.

Here are the steps to recreate the environment I tested this in:

```sh
mkdir isort-sandbox
cd isort-sandbox/
python3 -m venv venv
source venv/bin/activate
pip install isort
```

Here is a minimal script to reproduce the issue, called `script.py` in later sections:

```python
def my_function():
    def inner_function():
        return 1

    # A comment
    return inner_function()

```

The line before `# A comment` gets removed when running in stream mode, for exemple the following command:

```sh
isort --lines-before-import=1 --diff - --file script.py < script.py
```

...produces this diff:

```diff
--- script.py:before    2023-07-13 14:09:33.457270
+++ script.py:after     2023-07-13 14:10:35.409390
@@ -1,6 +1,5 @@
 def my_function():
     def inner_function():
         return 1
-
     # A comment
     return inner_function()
```

Whereas running this command:

```sh
isort --lines-before-import=1 --diff script.py
```

...does not produce any output.

Furthermore, running `isort` on the file directly, but with the `--stdout` flag, results in the same issue: the newline disappears.

```python
isort --lines-before-import=1 --stdout script.py 
```

...produces this output:

```python
def my_function():
    def inner_function():
        return 1
    # A comment
    return inner_function()
```

After digging into `isort.api` and `isort.core` a bit, I think this may be because when streaming directly to output, we don't have the chance to ensure we actually changed anything relevant in import blocks, whereas when we write to file after the fact (the regular `isort script.py` call), we first check if we made any changes to import blocks before computing a diff and applying it to the file.

As for why the newline disappears in the first place, I think the issue is caused by the different indentation levels: for example, if I add a `pass` just before the comment then there is no longer a diff when running in stream mode:

```python
def my_function():
    def inner_function():
        return 1

    pass
    # A comment
    return inner_function()
```

From debugging the code, without the `pass` statement the comment was considered an `import_section`, which made `isort` ignore my new line.

Here's the relevant condition: https://github.com/PyCQA/isort/blob/5.11.5/isort/core.py#L284

Which has the consequence of not writing lines above the comment (in my example, a newline): https://github.com/PyCQA/isort/blob/5.11.5/isort/core.py#L365

So, there is really to separate issues:

1. `isort` may produce an output that is not clean in some cases (especially since the example script did not have any import, I would expect no difference between the input and output)
2. Writing directly to a stream does not allow us to safeguard against such edge cases.

I don't know if there is a good solution for both those issues, but solving 1. kind of makes 2. irrelevant.

For what it's worth, in this particular case using the `--treat-all-comment-as-code` flag allows us to keep the newline.

Python version: 3.7.7
isort version: 5.11.5

EDIT: Possibly the same root cause as #2095 ?

fzlzjerry:
## Summary

Preserve buffered blank lines when a potential import section turns out to contain only comments. `lines_before_imports` should affect imports, not ordinary code comments.

This fixes the discrepancy where stdin, `--stdout`, and `isort.code()` remove a blank line that the in-place file path leaves alone.

## Changes

- Restore the buffered whitespace when flushing a comment-only section.
- Add 37 API and CLI regression/control cases: configured and default spacing, LF/CRLF, comments at EOF, preceding real imports, and stdin/stdout/diff/check/in-place modes.
- Include a separate annotation-only prerequisite commit for the existing Bandit B101 finding in `parse.py`. The type-narrowing assertion is retained; it already had a Ruff S101 exception, and its AST is unchanged.

## Validation

- Final regression tests against the original core: 27 fail and 10 controls pass; all 37 pass with the fix, including the mypyc build.
- Pure Python on 3.10.20 and 3.13.5: 662 passed, 1 skipped on each. Python 3.13 coverage: 99.43% with branches enabled; `core.py` has 100% line/branch coverage.
- Python 3.13, mypyc: 659 passed, 4 skipped. Integration suite: 12 passed, 1 skipped.
- The complete repository `tox -e lint` pipeline passes: mypy, isort checks, flake8, Ruff checks/formatting, and Bandit.

Closes #2156.


fzlzjerry:
DeepSource reports one new security issue and one existing issue fixed, but its GitHub check provides no rule or file/line annotations. An anonymous request to the linked detailed report returned HTTP 403. Could someone with project access share the rule and location?

The separate lint-prerequisite commit only updates the checker annotation on the pre-existing type-narrowing assertion in `parse.py`; the assertion is retained and the normalized AST is unchanged. I'd like to confirm the reported finding before changing runtime logic or adding any further suppressions.

DanielNoord (approved):
Thanks!

PR #2659, opened 2026-09-18
Reject bool for int config settings and remove backwards compatibility for `force_grid_wrap`

Pitchfork-and-Torch:
## Summary
Config coercion uses `type(default_value)(value)`. For int settings that is `int(value)`, and `int(True) == 1`, so `Config(line_length=True)` silently became `line_length=1` (and `False` → `0`).

Reject `bool` before coercing int settings. Boolean flags (e.g. `force_single_line=True`) are unchanged.

## Test plan
- [x] New test `test_rejects_bool_for_int_settings` passes
- [x] `Config(line_length=88)` still works; bool flags still accepted


DanielNoord (approved):

DanielNoord:
Decided to close. This actually works already and I don't see benefit in removing the backward compatibility.

